PR #12, opened 2020-03-22
Horizon

theMetalkate:
Create Horizont component:

Screenshot 2020-03-22 at 11 38 10
Screenshot 2020-03-22 at 11 38 32


src/horizon.tsx
```diff
@@ -0,0 +1,71 @@
...
+    text: string
+    url: string
+}
+
+export default function Horizon(entries: HorizonProp[]) {
+    return (
+        <Wrapper>
+            {entries.map((horizonEntry) => {
+                return (
+                    <Item href={horizonEntry.url}>
```
Entkenntnis: please add a `key` prop here

src/horizon.tsx
```diff
@@ -0,0 +1,71 @@
...
+            })}
+        </Wrapper>
+    )
+}
+
+const Wrapper = styled.div`
+    align-items: center;
+    display: flex;
+    justify-content: space-between;
+    padding: 2rem 1.5rem 1.5rem;
```
Entkenntnis: use px here

src/horizon.tsx
```diff
@@ -0,0 +1,71 @@
...
+    max-width: 400px;
+    width: 98%;
+    @media (max-width: ${props => props.theme.breakpoints.sm}) {
+        max-width: 400px;
+    }
+`
+
+const Headline = styled.p`
+    font-weight: bold;
+    font-size: 20px;
```
Entkenntnis: use `rem` for font size

src/horizon.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react'
```
Entkenntnis: for consistency, change to `import React from 'react'`

src/horizon.tsx
theMetalkate: please take a look

src/horizon.tsx
theMetalkate: fixed

src/horizon.tsx
theMetalkate: fixed

src/horizon.tsx
Entkenntnis: key={index} is probably enough

PR #20, opened 2020-03-26
Breadcrumbs updates

elbotho:
Only breadcrumbs for now.

### DIFF
--- a/src/components/navigation/Breadcrumbs.tsx
+++ b/src/components/navigation/Breadcrumbs.tsx
@@ -104,7 +104,7 @@ const Breadcrumb = styled.a`
 const BreadcrumbLast = styled(Breadcrumb)`
 
   &:after {
-    display: none !important;
+    display: none;
   }
 
   @media (max-width: ${props => props.theme.breakpoints.sm}) {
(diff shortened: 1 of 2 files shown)


src/components/navigation/Breadcrumbs.tsx
```diff
@@ -81,15 +90,30 @@ const Breadcrumb = styled.a`
...
+      position: absolute;
+      margin-left: 13px
+    }
+  }
 `
 
 const BreadcrumbLast = styled(Breadcrumb)`
+
+  &:after {
+    display: none !important;
```
Entkenntnis: *theoretically* the extended styles should override existing ones. If this `!important` is necessary, consider adding a comment explaining why

elbotho:
what do you think is it okay to do the array changes (remove "alle themen" and skip last item) on the fronend for now?

src/components/navigation/Breadcrumbs.tsx
elbotho: overwrite works, just forgot to remove. thanks!

Entkenntnis:
> what do you think is it okay to do the array changes (remove "alle themen" and skip last item) on the fronend for now?

This is completely fine at the moment, the data fetching will be refactored soon

Thanks for the PR, I'm merging it.

PR #22, opened 2020-03-27
Footer style changes

elbotho:
For #14

### DIFF
--- a/src/components/content/Horizon.tsx
+++ b/src/components/content/Horizon.tsx
@@ -11,7 +11,7 @@ interface HorizonProp {
 export default function Horizon({ entries }: { entries: HorizonProp[] }) {
   return (
     <Wrapper>
-      {entries.map((horizonEntry, index) => {
+      {entries.sort(() => Math.random() - 0.5).map((horizonEntry, index) => {
         return (
           <Item href={horizonEntry.url} key={index}>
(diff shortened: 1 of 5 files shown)


src/components/content/Horizon.tsx
```diff
@@ -11,7 +11,7 @@ interface HorizonProp {
 export default function Horizon({ entries }: { entries: HorizonProp[] }) {
   return (
     <Wrapper>
-      {entries.map((horizonEntry, index) => {
+      {entries.sort(() => Math.random() - 0.5).map((horizonEntry, index) => {
```
Entkenntnis: :)

src/components/navigation/Footer.tsx
```diff
@@ -110,16 +112,14 @@ const TopButton = styled.div`
...
+  width: 100%;
 `
 
 const Summary = styled.div`
   background-color: ${props => props.theme.colors.lightblue};
-  display: flex;
-  flex-direction: column;
-  padding: 12px 15px 0 15px;
+  padding: 32px 0 16px;
+  ${makeResponsivePadding}
```
Entkenntnis: use explicit padding-top and padding-bottom here to make clear that the css helper is doing something

src/components/content/Horizon.tsx
elbotho: not true random, so sad :)

PR #27, opened 2020-03-27
Fix/article whitespace

Entkenntnis:
and: testing new colors


Entkenntnis:
leave colors out here, adding a lot of whitespace fixes

PR #28, opened 2020-03-29
Horizon: add headline and hover styling

theMetalkate:
Fix the styling for headline and hover.
* The headline is now a real headline tag (go with h4 to not interupt with SEO important headline tags)
* implement some hover styling for better UX (text get dark gray like navigation and a box shadow)

Screenshot 2020-03-29 at 11 45 03


Entkenntnis:
Thank you @theMetalkate, this looks awesome!

PR #30, opened 2020-03-29
feature notifications

theMetalkate:
Notifications
* highlight every second row for better reading and differentiation
* unreaded items are highlighted in bold and with a small dot indicator infront of the line

Screenshot 2020-03-29 at 14 29 58
Screenshot 2020-03-29 at 14 30 13


src/components/content/Notifications.tsx
```diff
@@ -0,0 +1,84 @@
...
+    timestamp: string,
+    readed: boolean
+}
+
+export default function Notifications(entries: NotificationProp[]) {
+    return (
+        <Wrapper>
+            {entries.map((entry, index) => {
+                return (
+                    <Item readed={entry.readed} className={ ( !entry.readed ? "unreaded" : "") }>
```
Entkenntnis: you are already passing the prop to the styled component, consider removing the className and adapt the styling using this prop

src/components/content/Notifications.tsx
```diff
@@ -0,0 +1,84 @@
...
+        &:before {
+            content: "";
+            display: inline-block;
+            background: ${props => props.theme.colors.brand};
+            border-radius: 50%;
+            width: 10px;
+            height: 10px;
+            margin-right: 7px;
+        }
+    }
```
Entkenntnis: you could write something like

```tsx
`
  ${props => props.unreaded ? `
      font-weight: ...;
      color: ${props.theme.colors.brand};
      ....;
    ` : ''}
`
```

theMetalkate: Ok, didn't know this is working. Should be fine noe

src/components/content/Notifications.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+interface NotificationProp {
+  author: User
+  body: string
+  timestamp: string
+  readed: boolean
+}
+
+export default function Notifications(entries: NotificationProp[]) {
```
Entkenntnis, line 16: Hi Kate, thank you for your contribution! To make this file a proper react function component, the argument must be an object with entries as a key, so we need `Notifications({entries} : NotificationsProps)` here, and the interface should be adapted, too.

Could you please fix this? The rest looks fine.

theMetalkate: fixed

PR #42, opened 2020-03-31
Share menu review

elbotho:


README.md
```diff
@@ -372,7 +372,7 @@ Show information to the user with modals. [react-modal](https://github.com/react
 
 ```tsx
 import React from 'react'
-import Modal from '../src/reactmodal' // our wrapper
+import Modal from '../Modal' // our wrapper
```
Entkenntnis: thanks

src/components/navigation/ShareModal.tsx
```diff
@@ -69,57 +75,78 @@ export default function ShareModal(props) {
...
+      <FontAwesomeIcon icon={entry.icon} /> {entry.title}
+    </Button>
+  ))
 }
 
-const StyledH2 = styled.h2`
-  font-weight: 700;
-  line-height: 1.35;
-  letter-spacing: -0.01em;
+export const StyledModal = styled(Modal)`
```
Entkenntnis: cool that this works. just to document this behaviour: add a small comment here pointing out that we are overriding the content styles

elbotho: Jup, thought about adding it to the readme as well, but it's an special case.
There seems to be a way to change all modal styles with styled components, it's okay but not super clean.

If we ever need it or if you prefer this to be the standard way: https://github.com/reactjs/react-modal/issues/627#issuecomment-578814799

elbotho:
Thanks for the review! Should be fine now.

PR #49, opened 2020-04-01
List and table styles

elbotho:
Small one this time :)

### DIFF
--- a/src/components/tags/StyledOl.tsx
+++ b/src/components/tags/StyledOl.tsx
@@ -9,6 +9,9 @@ export const StyledOl = styled.ol`
   padding-left: 25px;
   list-style-type: none;
   counter-reset: list-counter;
+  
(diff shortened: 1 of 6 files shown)


elbotho:
https://frontend-git-list-and-table-styles.dal123.now.sh/1565
has ul and table, did not find real ol so far

Entkenntnis:
Tables are fine!

Lists are a bit more involved: I want to support mixed content within list items (p/img/math) and nested lists, therefore all text is wrappen in p at the moment. The reduced margin between paragraph and list is intend. You can adapt the margins in `theme.spacing.mb.slim` and `theme.spacing.mb.li`:

grafik

So I have removed the styles that where conflicting and left everything that works fine.

Entkenntnis:
Nested lists: https://frontend-sooty-ten.now.sh/transparenz
ordered list: https://frontend-sooty-ten.now.sh/1553

PR #57, opened 2020-04-05
Add support for docker and docker-compose

kulla:
Docker is needed so that we can add the frontend to our Kubernetes cluster ( https://github.com/serlo/infrastructure-env-production/issues/16 ). To test it you can run `yarn start`. Nextjs gets deployed via docker and is still available via `localhost:3000`. `yarn stop` will stop and remove the created container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+\.git
+Dockerfile
+\.gitignore
+\.prettierrc
+LICENSE
+\.next
+node_modules
+README\.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:lts-alpine as build
+
+WORKDIR /app
+COPY . .
+
+RUN yarn install --frozen-lockfile
+RUN yarn build
+
+FROM node:lts-alpine
+
+WORKDIR /app
+COPY public public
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /app
+COPY public public
+COPY --from=build /app/package.json package.json
+COPY --from=build /app/yarn.lock yarn.lock
+COPY --from=build /app/.next .next
+
+RUN yarn install --production --frozen-lockfile
+
+EXPOSE 3000
+CMD ["yarn", "start:next"]
```
kulla, line 20: I need to organize feedback from someone, who is already experienced in docker...

.dockerignore
```diff
@@ -0,0 +1,8 @@
+\.git
+Dockerfile
+\.gitignore
+\.prettierrc
+LICENSE
+\.next
+node_modules
+README\.md
```
kulla, line 8: @Entkenntnis Can you check whether this contains all files which are not necessary for building the frontend with nextjs?

Dockerfile
Entkenntnis: I would let @inyono feedback this

PR #60, opened 2020-04-05
Add makeMargin() to SpoilerContainer

kulla:
Before:

2020-04-05-220046_866x604_scrot

After:

2020-04-05-220056_874x605_scrot

### DIFF
--- a/src/components/content/SpoilerContainer.tsx
+++ b/src/components/content/SpoilerContainer.tsx
@@ -2,7 +2,9 @@ import styled from 'styled-components'
 import { makeMargin } from '../../helper/csshelper'
 
 const SpoilerContainer = styled.div<{ hide?: boolean }>`
-  ${makeMargin}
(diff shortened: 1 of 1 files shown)


kulla:
@Entkenntnis I noticed that for `SpoilerContainer` the width is set to 100%: https://github.com/serlo/frontend/blob/ca2a0d721c884a7626ad58d01af9d7751d6cd95c/src/components/content/SpoilerContainer.tsx#L3-L10 However this is missing for example in https://github.com/serlo/frontend/blob/ca2a0d721c884a7626ad58d01af9d7751d6cd95c/src/components/tags/StyledH2.tsx#L5-L18 Is this intented? Thus the Spoiler has a bigger width than the rest of the content...

Entkenntnis:
This is intended behaviour for mobile - making the spoiler more integrated on the screen. Probably not needed for desktop. I would suggest to add the margin on bigger screens above 500px

Entkenntnis:
done

PR #67, opened 2020-04-08
RFC: cache GraphQL API via ZEIT Smart CDN

inyono:
ZEIT updated [their pricing](https://zeit.co/pricing) which makes it way more feasible for us (since there are no Bandwidth limits anymore). For that to work well for our users, we still need to reduce the latency to our API server. For that, I created an API route proxying the actual GraphQL Query as GET Request so that the ZEIT Smart CDN can cache the request. Using [stale-while-revalidate](https://zeit.co/docs/v2/network/caching#stale-while-revalidate), this should introduce latency only when the ZEIT Smart CDN hasn't cached the API request yet and asynchronously update the cache in the background.

So basically my preferred approach (which is open for feedback) would be:
- Deploy frontend via ZEIT Now. This way, we don't need to think about scalability of the frontend at all, only the GraphQL API. And we can basically throw our current caching of HTML away (since ZEIT handles that).
- Deploy API in our cluster (w/ smart caching), so the API should also scale well since we have many reads but few writes.
- When we need a GraphQL Query in frontend, we proxy it via an API route w/ stale-while-revalidate (or other appropriate caching rules) so that ZEIT Smart CDN can cache it. We can decide per use case whether this is only done server-side (as it is currently the case), or client-side / both / something fancy like [zeit/swr](https://swr.now.sh/)
- If the request is only used on client-side (e.g. user-specific requests, or mutations when saving), we can talk to the GraphQL API directly.

Thoughts? You can test out the loading times / latency on https://frontend-r0mxv4d2u.now.sh/mathe

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "isomorphic-unfetch": "^3.0.0",
     "katex": "0.11.1",
     "next": "^9.3.2",
+    "next-absolute-url": "^1.0.0",
     "polished": "^3.0.0",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -10,6 +10,8 @@ import Breadcrumbs from '../src/components/navigation/Breadcrumbs'
 import HSpace from '../src/components/content/HSpace'
 import Horizon from '../src/components/content/Horizon'
 import { horizonData } from '../src/horizondata'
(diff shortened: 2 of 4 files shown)


Entkenntnis:
Load times are great! Stale-While-Revalidate seems to offer exactly the functionality, we are looking for, avoiding api request slow down the page.

Even better: The API-Route can also parse the content and load injection to further increase performance!

100% for this approach :)

Issue #64, opened 2020-04-06
Discussion: What else should we implement from serlo.org?


kulla:
In this issue I list the features which are currently missing in the frontend project but which are implemented at serlo.org

- [ ] Support for Hotjar (we might want to discuss this issue)
- [ ] Popup for the email newsletter https://github.com/serlo/serlo.org/blob/6983e369dff10dadd6b3832bf9be489f4619435d/packages/public/server/src/module/Ui/templates/layout/partials/popup.phtml
- [ ] The following MathJax extensions have been enabled. Do we need them?
     - [ ] [mediawiki-texvc](http://docs.mathjax.org/en/latest/input/tex/extensions/mediawiki-texvc.html)
     - [ ] [mhchem](http://docs.mathjax.org/en/latest/input/tex/extensions/mhchem.html)

Entkenntnis:
The cookie consent banner is alos missing?

kulla:
> The cookie consent banner is missing as well?

Yes, that's right :-)

kulla:
What I need to check:

- [ ] https://github.com/serlo/serlo.org/tree/master/packages/public/client
- [ ] https://github.com/serlo/serlo.org/tree/master/packages/public/server/src/module/Ui/templates/layout

Entkenntnis:
see #65

inyono:
- [ ] Sentry

inyono:
- [ ] Add custom tag to Google Analytics so that we can filter between old & redesign

elbotho:
@inyono could you elaborate or point to a person that can implement that? :)

elbotho:
`<meta robot` probably not useful. [See here](https://webmasters.stackexchange.com/questions/48153/what-does-all-mean-in-meta-robots-tag)
I opted to add our robots.txt but not include this <meta tag

elbotho:
favicon can also be omitted if it's in the default location. (memo to self: check if we should switch to PNG and offer bigger sizes as well)

elbotho:
Moved opensearch to backlog (#88) because it would need changes in the current search implementation.

elbotho:
**Content Types:**

`<meta name="content_type" content="article">`
I'm working on a rough implementation, but we probably need some normalizing.

these should be the strings we need to match:
`course, course-page, topic-folder, text-exercise-group, text-exercise, curriculum-topic-folder, applet
topic, curriculum-topic, topic-folder …`

Entkenntnis:
[...]

PR #76, opened 2020-04-12
feature cookie bar

theMetalkate:
I saw that in the old implementation you used local storage instead of a cookie with some other parameters. I'm not sure for what you need them. I used a normal cookie now but if you need can changed it to local storage implementation

Screenshot 2020-04-12 at 13 45 15
Screenshot 2020-04-12 at 13 45 32

### DIFF
--- a/src/components/content/CookieBar.tsx
+++ b/src/components/content/CookieBar.tsx
@@ -1,0 +1,76 @@
+import React from 'react'
+import styled from 'styled-components'
+import { useCookies } from 'react-cookie'
+
+export default function CookieBar() {
+  const [cookies, setCookie] = useCookies(['consent'])
+  let showCookieBar = cookies.consent
+
+  function accept() {
+    const newDate = new Date()
+    const day = newDate.getDate()
+    const month = newDate.getMonth() + 1
+    const year = newDate.getFullYear()
+
+    setCookie(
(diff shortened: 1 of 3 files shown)


src/components/content/CookieBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const month = newDate.getMonth() + 1
+    const year = newDate.getFullYear()
+
+    setCookie(
+      'consent',
+      `${year}-${month < 10 ? `0${month}` : `${month}`}-${day}`
+    )
+  }
+
+  if (!showCookieBar) return null
```
inyono, line 21: That doesn't match the current behavior of the consent banner. What we do currently:

1. Fetch the current revision of the privacy policy via https://de.serlo.org/privacy/json
2. Check whether the consented revision matches the current revision (if any)
3. If not, we show the cookie banner
4. On consent, we persist the revision the user has consented to. Furthermore, we track that via Google Analytics to get a feeling how many people ignore the consent banner (not necessary as it depends on #64)

Please have a look at the current implementation at https://github.com/serlo/serlo.org/blob/master/packages/public/client/src/frontend/modules/consent-banner.ts#L36

Entkenntnis:
adopt changes in #110

PR #104, opened 2020-04-20
Testpage for containers

Entkenntnis:
troubleshooting container issues:

https://frontend-git-test-containers.dal123.now.sh/_containers


elbotho:
Woah, so in depth 👍 
Thank you!

Entkenntnis:
Results: use margin auto for content container, use additional div for image max width, set word wrap to break word. Be careful with width 100% and margins.

This solves all issues at the moment.

PR #105, opened 2020-04-20
Sentry intergration

elbotho:
Basic implementation.
Server- and clientside report errors:

image

If we need more (e.g. build-id) at some point:
https://github.com/zeit/next.js/tree/canary/examples/with-sentry

If we need sourcemaps above check here:
https://github.com/zeit/next.js/tree/canary/examples/with-sentry
https://github.com/zeit/next-plugins/issues/309
https://spectrum.chat/next-js/general/config-nextjs-sourcemaps-in-sentry~630c66f8-69b4-41bf-8cad-98590b782322

I removed the whitelisted URLs because [the docs say](https://docs.sentry.io/error-reporting/configuration/?platform=javascript#whitelist-urls) this is not available for js.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -3,6 +3,10 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 })
 
 module.exports = withBundleAnalyzer({
+  env: {
+    SENTRY_DSN:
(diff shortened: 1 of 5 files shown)


Entkenntnis:
looks good.

The only thing that hurts its the bundle size of 18kb (gzip/minified) that gets added to every page. But this is probably worth it for now.

PR #109, opened 2020-04-20
Add Analytics

### LINKED ISSUE #77: Analytics
- [ ] Add Google Analytics tracking code
- [x] Check if a tag is a good way to filter old & redesign
- [x] Add custom tag

elbotho:
Working now. Basically the same implementation we currently use.
Closes #77.

I added a custom dimension "redesign" that should help us differentiate traffic from the new frontend.
Atm. it's not tracking into the main Serlo.org analytics project while we are still debugging. 

Just tracks on pageload. If we switch to routing at any point we need to add code for that.

Backlog: Easiest way for nextjs would be this plugin, that is not stable atm
https://github.com/zeit/next.js/tree/canary/packages/next-plugin-google-analytics

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -3,10 +3,6 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 })
 
 module.exports = withBundleAnalyzer({
-  env: {
-    SENTRY_DSN:
-      'https://b7a5f46510b945b7a3a78c47c6a6048a@o115070.ingest.sentry.io/5206333'
-  },
   webpack(config) {
     config.module.rules.push({
       test: /\.svg$/,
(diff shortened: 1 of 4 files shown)


elbotho:
Draft because I need some time to check if traffic is coming in correctly.

Entkenntnis:
just give me a notice if this needs to be merged

elbotho:
Okay @Entkenntnis I can't really test this, but I think it's ready!
Removed both `GA_TRACKING_ID` and `SENTRY_DSN` to be added in Vercel

Entkenntnis:
I have added the value to vercel (as in this PR): 

grafik

so hopefully this works out :+1:

PR #152, opened 2020-04-30
Meta description and og:image

elbotho:
okay, ready for review.
a bit more logic than expected.
The meta-description fallback is not super elegant but was the simplest way to go through the data.values I could think of. 

probably time to move the functions into another file.


Entkenntnis:
Thank you @elbotho for this PR! Totally agree that we should move some logic away from [...slug].js, most likely after our initial release

elbotho:
Thank you!

PR #155, opened 2020-05-05
WIP: Browser fixes 

elbotho:
PR so I can debug IE11 problems with the preview.


Entkenntnis:
Very nice! Can confirm that page is now quite usable on IE 11

PR #156, opened 2020-05-05
/api/ -> /api/frontend/

Entkenntnis:
to avoid possible name conflicts with existing code


inyono:
There shouldn't be a name collision since we reserved /api in the legacy backend for the API routes used by the GraphQL API. Did you find one?

Entkenntnis:
We found none, but where unsure about it, so decided to go the safe way.

PR #158, opened 2020-05-05
Rework MobileMetaMenu

elbotho:
PR to test simplified mobile meta menu online. 
Might work better on mobile, not as good on small laptop screens is my guess.


src/components/navigation/MetaMenu.tsx
```diff
@@ -15,35 +14,40 @@ interface MetaMenuLink {
 
 export default function MetaMenu(props: MetaMenuProps) {
   const { navigation, pagealias } = props
-  const links = navigation
+
+  const activeRef = useRef(null)
+  const containerRef = useRef(null)
```
Entkenntnis, line 19: Should we remove these two or are you planning to use them later on? (because they seem unused here)

src/components/navigation/MobileMetaMenu.tsx
```diff
@@ -1,98 +0,0 @@
-import React from 'react'
-import styled from 'styled-components'
-import ScrollMenu from 'react-horizontal-scrolling-menu'
```
Entkenntnis, line 3: Now you can also remove the package from our dependencies

src/components/navigation/MetaMenu.tsx
```diff
@@ -1,7 +1,6 @@
-import React from 'react'
+import React, { useRef, useEffect } from 'react'
```
Entkenntnis, line 1: useEffect is unused

src/components/navigation/MetaMenu.tsx
elbotho: thank you! Nice catch, actually forgot to move some code.

Entkenntnis:
Thank you!

PR #171, opened 2020-05-07
Ex group improvements

### LINKED ISSUE #161: Feedback: Exercises and groups
_by wolfgang, kathi and mike(?)_ ([source](https://www.loom.com/share/8adc3091bed64ec9b14dd8c56c5e566f))

- [x] lacks structure, clearer separation between exercises 
- [x] also clearer separation to solution
- [x] maybe solution more like in the editor currently (e.g. change text to "lösung einklappen")
- [x] exercise explanation should not have a colored background
- [ ] maybe reintroduce numbers (but clearer separation is the most important part)
- [x] MC/SC Ex: Options should not be full circles, to make them more inviting to click

elbotho:
closes #161

### DIFF
--- a/src/components/content/ExerciseGroup.tsx
+++ b/src/components/content/ExerciseGroup.tsx
@@ -1,15 +1,23 @@
+import React from 'react'
 import styled from 'styled-components'
 
 import SpoilerContainer from './SpoilerContainer'
 import SpoilerTitle from './SpoilerTitle'
 
 export default function ExerciseGroup({ children, license, groupIntro }) {
+  const groupChildren = React.Children.map(children, child => {
+    return React.cloneElement(child, { inGroup: true }, null)
+  })
+
   return (
     <Container>
       <SpoilerTitle open disabled>
-        <Label>{groupIntro}</Label>
+        <Label>Aufgabengruppe</Label>
       </SpoilerTitle>
-      <Content>{children}</Content>
+      <Content>
+        <ExerciseIntro>{groupIntro}</ExerciseIntro>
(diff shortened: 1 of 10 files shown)


src/components/content/ExerciseGroup.tsx
```diff
@@ -1,40 +1,50 @@
+import React from 'react'
 import styled from 'styled-components'
 
 import SpoilerContainer from './SpoilerContainer'
 import SpoilerTitle from './SpoilerTitle'
 
 export default function ExerciseGroup({ children, license, groupIntro }) {
+  const groupChildren = React.Children.map(children, child => {
+    return React.cloneElement(child, { inGroup: true }, null)
+  })
```
Entkenntnis, line 10: Well, well ... I think we can do better than this. (anyway, cool that this hack works)

Ok, the missing feature is, that the api fetcher is not passing the difference between normal exercises and grouped exercises forward, this can be done here:

https://github.com/serlo/frontend/blob/e5bce814f62abb9f8a99fa82af9c009d813eab43/src/fetcher/createData.tsx#L72-L87

You can add `inGroup: uuid.__typename === 'GroupedExercise'` below line 77, and also adapt the renderer: https://github.com/serlo/frontend/blob/e5bce814f62abb9f8a99fa82af9c009d813eab43/src/schema/articleRenderer.tsx#L355-L360

elbotho: thanks for the pointers!

Entkenntnis:
Thank yo for the improvements and fixes!

Issue #221, opened 2020-05-13
missing margin


simonkoehl:
![image](https://user-images.githubusercontent.com/3379377/81816867-36e07200-952c-11ea-8e3c-8a37e60f5604.png)

elbotho:
never seen this before, could you add a link?

simonkoehl:
https://de.serlo.org/mathe/geometrie/dreiecke,-vierecke,-kreise-andere-ebene-figuren/kreise-kreisteile/aufgaben-kreisen-kreisteilen

2a) and type in anything in the input field

elbotho:
@simonkoehl thanks! will be fixed soon.

Issue #186, opened 2020-05-12
Fix subtle react warnings


Entkenntnis:
~~`<div>` within `<p>`:~~

~~http://localhost:3000/mathe/funktionen/wichtige-funktionstypen-eigenschaften/quadratische-funktionen---parabeln/scheitelpunkt/aufgaben-berechnung-scheitelpunktes~~

Missing key prop:

http://localhost:3000/mathe/zahlen-groessen/grundrechenarten/rechnen-negativen-zahlen/addition-subtraktion-ganzer-zahlen

http://localhost:3000/mathe/zahlen-gr%C3%B6%C3%9Fen/prozent--zinsrechnung/aufgaben-grundwert,-prozentwert-prozentsatz

~~Prop ID did not match:~~

~~http://localhost:3000/mathe/zahlen-gr%C3%B6%C3%9Fen/gr%C3%B6%C3%9Fen-einheiten/aufgaben-umrechnen-zeiteinheiten~~

(only visible during development)

Entkenntnis:
@elbotho The last warning is caused by this line:

https://github.com/serlo/frontend/blob/a78a9adb25e5a80e377d78c51d77856df5a13358/src/components/content/ScMcExercise.tsx#L34

I see what you are trying to accomplish and this actually rely on random numbers ... well, needs some thoughts, but is not critical atm

elbotho:
jup, was a quick fix at that point, now that I know that this annoys next and we forward the ex numbers I can find a different solution.

Entkenntnis:
The first warning will be fixed by #214

PR #222, opened 2020-05-13
remove key warning, etc.

### LINKED ISSUE #216: improve sharing menu
image

Suggestion for improvement:

image

And btw. you could do the distance between icon and text with margin instead of blank character ;)

elbotho:
should fix last one `Prop ID did not match` in #186

also minor design fix for #221

closes #216


Entkenntnis:
Super cool, thank you!

PR #225, opened 2020-05-14
barebone i18n setup

Entkenntnis:
creating a barebone implementation to investigate possible implications of using i18next.

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import App from 'next/app'
+import i18next from 'i18next'
 
 import { config } from '@fortawesome/fontawesome-svg-core'
 import '@fortawesome/fontawesome-svg-core/styles.css'
@@ -14,6 +15,8 @@ import { theme } from '../src/theme'
 import * as Sentry from '@sentry/browser'
 
 import { version } from '../package.json'
+
+import de from '../translations/de.json'
 
 if (process.env.SENTRY_DSN !== undefined) {
   Sentry.init({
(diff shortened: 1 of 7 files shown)


pages/_app.js
```diff
@@ -31,6 +34,11 @@ const FontFix = createGlobalStyle`
   }
 `
 
+i18next.init({
+  lng: 'de',
+  resources: { de: { translation: de } }
+})
+
```
inyono, line 41: You probably want to use a more intelligent backend for that since this would always include the translation files for all languages. See e.g. the implementation in serlo.org

Entkenntnis:
Would it be possible to export the crowdin resources for german as .json (instead of .po?)

inyono:
The translation files for i18next are already uploaded & exported as JSON: https://github.com/serlo/serlo.org/tree/master/packages/private/i18n/resources. Please have a look at the serlo.org implementation since this pretty much handles most of the edge cases (Automatic string extraction, React integration, async loading of translation files)

Entkenntnis:
Just to clearify: As far as I can see the json files does not contain strings like "Die freie Lernplattform", "Mitmachen", etc ..., but only edtr-io related content?

Unfortunately, there are some unresolved issues with next-i18next and Vercel Now deployment :grimacing: 

grafik

(https://github.com/isaachinman/next-i18next)

reading through the discussion reveals some deeper edge cases regarding ssr, async and bundling ..

inyono:
> Just to clearify: As far as I can see the json files does not contain strings like "Die freie Lernplattform", "Mitmachen", etc ..., but only edtr-io related content?

Yes, strings defined by the PHP server and the client are completely independent. So you can't use that translations.

> Unfortunately, there are some unresolved issues with next-i18next and Vercel Now deployment
Might need to handle that ourselves, then. Basically:
- Client can be handled as in serlo.org (i.e. pass down language some (e.g. by reserving the first path segment to the language) & asynchronously load the translation files
- Server (i.e. serverless) needs to bundle the translation files in the code itself so that you can access them during server side rendering

Entkenntnis:
closed because of #226

PR #232, opened 2020-05-14
Share menu feedback fixes etc.

### LINKED ISSUE #211: Too many share-buttons
For example on the https://de.serlo.org/jobs page there are three-share-buttons: this seems a bit much (one on top, one sticky and one below)

elbotho:
closes #211


Entkenntnis:
@elbotho Thank you for adding the fixes! The contentId was indeed missing from the api fetcher, I have added it, could you check again that this is working as expected?

elbotho:
Thanks! Works fine.

PR #241, opened 2020-05-19
Add newsletter popup

### LINKED ISSUE #181: Newsletter-Popup
While surfing the site, at some points a popup for the newsletter will pop up:

https://github.com/serlo/serlo.org/blob/6983e369dff10dadd6b3832bf9be489f4619435d/packages/public/server/src/module/Ui/templates/layout/partials/popup.phtml

This is currently missing in the frontend.

elbotho:
closes #181
Maybe not the react way, but this way we make sure code only loads if needed.

### DIFF
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -23,6 +23,9 @@ const Breadcrumbs = dynamic(() =>
   import('../src/components/navigation/Breadcrumbs')
 )
 const Topic = dynamic(() => import('../src/components/content/Topic'))
+const NewsletterPopup = dynamic(() =>
+  import('../src/components/NewsletterPopup')
+)
 
 enum MetaImageEnum {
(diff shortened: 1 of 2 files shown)


pages/[...slug].tsx
```diff
@@ -23,6 +23,9 @@ const Breadcrumbs = dynamic(() =>
   import('../src/components/navigation/Breadcrumbs')
 )
 const Topic = dynamic(() => import('../src/components/content/Topic'))
+const NewsletterPopup = dynamic(() =>
+  import('../src/components/NewsletterPopup')
```
Comment (inyono) [inline pages/[...slug].tsx:27]: You might want to disable SSR for that component (see https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr)

Comment (elbotho) [reply pages/[...slug].tsx:27]: nice!

PR #242, opened 2020-05-19
add basic lightbox

### LINKED ISSUE #195: Missing Feature: Lightbox
serlo.org has a simple lightbox to enlarge small images: [Example](https://de.serlo.org/nachhaltigkeit/anbau/gelaende-anbauplanung/pflanzenfamilien)

elbotho:
@Entkenntnis any elegant idea how we could only add this functionality to images that don't have the whole width? (I guess that equals childrens of `Col`)

closes #195


Entkenntnis:
Maybe checking https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalWidth?

And probably adjusting z-index here?


grafik

elbotho:
thanks, will investigate!

elbotho:
Closed until there is support for image columns in the new editor.

PR #243, opened 2020-05-19
add favicon sizes etc

### LINKED ISSUE #97: Favicon: PNG and sizes
We should consider adding our Favicon in bigger sizes as png and add the relevant meta tags.

elbotho:
closes #97


elbotho:
Vote was inconclusive, the new feather is good enough for now I'd say.

Entkenntnis:
There is one problem left here: The cloudflare worker is only redirecting assets within `public/_assets/` to the frontend, so the icons in the PR wouldn't be accessible on `de.serlo.org`.

There are at least two options:

1. Add icons to serlo.org directly (making an additional PR there)
2. Or load icons from `/_assets/`

elbotho:
Okay, chose second solution.

PR #246, opened 2020-05-20
add youtube placeholder

### LINKED ISSUE #197: Use Placeholder for Video injections.
Lazy load videos by using a placeholder that replaces itself with the actual iframe on activation.

elbotho:
Uses https://github.com/ibrahimcesar/react-lite-youtube-embed based on https://github.com/paulirish/lite-youtube-embed
just displays a preview image (and preconnects) until activated.

possible improvements in the future
- ~~check if `sddefault` thumbnail size actually exists and replace if not.~~
- even lazy load the preview image
- make use of webp images provided by yt

but should be good enough for now (224× faster if their claim is true :D)

closes #197

### DIFF
--- a/external/LiteYouTubeEmbed.tsx
+++ b/external/LiteYouTubeEmbed.tsx
@@ -1,0 +1,198 @@
+import React from 'react'
+import styled from 'styled-components'
+import { makeMargin } from '../src/helper/csshelper'
+
+//fixed and modified version of https://github.com/ibrahimcesar/react-lite-youtube-embed
+//embeded here for faster developement and tracking of changes
+
+interface LiteYouTubeEmbedProps {
+  id: string
+  playlist?: boolean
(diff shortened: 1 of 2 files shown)


elbotho:
> Wow, this really improves load performance! Thank you for investigating this issue.
> 
> Would you mind moving 3party code into the `/external` folder in the root of this repo? This will make it clearer where the code is coming from.

okay, just moved the code. 
in the long term I'd like to PR my fixes/improvements to the original repo again, but there are some questions to answer first.

elbotho:
@Entkenntnis moving to jsx inline styles would be the simplest approach, or how would you handle this?

Entkenntnis:
something in the direction of https://stackoverflow.com/a/58983746 maybe?

In addition, react-lite-youtube-embed could have an option to disable importing the css file

PR #247, opened 2020-05-22
use x-serlo-api header

### LINKED ISSUE #177: Use HTTP header X-SERLO-API
With https://github.com/serlo/serlo.org-cloudflare-worker/pull/11 the cloudflare worker sends the HTTP-header `X-SERLO-API` which points to the graphql endpoint which shall be used. The frontend client should use this variable to determine the endpoint. The possible values are:

* `https://api.serlo-development.dev/graphql`
* `https://api.serlo-staging.dev/graphql`
* `https://api.serlo.org/graphql`

Entkenntnis:
fixes #177

### DIFF
--- a/src/fetcher/serlo-api.tsx
+++ b/src/fetcher/serlo-api.tsx
@@ -2,9 +2,20 @@ import { request } from 'graphql-request'
 import { dataQuery, idQuery } from './query'
 import { processResponse } from './processResponse'
 
-const endpoint = 'https://api.serlo.org/graphql'
+const availableEndpoints = [
+  'https://api.serlo-development.dev/graphql',
+  'https://api.serlo-staging.dev/graphql',
+  'https://api.serlo.org/graphql'
+]
 
-export default async function fetchContent(alias: string, redirect) {
+export default async function fetchContent(
+  alias: string,
+  redirect,
+  forwardedEndpoint = ''
+) {
+  const endpoint = availableEndpoints.includes(forwardedEndpoint)
+    ? forwardedEndpoint
+    : availableEndpoints[2]
   try {
     if (redirect && /^\/[\d]+$/.test(alias)) {
       // redirect id to alias
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -218,10 +218,16 @@ const StyledIcon = styled(FontAwesomeIcon)`
 
 export async function getServerSideProps(props) {
   const { origin } = absoluteUrl(props.req)
+  const endpoint = props.req.headers['x-serlo-api']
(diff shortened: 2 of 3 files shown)


src/fetcher/serlo-api.tsx
```diff
@@ -2,9 +2,20 @@ import { request } from 'graphql-request'
 import { dataQuery, idQuery } from './query'
 import { processResponse } from './processResponse'
 
-const endpoint = 'https://api.serlo.org/graphql'
+const availableEndpoints = [
+  'https://api.serlo-development.dev/graphql',
+  'https://api.serlo-staging.dev/graphql',
+  'https://api.serlo.org/graphql'
+]
```
kulla, line 9:
```suggestion
const defaultEndpoint = 'https://api.serlo.org/graphql'
const availableEndpoints = [
  'https://api.serlo-development.dev/graphql',
  'https://api.serlo-staging.dev/graphql',
  defaultEndpoint
]
```

src/fetcher/serlo-api.tsx
```diff
@@ -2,9 +2,20 @@ import { request } from 'graphql-request'
...
 
-export default async function fetchContent(alias: string, redirect) {
+export default async function fetchContent(
+  alias: string,
+  redirect,
+  forwardedEndpoint = ''
+) {
+  const endpoint = availableEndpoints.includes(forwardedEndpoint)
+    ? forwardedEndpoint
+    : availableEndpoints[2]
```
kulla, line 18:
```suggestion
  const endpoint = availableEndpoints.includes(forwardedEndpoint)
    ? forwardedEndpoint
    : defaultEndpoint
```

src/fetcher/serlo-api.tsx
inyono: I'd prefer if we'd use `api.serlo-development.dev` by default.

inyono:
Another thought: we won't be able to test auth & features for logged in users with preview deployments since Hydra validates that the request originate from the correct client URL. To prevent this, we could also tell CF which URL it should proxy for the front end (e.g. with a cookie). This way, we can quickly switch between different API environments and also test the compatibility of a preview deployment with the rest of the system (including login / logout etc.). So basically:

- CF proxies cookie.frontend-url (or frontend.serlo.now = master if no cookie set)
- When proxying, CF sends down the API that should be used via Header
- CF exposes routes that allow to set / unset cookie.frontend-url
- Might want to automatically show a banner or something if that cookie is set.

Thoughts?

Entkenntnis:
after some discussion, we will use another approach to handle this

PR #254, opened 2020-05-22
render code

### LINKED ISSUE #245: Code is not displaying, Informatik somewhat unusable
Subject Informatik uses code-blocks from the old editor, e.g.
https://de.serlo.org/informatik/algorithmen/robot-karol/wiederholungen

Currently those are not displayed in the new design, which makes a lot of content currently unusable.
@yolo-bird do you know if and when those will be converted?

---

Also there is a course that uses inline-code, which is currently not supported in the new editor:
https://de.serlo.org/informatik/internet-dinge-circuitpython/anf%C3%A4ngerkurse/grundlagen-1/mu-starten
@yolo-bird who could rework this one?

Entkenntnis:
fixes #245

### DIFF
--- a/src/components/content/Code.tsx
+++ b/src/components/content/Code.tsx
@@ -1,0 +1,15 @@
+import styled from 'styled-components'
+import { makeMargin } from '../../helper/csshelper'
+
+export default function Code({ content }) {
+  return (
+    <Pre>
+      <code>{content}</code>
+    </Pre>
+  )
+}
+
+const Pre = styled.pre`
+  ${makeMargin}
+  margin-bottom: ${props => props.theme.spacing.mb.block};
+`
--- a/src/components/content/Code.tsx
+++ b/src/components/content/Code.tsx
@@ -11,6 +11,7 @@ export default function Code({ content }) {
 
 const Pre = styled.pre`
   ${makeMargin}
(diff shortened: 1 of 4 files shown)


elbotho:
okay @Entkenntnis basic styling for code blocks is done. 
I ran into a problem with wrapping again. The only way I see to differentiate between inline code and blocks is to check the parent (or forward that information to the component). 

If it's a `<p>` it should render inline (without `<pre>`) otherwise it breaks the layout and also results in a warning
>Warning: Expected server HTML to contain a matching pre> in p>.

Do you have an idea how we could forward that?

Entkenntnis:
@elbotho Can you give me the links to the corresponding pages? Haven't tested it much, yet - we can fix many html problems by tweaking the converter ... and probably adding a new element type for inline code

elbotho:
It's [here](http://localhost:3000/informatik/internet-dinge-circuitpython/anf%C3%A4ngerkurse/grundlagen-1/erstes-programm) but this is not currently supported in the new editor, is not widely used and we should not put a lot of work into it!

If you don't have a quick solution I'd be okay with a dirty but well documented global css hack for that.

Entkenntnis:
Quickfix added: ignoring inline code

PR #265, opened 2020-05-26
Pretty links

### LINKED ISSUE #194: Use readable hrefs for links
Currently the links use the content id in the hrefs.
I'd prefer the longer full links. 
@Entkenntnis Is there an elegant way to add that to the articleRenderer? 
https://github.com/serlo/frontend/blob/b8460970768e23223f2176729cc4417be188ba67/src/schema/articleRenderer.tsx#L148

elbotho:
closes #194

### DIFF
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -50,27 +50,6 @@ function PageView(props) {
     prettyLinks
   } = data
 
-  // const extractLinks = (arr, links) => {
-  //   arr.forEach(obj => {
-  //     if (obj.type === 'a') {
-  //       if (!/^\/[\d]+$/.test(obj.href)) return
-  //       console.log('t: ' + obj.href.substring(1))
-  //       const id = parseInt(obj.href.substring(1))
-  //       if (links.includes(id) === false)
-  //         console.log('t: ' + obj.href.substring(1))
-  //       // links.push(id)
-  //       // }
-  //     }
-  //     // console.log(obj)
-  //     // console.log(obj.children)
-  //     // console.log(obj.children?.length)
-  //     if (obj.children?.length > 0) extractLinks(obj.children, links)
-  //   })
-  //   return links
-  // }
-
-  // console.log(extractLinks(data.data.value.children, []))
-
   function getMetaContentType() {
     //match legacy content types that are used by google custom search
     if (contentType === undefined) return ''
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -50,6 +50,27 @@ function PageView(props) {
     prettyLinks
   } = data
 
+  // const extractLinks = (arr, links) => {
+  //   arr.forEach(obj => {
+  //     if (obj.type === 'a') {
+  //       if (!/^\/[\d]+$/.test(obj.href)) return
+  //       console.log('t: ' + obj.href.substring(1))
+  //       const id = parseInt(obj.href.substring(1))
+  //       if (links.includes(id) === false)
+  //         console.log('t: ' + obj.href.substring(1))
+  //       // links.push(id)
+  //       // }
+  //     }
+  //     // console.log(obj)
+  //     // console.log(obj.children)
+  //     // console.log(obj.children?.length)
+  //     if (obj.children?.length > 0) extractLinks(obj.children, links)
+  //   })
+  //   return links
+  // }
+
+  // console.log(extractLinks(data.data.value.children, []))
+
   function getMetaContentType() {
     //match legacy content types that are used by google custom search
     if (contentType === undefined) return ''
@@ -104,7 +125,7 @@ function PageView(props) {
   return (
     <>
       <Head>
-        <title>{title}</title>
+        <title>{title ? title : 'Serlo.org'}</title>
         <meta name="content_type" content={getMetaContentType()} />
         {metaDescription && (
           <meta name="description" content={metaDescription} />
--- a/pages/[...slug].tsx
+++ b/pages/[...slug].tsx
@@ -46,7 +46,8 @@ function PageView(props) {
     contentType,
     title,
     navigation,
-    license
+    license,
+    prettyLinks
   } = data
 
   function getMetaContentType() {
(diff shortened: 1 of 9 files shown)


src/components/content/Link.tsx
```diff
@@ -0,0 +1,38 @@
...
+    request(endpoint, idQuery(element.href.substring(1))).then(res => {
+      if (res && res.uuid && res.uuid.alias) setPrettyHref(res.uuid.alias)
+    })
+  }, [])
+
+  if (!element.href)
+    return <React.Fragment {...attributes}>{children}</React.Fragment>
+
+  return (
+    <StyledA href={prettyHref ? prettyHref : element.href} {...attributes}>
```
inyono, line 33: Doesn't this still use the non-pretty href on server? (which is IMHO the important part were we want the pretty href because Google)

elbotho: totally right, was only focusing on UX

elbotho: will try to make it work on server

elbotho: Just to make sure: would it be appropriate to use `getServerSideProps` in this case?

inyono: If you get the server side props down to the Link component somehow, sure. We probably need to handle that universally in the API fetch logic we have currently in place, e.g.
- Fetch basic data by alias or id (that is what we currently do)
- Go over that and make a list of links we need to prettify
- Fetch pretty alias for that links
- Pass that down to the page (and to the links via context)

This way, Vercel can still cache the API request and the server-side rendered stuff contains the pretty links.

elbotho: sorry, this is getting more complex then anticipated (well that never happened before :).
Maybe one last question before we move this to a later stage: Are you proposing we basically go through the returned `data.value.children` and collect all the `a` and build a new query to get their pretty links?

inyono: Yep. If you see another way, I'm all ears. The main issue we need to work around is: we can't do `getServerSideProps` on component level (only on page level). This means: the page itself must fetch the data itself (since we want them during server-side rendering) and pass them down somehow via context.

elbotho: links as async components? could that work? otherwise I'll try your suggestion.

elbotho: scratch that, I thought privacy was loaded this way, but it's front end only as well.

inyono:
#### 📝Changed routes:


| [**/\[...slug\]**](https://frontend-git-pretty-links-test.serlo.now.sh/[...slug]) |
|:-:|
|<a href="https://frontend-git-pretty-links-test.serlo.now.sh/[...slug]">Screenshot of /[...slug]</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/SRIDl3MbYyQrd7pkLQIhg0j6R88_.png">(view full size)</a> |







Commit 110c9fb3399d7ef03ab72911a170cda8929c0912 (https://frontend-git-pretty-links-test.serlo.now.sh).

elbotho:
So far so good, 
I decided to keep the prettyLink lookup object separate and not rewrite the whole data with some new aliases in`serlo-api.tsx`
Don't know if it's really more performant this way, but it seems to work now.

@inyono and @Entkenntnis this is a bit more technical than my usual PRs so feedback/improvements are welcome.

pages/[...slug].tsx
```diff
@@ -46,9 +46,31 @@ function PageView(props) {
...
+  //     // console.log(obj)
+  //     // console.log(obj.children)
+  //     // console.log(obj.children?.length)
+  //     if (obj.children?.length > 0) extractLinks(obj.children, links)
+  //   })
+  //   return links
+  // }
+
+  // console.log(extractLinks(data.data.value.children, []))
+
```
Comment (inyono) [inline pages/[...slug].tsx:73]: Is this still needed?

src/components/content/ArticlePage.tsx
```diff
@@ -55,7 +60,7 @@ export default function ArticlePage({ data, contentId, contentType }) {
           <FontAwesomeIcon icon={faShareAlt} size="1x" /> Teilen
         </ToolLineButton>
       </ToolLine>
-      {data.value && renderArticle(data.value.children)}
+      {data.value && renderArticle(data.value.children, false, prettyLinks)}
```
inyono, line 63: We should refactor that into an object parameter for readability.

src/fetcher/serlo-api.tsx
```diff
@@ -39,7 +40,15 @@ export default async function fetchContent(alias: string, redirect) {
       return { redirect: reqData.uuid.alias }
     }
     const contentId = reqData.uuid.id
-    return { contentId, alias, ...processResponse(reqData) }
+
+    const processed = processResponse(reqData)
+    const processedData = processed.data as any
```
inyono, line 45: *sad face* 😉, should probably refactor / type / unit test that as a follow-up sometime.

elbotho: yeah… I know.
and the really sad thing is that even that hack took me some time to figure out. 
TS is great in making me feel stupid, that's for sure :) if you have time and hints how I could fix it let me know.

src/components/content/ArticlePage.tsx
elbotho: something like?
``` tsx
{data.value && renderArticle({children: data.value.children, parameter: false, prettyLinks: prettyLinks)}
```

inyono: Yep, except maybe a more descriptive name for parameter (not sure what it does currently). IIRC, the second parameter was optional anyways (so you can remove it from your call again)

src/fetcher/serlo-api.tsx
inyono: I'm fine with leaving that at it is for now. My PR #267 showed that there are many places where the typing is `any` and we should go over that sometime. This particular line would probably okay if `processResponse` was properly typed.

elbotho:
@inyono uses Context now. I think I addressed the main points now. 
The typing is preliminary, but better than nothing? :)

PR #283, opened 2020-06-04
Work on types

elbotho:
working through `/components/*`.
`/content` in progress.

### Stats:

master:
yarn lint:eslint -> ✖ 7 problems (0 errors, 7 warnings)
yarn lint:eslint:strict -> ✖ 1850 problems (1672 errors, 178 warnings)

branch: 
yarn lint -> no warnings
yarn yarn lint:eslint:strict -> ~~✖ 1739 problems (1580 errors, 159 warnings)~~
✖ 1687 problems (1543 errors, 144 warnings)


elbotho:
@inyono maybe you can have a quick look just to make sure I go in the wrong direction somewhere.

src/components/content/topic.tsx
```diff
@@ -87,7 +87,7 @@ export function Topic({ data, contentId }: TopicProps) {
           ))}
         {data.exercises &&
           // TODO: needs type declaration
-          data.exercises.map((exercise: any, i: any) => (
+          data.exercises.map((exercise: any, i: number) => (
```
inyono, line 90: shouldn't need any type declaration since `data.exercises` is already typed

src/components/landing/landing-subjects.tsx
```diff
@@ -54,19 +54,21 @@ export function LandingSubjects() {
...
 
-// TODO: needs type declaration
-type SubjectProps = any
-
-function Subject(props: SubjectProps) {
-  const { url, title, SubjectSVG } = props
+interface SubjectProps {
+  url: string
+  title: string
+  SubjectSVG: React.ReactNode
```
inyono, line 60: should be in lowercase since `SubjectSVG` suggests this is a component (which it is not)

src/components/navigation/course-footer.tsx
```diff
@@ -5,8 +5,10 @@ import styled from 'styled-components'
 
 import { makeMargin, makeDefaultButton } from '../../helper/css'
 
-// TODO: needs type declaration
-export type CourseFooterProps = any
+export interface CourseFooterProps {
+  opener: (e: React.MouseEvent<HTMLAnchorElement>) => void
```
inyono, line 9: would probably rename that to `on*` to be consistent with other event handlers. Haven't looked too much into the code to know when this is actually triggered

src/components/navigation/course-navigation.tsx
```diff
@@ -26,12 +38,14 @@ export function CourseNavigation({
       {open ? (
         <StyledOl>
           {/* TODO: needs type declaration */}
```
inyono, line 40:
```suggestion

```

src/components/navigation/mobile-menu.tsx
```diff
@@ -25,17 +30,19 @@ export function MobileMenu(props: MobileMenuProps) {
...
-type EntryProps = any
+interface MobileMenuLink {
+  childKey?: string
+  isChild?: boolean
+}
 
 function Entry({
   url,
   title,
+  icon = undefined,
```
inyono, line 41: That shouldn't be needed since this is the default JS behavior (keys that aren't set are `undefined`)

src/components/navigation/mobile-menu.tsx
```diff
@@ -25,17 +30,19 @@ export function MobileMenu(props: MobileMenuProps) {
...
+  children = undefined,
```
inyono, line 42: same

src/components/navigation/mobile-menu.tsx
```diff
@@ -25,17 +30,19 @@ export function MobileMenu(props: MobileMenuProps) {
...
   childKey = undefined,
```
inyono, line 43: same

src/components/navigation/mobile-menu.tsx
```diff
@@ -25,17 +30,19 @@ export function MobileMenu(props: MobileMenuProps) {
   )
 }
 
-// TODO: needs type declaration
-type EntryProps = any
+interface MobileMenuLink {
```
inyono, line 33: I'd probably let the interface name match `ComponentNameProps` as long as there is no reason not to do that

src/components/navigation/search-results.tsx
```diff
@@ -3,11 +3,13 @@ import styled from 'styled-components'
...
-type SearchResultProps = any
+interface SearchResultProps {
+  className?: string
+  children: React.ReactNode
+}
 
-export function SearchResults(props: SearchResultProps) {
-  return <SearchResultsWrap {...props}>{props.children}</SearchResultsWrap>
+export function SearchResults({ className, children }: SearchResultProps) {
+  return <SearchResultsWrap className={className}>{children}</SearchResultsWrap>
```
inyono, line 12: What's the point of this React Component if it justs passes the the same props down?

src/components/navigation/share-modal.tsx
```diff
@@ -119,12 +120,23 @@ export function ShareModal(props: ShareModalProps) {
   )
 }
 
-// TODO: needs type declaration
-function buildButtons(list: any, copyToClipboard: any) {
-  // TODO: needs type declaration
-  return list.map((entry: any) => (
+interface EntryProps {
```
inyono, line 123: Name suffix shouldn't be `Prop` since this isn't inteded for a React Component, right? So rather `Data` or `Payload` or something.

src/schema/article-renderer.tsx
```diff
@@ -120,15 +120,25 @@ export const articleColors = {
...
 // TODO: needs type declaration
+
+interface RenderLeafProps {
+  leaf: {
+    color?: 'blue' | 'green' | 'orange'
+    em?: boolean
+    strong?: boolean
+  }
+  attributes: any
+  children: any
```
inyono, line 131: React children should be `React.ReactNode` usually

inyono:
> @inyono maybe you can have a quick look just to make sure I go in the wrong direction somewhere.

@elbotho Noted some things :)

elbotho:
Thank you! 
But in general this is helpful, yes? :)

inyono:
Yep, definitely 👍

inyono:
Feel free to mark this as "ready for review" as soon as you have a state that is mergable (to avoid rebase "fun" ^^)

src/components/landing/landing-subjects.tsx
elbotho: It's an React.FunctionComponent no?

src/components/navigation/mobile-menu.tsx
elbotho: This uses [Declaration Merging](
https://www.typescriptlang.org/docs/handbook/declaration-merging.html). Is that a good enough reason?

src/components/landing/landing-subjects.tsx
inyono: Nope. `React.ReactNode` represents a rendered component, e.g. `<FooBar />`. The corresponding component would be `FooBar` (so without JSX). So from naming, this should be:
```tsx
const foo = {
  subjectSVG: <Subject /> // type React.ReactNode   or
  SubjectSVG: Subject // type React.ComponentType<{}>
}
```

src/schema/article-renderer.tsx
elbotho: what about `attributes`? this comes up more often.

src/components/navigation/mobile-menu.tsx
inyono: I wouldn't do that, especially not for React components. Instead:
```ts
interface EntryProps extends MobileMenuLink {
  childKey?: string
  isChild?: boolean
}
```
achieves the same and you don't need to know what there is another `MobileMenuLink` in scope that is also used.

src/components/navigation/search-results.tsx
elbotho: just to export messy stying to be honest

src/schema/article-renderer.tsx
inyono: Not sure, seems do be David's stuff (which is insipired by Slate's code).

src/components/navigation/search-results.tsx
inyono: Mh... but isn't this current thing equivalent to
```tsx
const SearchResults = SearchResultsWrap
```
since `SearchResults` does absolutely nothing (no additional styles, no wrapping, nothing)

src/components/navigation/mobile-menu.tsx
elbotho: nice, did not find this one in the docs.

src/components/navigation/search-results.tsx
elbotho: ```export const SearchResults = styled.div```
seems to work, there was a reason for doing it like this at some point… well well.

PR #286, opened 2020-06-04
use client-side routing

### LINKED ISSUE #259: client side navigation
It's probably worth to give client side navigation another try and test performance and UX, especially as we have a new infrastructure and cache mechanisms

Entkenntnis:
exploring client-side routing and it looks promising, fixes #259 

Currently available within:

- taxonomy (headings and links)
- articles (prettyLink to other uuids)
- breadcrumbs


elbotho:
pretty snappy if the content is in cache already 👍 
for everything else we definitely need a loading indicator.

Entkenntnis:
@elbotho Finally I had some time exploring client-side-navigation and the code is quite straight forward. Are you interested in looking into this PR and test the UX?

maybe relevant: https://github.com/rstacruz/nprogress

elbotho:
Another option would be to trigger the browsers native load indicators:
https://stackoverflow.com/a/16745874/3294973

elbotho:
Let's try to add https://github.com/tanem/react-nprogress 
there is even a nextjs example, but it uses next router (which we don't, right?)

Entkenntnis:
now we are using next router

inyono:
Uh, I like 🎉 Seems like this will be a major improvement and only requires a little work for the non-cached edge cases.

Entkenntnis:
There is still an issue with back behaviour: The page scrolls first, and after some time the page loads. Here is a lengthy discussion with a possible fix at the end: https://github.com/vercel/next.js/issues/3303

elbotho:
Okay, spend way to long on this. We could implement a local cache that sould fix the back-button problem [according to this](https://levelup.gitconnected.com/6-tips-using-next-js-for-your-next-web-app-e3f056fa46). 
That sound cool, because it's instant and we save a server call. 
But it's also more complex than in the example because we use getServerSideProps.
@Entkenntnis @inyono  do you think that would that make sense? 

It would go like this: we use shallow routing (no server call) if the requested page is already in the cache (which is only a js variable) and leave everything else like it is currently.

Entkenntnis:
There is a very recent PR adressing this issue: https://github.com/vercel/next.js/pull/13914, hopefully it gets some attention

Entkenntnis:
@elbotho I have tried to implement it as you described: Returning to getInitalProps, we have more control over how to fetch data, every page load is caches (in js) and if a page has already been called, the cache is used.

elbotho:
Nice! sure makes the back button more fun: wooosh 😸 
What was the reason to go from`getInitalProps` vs `getServerSideProps` before?  
Any drawbacks?

Entkenntnis:
getServerSideProps was recommended in a release some months ago, so we switched - but I found no real adventage for us yet and we are using getInitialProps properly (bundling, etc...) so it should be safe, in hintsight there was actually no real reason to switch anyway :grinning:

Issue #262, opened 2020-05-25
course broken


Entkenntnis:
Reported by Nish:

Not all pages of the course "Vektoren" https://de.serlo.org/43174 are loading.

More explizit, the pages:

Was ist ein Vektor? (1/4) (https://de.serlo.org/mathe/geometrie/analytische-geometrie/grundbegriffe-vektorrechnung/vektorbegriff/einf%C3%BChrung-vektorbegriff-(vektoren-ebene-)/vektor?-(1/4))
Was ist ein Vektor? (2/4) (https://de.serlo.org/mathe/geometrie/analytische-geometrie/grundbegriffe-vektorrechnung/vektorbegriff/einf%C3%BChrung-vektorbegriff-(vektoren-ebene-)/vektor?-(2/4))

are not loading and lead to a 404.

elbotho:
There is a `?` in the url that is not escaped.
(potential security issue?)
`/vektor?-(1/4)` should be `vektor%3F-(1/4)`

Could that be a bug in the API @inyono ? 

Also the URLs are a mess anyway :)

Entkenntnis:
@inyono Is it intended that an alias is allowed to contain question marks? If this is the case, I would look through the api fetcher and add proper escaping. (I hope, slashes are forbidden :eyes:  )

inyono:
That's currently possible, yes. Seems like we don't handle that correctly in the API so far, we should probably url-encode the alias.

inyono:
cc https://github.com/serlo/api.serlo.org/issues/19

inyono:
Okay, just checked: `document.location` also reports the url-encoded path. Is it fine for you, if I expect the url-encoded alias as input & output the url-encoded alias, always?

Entkenntnis:
This would be fine! (Just to clearify: this means the graphql request will expect the alias to be url-encoded and will only output the alias as url-encoded?)

inyono:
Fixed, API can work both with url-encoded and non-url-encoded stuff, so this needs no changes to frontend at all :)

PR #288, opened 2020-06-05
Testing url-encoded aliases

inyono:
Test deployment so that Vercel empties the cache. Aliases should now outputed url-encoded all the time. Should need no change to code base at all.


Entkenntnis:
fixes #262

inyono:
Yep, looks good now. Pushing to prod.

PR #297, opened 2020-06-08
Complete horizon data, add weight shuffle

### LINKED ISSUE #280: Complete the horizon data
In the file https://github.com/serlo/frontend/blob/master/src/data/horizon.tsx all the current horizon ads of Serlo need to be added.

elbotho:
closes #280

### DIFF
--- a/src/components/content/horizon.tsx
+++ b/src/components/content/horizon.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 
 import { makePadding } from '../../helper/css'
 
-interface HorizonEntry {
+export interface HorizonEntry {
   title: string
   imageUrl: string
   text: string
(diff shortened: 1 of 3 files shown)


src/components/content/horizon.tsx
```diff
@@ -3,21 +3,22 @@ import styled from 'styled-components'
...
-interface HorizonProps {
+export interface HorizonProps {
   entries: HorizonEntry[]
 }
 
 export function Horizon({ entries }: HorizonProps) {
   return (
     <Wrapper>
-      {entries.map((horizonEntry, index) => {
+      {entries.slice(0, 3).map((horizonEntry, index) => {
```
Entkenntnis, line 21: out of curiosity: is there any case where entries are longer then 3?

elbotho: nope, leftover from when I tried to implement the logic in the component. 
removed now.

PR #298, opened 2020-06-08
try to fix #272

### LINKED ISSUE #272: Weird: wrong images get loaded
In this table the colors are different from the same table in the old design:
https://de.serlo.org/community/hilfe-bearbeitung/richtlinien-inhalten/bilder-grafiken#SerloFarben

elbotho:


elbotho:
please merge if you see no harm :) so we can see if it fixes the problem on 'de.serlo.org'

PR #300, opened 2020-06-08
Type fixing: Components

elbotho:
@inyono components are basically done now.
Nearly all that is left needs definitions of the data types that come in.
Here I need some input on where it would make sense to define and export them, maybe in the next pair session?


elbotho:
`✖ 101 problems (85 errors, 16 warnings)`
for
`yarn eslint "src/components/**/*.{js,jsx,ts,tsx}" --config .eslintrc-strict.js`

elbotho:
yay #300 🥳 🎉

src/components/content/geogebra.tsx
```diff
@@ -5,6 +5,18 @@ export interface GeogebraProps {
...
+  responses: {
+    response: {
+      item: {
+        width: number
+        height: number
+      }
+    }
+  }
+}
+type testData = Response | PromiseLike<Response>
```
inyono, line 18:
```suggestion
type TestData = Response | PromiseLike<Response>
```

src/components/content/geogebra.tsx
```diff
@@ -35,7 +47,9 @@ export function Geogebra({ id }: GeogebraProps) {
       .then((res) => res.json())
       .then((res) => {
         try {
-          const data = res.responses.response.item
+          const _res = res as responseData
+          const data = _res.responses.response.item
```
inyono, line 51: You can add the type to function argument instead:
```tsx
promise.then((res: ResponseData) => {
})    
```

src/components/content/geogebra.tsx
inyono: Where is this used by the way?

src/components/scripts/google-analytics.tsx
```diff
@@ -1,4 +1,5 @@
 export function GoogleAnalytics() {
+  if (process.env.GA_TRACKING_ID === undefined) return null
```
inyono, line 2: Since we add and remove this line pretty much in every other commit: Maybe remove the conditional from `_document` and always render `GoogleAnalytics` since you added the conditional here.

inyono:
> Here I need some input on where it would make sense to define and export them, maybe in the next pair session?

Noice :) We can schedule the next session via Slack!

src/components/content/geogebra.tsx
elbotho: nice, tried some things here and forgot to put it back 👍

src/components/scripts/google-analytics.tsx
elbotho: Are `process.env.GA_TRACKING_ID` and `process.env.NEXT_PUBLIC_GA_TRACKING_ID` identical?

inyono: Ah, this should be `process.env.NEXT_PUBLIC_GA_TRACKING_ID ` everywhere

elbotho: :D

elbotho:
Thanks for the review!

src/components/scripts/google-analytics.tsx
inyono: My bad 😓

elbotho: No, no, just celebrating the good catch :)

PR #302, opened 2020-06-09
Equations Plugin Prototypes

inyono:
Deployment for demo purposes, don't even think about merging this 💩 

https://frontend-git-equations-plugin-prototypes.serlo.now.sh/equations


Entkenntnis:
Closing a PR will keep the branch, so the changes are preserved.

PR #306, opened 2020-06-12
Types for `data` and some refactor

elbotho:
stared with a pair programming session with @inyono 

there's a lot of weird stuff going on still.
Some notes:

- ~~why is `[...slug]` not `[slug]`?~~
- can and should we merge Topic and topic  types into `Entity`?
- …


inyono:
>     * why is `[...slug]` not `[slug]`?

Because the alias (= slug) may contain slashes:
- `[...slug]` matches an arbitrary number of path segments split by `/`
- `[slug]` would only match a single path segment, e.g. `/mathe` but not `/mathe/foo`

src/components/content/exercise.tsx
```diff
@@ -31,33 +67,40 @@ export function Exercise(props: ExerciseProps) {
...
           <ScMcExercise
-            state={state.interactive.state}
+            state={state.interactive.state as ScMcExerciseProps['state']}
             idBase={`ex-${positionOnPage}-${positionInGroup}-`}
           />
         )
       }
       if (state.interactive.plugin === 'inputExercise') {
-        interactiveComp = <InputExercise data={state.interactive.state} />
+        interactiveComp = (
```
inyono, line 90: Nitpicking:
- I'd wouldn't suffix `interactiveComp` with `Comp` since it is not a component (but a React Node)
- I'd usually move that logic into its own function, mostly because I don't like non-const variables that are global to the function but differently modified in the if-else clauses. By moving that into its own function, you can instead forego with the variable altogether.

src/components/content/exercise.tsx
```diff
@@ -31,33 +67,40 @@ export function Exercise(props: ExerciseProps) {
   } = props
   const [solutionVisible, setVisible] = React.useState(false)
 
-  let taskValue = task.children
-  let solutionValue = solution.children
+  let taskValue = {}
+  let solutionValue: unknown[] = []
```
inyono, line 71: is this even used anymore? Saw you referring to `solution.children` directly in the code below.

src/components/content/exercise.tsx
```diff
@@ -84,6 +127,7 @@ export function Exercise(props: ExerciseProps) {
     <Wrapper grouped={grouped}>
       {!grouped && <ExerciseNumbering index={positionOnPage} />}
 
+      {/* @ts-expect-error */}
```
inyono, line 130: Why is this needed?

src/components/content/entity.tsx
```diff
@@ -0,0 +1,247 @@
...
+  ) => {
+    e.preventDefault()
+    setCourseNavOpen(true)
+  }
+
+  // TODO: Should Taxonomy and other content really share a type?
+  if (isTaxonomyTerm(contentType, data)) {
+    return (
+      <main>
+        {/* @ts-expect-error */}
```
inyono, line 110: Why is this needed?

src/components/content/injection.tsx
```diff
@@ -27,10 +27,11 @@ export function Injection({ href }: InjectionProps) {
         else return res.text()
       })
       .then((data) => {
-        if (data.contentType && data.data) {
-          setValue(data.data.value)
-          if (data.data.license) {
-            setLicense(data.data.license)
+        const returnData = data as EntityProps
```
inyono, line 30: Like last time: you can move that to the argument itself: `then((data: EntityProps)`

src/components/navigation/mobile-menu.tsx
```diff
@@ -21,42 +21,64 @@ interface MobileMenuLink {
...
   const { links } = props
+
+  const [openEntry, setOpenEntry] = React.useState(-1)
+
+  function toggle(
+    e: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+    index: number
+  ) {
+    e.preventDefault()
+    if (index === openEntry) setOpenEntry(-1)
```
inyono, line 32: I'd really prefer `number | undefined` or `number | null` over some magical `-1` value. The typing of `number | undefined` (or `null`) makes it obvious that there is a special case I need to handle. On the other hand, I can pass `-1` just fine to any other function / component that accepts `number` (which would most likely be a bug in this case).

src/components/navigation/mobile-menu.tsx
```diff
@@ -21,42 +21,64 @@ interface MobileMenuLink {
...
+          index={index}
+        />
       ))}
     </List>
   )
 }
 
 interface EntryProps extends MobileMenuLink {
-  childKey?: string
+  // childKey?: string
```
inyono, line 52: is this still needed?

src/components/navigation/mobile-menu.tsx
```diff
@@ -21,42 +21,64 @@ interface MobileMenuLink {
...
+  ) => void
 }
 
 function Entry({
   url,
   title,
   icon,
   children,
-  childKey,
+  // childKey,
```
inyono, line 67: same

src/components/navigation/mobile-menu.tsx
```diff
@@ -21,42 +21,64 @@ interface MobileMenuLink {
...
+  onToggle,
+  index,
 }: EntryProps) {
-  const [open, setOpen] = React.useState(false)
   return (
     <>
       <li>
         <EntryLink
-          key={childKey}
+          // key={childKey}
```
inyono, line 77: same

src/components/navigation/mobile-menu.tsx
```diff
@@ -85,7 +107,7 @@ function Entry({
       {open && children ? (
         <>
           {children.map((entry, index) => (
-            <Entry {...entry} isChild key={`${index}--${childKey}`} />
+            <Entry {...entry} isChild key={`${index}`} />
```
inyono, line 110:
```suggestion
            <Entry {...entry} isChild key={index} />
```

src/components/slug-head.tsx
```diff
@@ -1,30 +1,37 @@
 import Head from 'next/head'
 
+import { EntityProps } from './content/entity'
+
 //TODO: define and export data types somewhere
```
inyono, line 5:
```suggestion
```

src/components/slug-head.tsx
```diff
@@ -1,30 +1,37 @@
...
+  alias,
+}: SlugHeadProps) {
   function getMetaContentType() {
     //match legacy content types that are used by google custom search
     if (contentType === undefined) return ''
     if (contentType === 'Exercise') return 'text-exercise'
     if (contentType === 'CoursePage') return 'course-page'
-    if (data.data?.type === 'topicFolder') return 'topic-folder'
+    /* @ts-expect-error */
+    if (data.type === 'topicFolder') return 'topic-folder'
```
inyono, line 27: Why is this needed? Seems like either a bug here or in the types.

src/pages/[...slug].tsx
```diff
@@ -59,91 +70,55 @@ function PageView(props: PageViewProps) {
...
-                  {renderArticle(data.data.value.children)}
-                </>
-              )}
-              {(contentType === 'Exercise' ||
-                contentType === 'ExerciseGroup') && (
-                <>{renderArticle(data.data.value.children)}</>
-              )}
-              {license && <LicenseNotice data={license} />}
-            </main>
+            {/* TODO: put this part into entity.tsx */}
```
Comment (inyono) [inline src/pages/[...slug].tsx:102]: Is this comment still needed? I don't know what it refers to.

src/pages/[...slug].tsx
```diff
@@ -189,20 +159,29 @@ const StyledIcon = styled(FontAwesomeIcon)`
 // -> You can not use getInitialProps with getServerSideProps. Please remove getInitialProps. /[...slug]
 
 // TODO: needs type declaration
-export const getServerSideProps: GetServerSideProps<any, any> = async (
-  props
-) => {
+export const getServerSideProps: GetServerSideProps = async (props) => {
```
Comment (inyono) [inline src/pages/[...slug].tsx:162]: This still needs a type, right? Because `props` is implicitly `any` here when you don't specify the type arguments in `GetServerSideProps`.

src/schema/article-renderer.tsx
```diff
@@ -40,6 +40,8 @@ import type { GeogebraProps } from '@/components/content/geogebra'
 import type { InjectionProps } from '@/components/content/injection'
 import type { MathProps } from '@/components/content/math'
 import type { VideoProps } from '@/components/content/video'
+// eslint-disable-next-line import/extensions
+import { EditorState } from '@/pages/[...slug]'
```
inyono, line 44: @inyono Looks like a bug in the ESlint config.

src/schema/article-renderer.tsx
```diff
@@ -64,7 +66,7 @@ const Code = dynamic<CodeProps>(() =>
 )
 
 // TODO: this is probably not the correct type.
```
inyono, line 68:
```suggestion
```
or 
```suggestion
// TODO: this is probably the correct type.
```
:)

.eslintrc.js
```diff
@@ -119,4 +120,12 @@ module.exports = {
...
     },
   },
+  overrides: [
+    {
+      files: ['src/pages/*'],
+      rules: {
+        'import/no-default-export': 'off',
+      },
+    },
+  ],
```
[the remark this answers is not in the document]
elbotho: only way I found that worked :)

.eslintrc.js
```diff
@@ -119,4 +120,12 @@ module.exports = {
       version: 'detect',
     },
   },
+  overrides: [
+    {
+      files: ['src/pages/*'],
```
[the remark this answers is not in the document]
elbotho: wasn't sure if this should apply to the files in pages/api/frontend as well but now I understand what they do :)

src/components/navigation/mobile-menu.tsx
elbotho: check, will keep that in mind.

src/pages/[...slug].tsx
Comment (elbotho) [reply src/pages/[...slug].tsx:162]: currently props has this type:
```
(parameter) props: GetServerSidePropsContext<ParsedUrlQuery>
```
not sure.

src/schema/article-renderer.tsx
elbotho: Full error:
`Unexpected use of file extension "slug]" for "@/pages/[...slug]"eslintimport/extensions`

elbotho: that's why I asked about the `...slug` above

elbotho:
@inyono thank you for the in-depth review! 
I addressed most of it and commented the rest so we can find it tomorrow, also merged in master again.

PR #323, opened 2020-06-19
Another type session

elbotho:
No more errors in `/pages`, `/components`, `/helper`, `/data`.
(used two `// @ts-expect-error` comments in because of `EditorState` problems)

Remaining:
- `fetcher`: 319 Problems
- `schema`: 621 Problems

### DIFF
--- a/src/components/content/entity.tsx
+++ b/src/components/content/entity.tsx
@@ -1,12 +1,4 @@
-import {
-  faGraduationCap,
-  faNewspaper,
-  faPlayCircle,
-  faCubes,
-  faCircle,
-  faFolderOpen,
-  faShareAlt,
-} from '@fortawesome/free-solid-svg-icons'
+import { faShareAlt } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import dynamic from 'next/dynamic'
 import React from 'react'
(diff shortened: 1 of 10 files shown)


src/components/content/entity.tsx
```diff
@@ -172,40 +165,15 @@ export function Entity({ data, contentId, contentType, license }: EntityProps) {
...
-        break
-      case 'Applet':
-        icon = faCubes
-        iconTitle = 'Applet'
-        break
-      case 'TaxonomyTerm':
-        icon = faFolderOpen
-        iconTitle = 'Bereich'
-    }
+    const iconAndTitle = getIconAndTitleByContentType(contentType)
```
inyono, line 168: Is there a reason that this isn't a Component `<TitleWithIcon contentType={contentType} />`?

src/helper/header-by-content-type.ts
```diff
@@ -0,0 +1,52 @@
...
+    case 'subfolders':
+      icon = faFolderOpen
+      title = 'Bereich'
+      break
+    case 'exercises':
+      icon = faFile
+      title = 'Aufgaben'
+  }
+
+  return { icon: icon, title: title }
```
inyono, line 51:
```suggestion
  return { icon, title }
```

src/components/content/entity.tsx
elbotho: Hmm, I guess. Topic uses them in a different way (h4 instead of h1) other style so I thought this is more flexible since we might also just load the icon in some places.

elbotho:
ui exciting, first merge, let's see it all go up in flames 🔥

src/components/content/entity.tsx
inyono: Fine by me :)

Issue #308, opened 2020-06-15
add client-side-navigation to menus


Entkenntnis:
Client-side-navigation is not yet implemented for:

- main menu
- secondary menu
- course menu
- footer

Not all links there should be routed client-side, though (like legacy-links ...)

elbotho:
maybe not exclusive, but here are some of the urls that we should not be using client side navigation for:
https://github.com/serlo/serlo.org-cloudflare-worker/blob/master/src/static-pages/config.ts
https://github.com/serlo/serlo.org-cloudflare-worker/blob/master/src/frontend-proxy.ts
https://github.com/serlo/serlo.org-cloudflare-worker/blob/master/src/index.ts

inyono:
Possible approaches:
- We wrap the API call in the Cloudflare Worker and possibly pass the info that this should be handled via legacy system.
- We expose the logic in the Cloudflare Worker as an npm package and import that in the frontend (problem: need to keep that package up-to-date)
- Duplicate the logic from Cloudflare Worker (meh)

I'd prefer the first approach to be honest. This way, we'd only need to redeploy the Cloudflare Worker when the config changes and we can wrap the client-side logic once on front-end side.

elbotho:
@Entkenntnis and @kulla what do you think?

Entkenntnis:
We can split this issue into two parts: static and dynamic links. For static links, we can set manually set the navigation state and this would work for menu and footer (a first attempt here: #317)

For dynamic links (secondary nav, course nav), another possibility would be to use the same logic as the pretty link part and only add client-side-navigation to the links that can be prettyfied.

Entkenntnis:
#317 and #319 add client-side-navigation to header, footer and courses.

Only remaining menu is secondary navigation, this has a separate issue now: #320

kulla:
> Possible approaches:
> 
>     * We wrap the API call in the Cloudflare Worker and possibly pass the info that this should be handled via legacy system.
> 
>     * We expose the logic in the Cloudflare Worker as an npm package and import that in the frontend (problem: need to keep that package up-to-date)
> 
>     * Duplicate the logic from Cloudflare Worker (meh)
> 
> 
> I'd prefer the first approach to be honest. This way, we'd only need to redeploy the Cloudflare Worker when the config changes and we can wrap the client-side logic once on front-end side.

@inyono How would the first solution look like? Can we talk about this next monday?

inyono:
[...]

PR #325, opened 2020-06-19
OAuth Token Workflow

### LINKED ISSUE #290: show logged-in state in the menu
For regular users, show: Anmelden
For logged-in users, show: Notification-Icon and Abmelden

We want to use some kind of "two-pass-rendering" to show this on the client without affecting ssr too much. We can start working on this when #289 is in progress.

inyono:
Adds the basic login / logout workflow with Hydra. Doesn't work in preview deployments since Hydra assumes that frontend runs on serlo-staging.dev resp. serlo.org (and we still redirect logged in users to the old system). To test locally:

- [Start the local dev environment for serlo.org](https://github.com/serlo/serlo.org#start).
- Run `yarn oauth:frontend` to create the OAuth client for the frontend in Hydra.

**TODO**
- [x] The Menu is only rendered on the client (as we discussed: we don't want to differentiate depending on auth state on server because of caching reasons). We might want to add an animation for that or something
- [x] Handle errors in `pages/api/auth` more gracefully @inyono
- [x] Login Workflow: after login, the page `src/pages/auth/login` is displayed for 5 seconds and then redirects to the previous page. We need to decide what to do here @elbotho (e.g. we could instead show a success notification and redirect directly)
- [x] Logout Workflow: after logout, the page `src/pages/auth/logout` is displayed for 5 seconds and then redirects to the previous page. We need to decide what to do here @elbotho (e.g. we could instead show a success notification and redirect directly)
- [ ] Logout should also clear the cookie in the legacy system @inyono ping https://github.com/serlo/serlo.org/issues/453
- [x] Use feature flag to disable this in production.


src/pages/api/auth/callback.ts
```diff
@@ -0,0 +1,42 @@
...
+    state: string
+  }
+  const { csrf, referer } = JSON.parse(state)
+  if (csrf === req.cookies['auth-csrf']) {
+    const result = await oauth2AuthorizationCode.getToken({
+      code,
+      redirect_uri: 'http://localhost:3000/api/auth/callback',
+      scope,
+    })
+    const { token } = oauth2ClientCredentials.createToken(result)
```
Entkenntnis: Just wondering: why are we not using `oauth2AuthorizationCode.createToken()` here?

inyono: Intended, the types are not up-to-date and this should be `clientCredentials` stuff

src/pages/api/auth/callback.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { code, scope, state } = req.query as {
+    code: string
+    scope: string
+    state: string
+  }
+  const { csrf, referer } = JSON.parse(state)
+  if (csrf === req.cookies['auth-csrf']) {
+    const result = await oauth2AuthorizationCode.getToken({
+      code,
+      redirect_uri: 'http://localhost:3000/api/auth/callback',
```
inyono: @inyono this is wrong

inyono:
Result from the discussion with @elbotho:
- Login / logout workflow is most likely going to be a flasher
- @elbotho does the first iteration
- @inyono fixes remaining bugs in the auth workflow & tries to get this working on staging

elbotho:
Next router reloads page if we use it like this
``` tsx
void router.replace(props.referer.replace(document.location.origin, ''))
```
Instead to have a client side redirect we would need the same logic we use with Links:
``` tsx
<NextLink href="/[...slug]" as={decodeURIComponent(prettyLink)}>
```
(took me ages to figure out why the context was not persisting…)

So for now I'm using session storage to display the notification-toast, but's it's a bit of a workaround.
Good enough for now I'd say.
I keep the code in login and logout and see how https://github.com/serlo/frontend/issues/328 and https://github.com/serlo/frontend/issues/308 work out.

inyono:
Okay, this still doesn't work correctly since we have multiple redirect URIs and then OAuth requires a [sector validation uri](https://openid.net/specs/openid-connect-registration-1_0.html#SectorIdentifierValidation). Continuing with this another time -.-

elbotho:
worked on the auth userflow again:

simplified it for now:
- no provider for toast
- redirect to referrer directly
- remove pages for /auth/login and logout
- use hash to display notification

feels a bit faster/nicer (no white page flash), code is quite clean and we can still expand it.

inyono:
Okay, going to look into that, thanks :)

src/components/navigation/header.tsx
```diff
@@ -1,29 +1,35 @@
...
     <BlueHeader>
       <MobileMenuButton onClick={() => setOpen(!isOpen)} open={isOpen} />
       <PaddedDiv>
-        <Menu links={menuData}></Menu>
+        <Menu links={links} auth={auth} />
         <Logo subline="Die freie Lernplattform" />
       </PaddedDiv>
       <SearchInput />
-      {isOpen && <MobileMenu links={menuData} />}
+      {isOpen && <MobileMenu links={links} />}
```
inyono: Mobile menu is missing auth stuff.

src/components/toast-notifications.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import Notification, { notify } from 'react-notify-toast'
+
+import { useAuth } from '@/auth/use-auth'
+import { theme } from '@/theme'
+
+export function ToastNotifications() {
```
inyono, line 7: This should probably use `ToastContext` in the future, right?

elbotho: it used to already, but I opted to use the simplest implementation for now and add in context etc. back in later when I understand the requirements of other notifications.

package.json
```diff
@@ -34,7 +34,7 @@
     "@sentry/browser": "^5.15.4",
     "@tippyjs/react": "^4.0.0-alpha.0",
     "@types/nprogress": "^0.2.0",
-    "@types/react-notify-toast": "^0.5.2",
+    "@types/react-notify-toast": "^0.5.0",
```
elbotho, line 38: @inyono did you downgrade this on purpose?

src/components/toast-notifications.tsx
```diff
@@ -16,8 +16,8 @@ export function ToastNotifications() {
 
   React.useEffect(() => {
     if (window.location.hash === '#auth') {
-      //leave no trace
-      history.pushState(null, '', window.location.href.split('#')[0])
+      // Remove auth URI fragment
+      history.replaceState(null, '', window.location.href.split('#')[0])
```
elbotho, line 20: thanks!

src/components/navigation/header.tsx
elbotho: right, will there be useful functions for logged in mobile users? 
commenting I guess.

src/components/toast-notifications.tsx
inyono: Currently, `ToastContext` is not in use and you only handle auth notifications.

package.json
inyono: Yes. My dependency workflow:
- Add the dependency with `^major.0.0` (resp. `^0.minor.0` for 0.x versions). If you need a specific feature (e.g. react hooks), specify the version that introduced the feature (i.e. `^16.8.0`).
- Use `yarn upgrade` to upgrade dependencies (instead of updating the version in `package.json`).

Reasoning: When some dependencies don't work together for some reason, we can look into `package.json` to see our "minimal" requirement for the dependency (and therefore know which dependencies we could downgrade).

So in this case, it's not an actual downgrade, since `yarn.lock` still contains the newest version. But it documents that we don't actually need `0.5.2`.

src/components/navigation/header.tsx
inyono: Yes. With the new editor it should also be possible to make small edits on mobile. And any learner features that we introduce in the future.

src/components/toast-notifications.tsx
elbotho: yes.

elbotho: removed unused file.

src/components/navigation/header.tsx
elbotho: My local auth setup is broken again, so I can't test thoroughly. 
But MobileMenu should now have the same functionality. It's a bit easier because we render the menu only when the user open it.

inyono:
Auth now works correctly in staging (after setting `frontendDomain` to `frontend-git-auth.serlo.vercel.app`).

Entkenntnis:
@inyono This is ready so far, auth is disabled for production (serlo.org), everything else looks fine.

Issue #329, opened 2020-06-20
Refactor Links


elbotho:
We now have some duplication and slightly different implementations of links for content and the different menus.

I'd love a smart Link component that handles PrettyLinks and can decide if a link should be routed client-side or with reload.
Related: #320, #308, #328 …

Entkenntnis:
Love this idea very much!

elbotho:
Discussion with @Entkenntnis

Todo: 
- Also check for absolute links in content (`de.serlo.org/[id]`) and use CSR for them
- Check error handling if a CSR fails (could it be fixes with a redirect?)

Currently there are following exceptions to client-side-routing:
- external links
- legacy pages (like auth and privacy)
- frontend pages that are not using `[...slug]` (index, search, donations)

Results:
- Legacy stuff should be handled by cloudflare worker to avoid duplication etc. like @inyono suggested
- Idea for the frontend pages (index, search, donations): Also use `[...slug]` and dynamically load to load those. This way we can use CSR for all links in the frontend that are not external or legacy.

elbotho:
this is currently the result when I try client side routing with an absolute link
![image](https://user-images.githubusercontent.com/1258870/85285202-c44ea480-b490-11ea-9daa-676b79009ecd.png)

elbotho:
result when I force CSR to (`/auth/login`)
``` js
{"error":
    "Error while fetching data: Cannot read property 'path' of null:
    {
    "response":{
    "errors":[{
    "message":"Cannot read property 'path' of null","locations":[{
    "line":3,"column":5}],"path":["uuid"],"extensions":{
    "code":"INTERNAL_SERVER_ERROR"}}],"data":{
    "uuid":null},"status":200},"request":{
        …
        }
    }
}
```

CSR to `/impressum` etc actually load the content in the redesign.

elbotho:
I found a `ref/21654` link that breaks. What's the deal with those, how should we handle them?

inyono:
`ref/<id>` is equivalent to `/<id>`. We don't really use them anymore (and while there probably was a reason for the ref links at some point, we encourage users to use `/</id`).

PR #339, opened 2020-06-22
Refactor links, client route all the things

elbotho:
See #329 for the plan.

- also use csr for absolute serlo links (e.g. https://de.serlo.org/1565)
- use [experimental feature](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes) for catch all routes with `[[...slug]]` even "/"
- move all pages into `[[...slug]]` (index, donations, search)
- use `<Link>` wherever csr might occur.
- let `<Link>` decide when to use csr
- currently uses a static array of legacyLinks to deactivate csr for those routes, those should be provided by the cloudflare worker in the future

PS: It's so much fun to navigate now, client side cache and routing is awesome 🔥


Entkenntnis:
One small bug: clicking "Fächer" in the mobile menu is not opening the submenu, but lead to navigation instead.

src/components/content/link.tsx
```diff
@@ -6,35 +6,74 @@ import { StyledA } from '../tags/styled-a'
...
+  return renderLink()
+
+  function getPrettyLink(href: string): string | undefined {
+    if (prettyLinks === undefined || prettyLinks === {}) return undefined
+
+    const prettyLink = prettyLinks[href.replace('/', 'uuid')]?.alias
+    if (prettyLink !== undefined) return prettyLink
+
+    //fallback for wrong absolute links
+    return prettyLinks['uuid' + href.split('de.serlo.org/')[1]]?.alias
```
inyono, line 60: I'd probably move that together with the Context into its own hook `const prettyifyLink = usePrettyLink()` (or something). This way, you have the logic in one place, we could unit test it and possibly include it somewhere else (though a single Link component is probably enough for the last point).

inyono: Also, the whole `isAbsolute` handling. Would be pretty nice if we have a hook that accepts a whatever url and tells you if it can be rendered client side and which href to use.

src/data/footer.tsx
```diff
@@ -7,10 +7,10 @@ export const footerNavEntries = [
...
-      { title: 'Partner & Förderer', url: '/partner', clientside: true },
-      { title: 'Presse', url: '/presse', clientside: true },
-      { title: 'Kontakt', url: '/kontakt', clientside: true },
+      { title: 'Über Serlo', url: '/serlo' },
+      { title: 'Partner & Förderer', url: '/partner' },
+      { title: 'Presse', url: '/presse' },
+      { title: 'Kontakt', url: '/kontakt' },
       {
         title: 'Serlo in anderen Sprachen',
         url: `https://en.${serloDomain}.org/global`,
```
inyono, line 16: Should be `https://en.${serloDomain}/global`

src/components/content/link.tsx
elbotho: I guess I get your point (although I never really worked with unit tests before), but isn't this what this refactor did?
The only place where PrettyLinksContext is consumed and checked currently is this component (same for the `isAbsolute` stuff).

elbotho: If you still think it's useful I'll go and read about custom hooks :)

PR #342, opened 2020-06-25
fix @ts-expect-error

### LINKED ISSUE #341: replace ts-expect-error with any?
In `src/content/exercise.tsx`, we have two ts-expect-error that show up as warnings. Can we use `as any` here to avoid the warnings?

Or is the warning expected here?

Entkenntnis:
fixes #341


inyono:
Please leave the annotation as it is for now, since this indicates a place where the typing is not correct / complete yet.

PR #348, opened 2020-06-29
Test Web Vitals

elbotho:


Entkenntnis (approved):
looks good, let's see what the data looks like.

Issue #347, opened 2020-06-29
Idea: Use service worker for caching


elbotho:
This might actually be a useful thing (offline access etc) and would be a big step towards PWA.

elbotho:
Might be relevant:
https://github.com/shadowwalker/next-pwa/

inyono:
Good idea. Would put that into the backlog for now, though, since this adds more complexity and I'd rather get client-side routing / other critical features working before we add this. Might also be nice to have a testing suite before that. What's your opinion on that?

elbotho:
Jup, backlog is good. 100% is more important than new (experimental) features.
The "session-cache" works surprisingly well.

Entkenntnis:
I will close this issue as we already have an issue for this topic in product management: https://github.com/serlo/serlo.org-rfcs/issues/39

PR #352, opened 2020-06-29
Add cache to injections

elbotho:
Without refactor for now

### DIFF
--- a/src/components/content/injection.tsx
+++ b/src/components/content/injection.tsx
@@ -34,6 +34,7 @@ export function Injection({ href }: InjectionProps) {
       const cachedData = sessionStorage.getItem(encodedHref)
       if (cachedData) {
         dataToState(JSON.parse(cachedData))
+        return
       }
     } catch (e) {
       //
--- a/src/components/content/injection.tsx
+++ b/src/components/content/injection.tsx
@@ -28,27 +28,47 @@ export function Injection({ href }: InjectionProps) {
   useEffect(() => {
     const origin = window.location.host
     const protocol = window.location.protocol
-    void fetch(
-      `${protocol}//${origin}/api/frontend${encodeURI(
-        href.startsWith('/') ? href : `/${href}`
-      )}`
-    )
+    const encodedHref = encodeURI(href.startsWith('/') ? href : `/${href}`)
+
+    try {
+      const cachedData = sessionStorage.getItem(encodedHref)
+      if (cachedData) {
+        dataToState(JSON.parse(cachedData))
+      }
+    } catch (e) {
+      //
+    }
+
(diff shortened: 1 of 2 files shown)


src/components/content/injection.tsx
```diff
@@ -28,27 +28,47 @@ export function Injection({ href }: InjectionProps) {
...
-      `${protocol}//${origin}/api/frontend${encodeURI(
-        href.startsWith('/') ? href : `/${href}`
-      )}`
-    )
+    const encodedHref = encodeURI(href.startsWith('/') ? href : `/${href}`)
+
+    try {
+      const cachedData = sessionStorage.getItem(encodedHref)
+      if (cachedData) {
+        dataToState(JSON.parse(cachedData))
```
Entkenntnis, line 36: not returning here meanings that we are still doing a request in the background, that is fine if we ...

src/components/content/injection.tsx
```diff
@@ -28,27 +28,47 @@ export function Injection({ href }: InjectionProps) {
...
-      .then((data: EntityProps) => {
-        if (data.contentType && data.data) {
-          setValue(data.data.value)
-          if (data.data.license) {
-            setLicense(data.data.license)
-          }
-          if (data.prettyLinks) {
-            setPrettyLinks(data.prettyLinks)
+      .then((fetchedData: EntityProps) => {
+        dataToState(fetchedData)
```
Entkenntnis, line 48: ... only set data when the injection is not loaded or data is different?

inyono: Is it, though? If we always do the request anyways, there's not much point in caching, is there? We'd need at least a lifetime for the local cache, e.g. only fetch the data again if the data is older than 10 minutes or whatever.

elbotho: the point of caching is immediate response without waiting/back-navigation hickups. it would even make sense if the do the request anyways. 
but to be honest I just forgot the `return` 😄

inyono: Okay, true. Still: we'd should optimize the requests, too, so that our site is also fast for mobile users (where additional requests would make other requests - where we don't have the data in cache - slower and waste mobile data).

src/components/content/injection.tsx
Entkenntnis: just return here :)

src/components/content/injection.tsx
elbotho: @inyono totally agree, my point was, that the cache makes sense, not that unnecessary requests do.

Let's talk about this again when we build a service worker (#347) or when we think about the session-cache for logged in users (edits and whatnot).

elbotho: (missing return is added in ca3f796bc9a60628ca4dc1b58045c9555953ecc6

PR #356, opened 2020-07-01
Lazy load Images, Videos, …

elbotho:
Using [`react-lazyload`](https://github.com/twobin/react-lazyload) to improve page load times.
Works fine so far. 
Active for Images, Videos, MathWrapper, Horizon and Geogebra

Possible drawbacks:
- SEO: Google Bot does not scroll an might not never those elements. ([workaround1](https://github.com/aFarkas/lazysizes/issues/215#issuecomment-171259449), [workaround2](https://github.com/twobin/react-lazyload/issues/109#issuecomment-438723391))
- Layout might jerk/reflow on slow connections/devices

**Results**
Hard to test because de.serlo.org is much faster overall, but the numbers look promising. 
For my benchmark page requests are down from 43 to 26 and transferred size is down from 693kb to 562kb.


Entkenntnis:
Nice, I am barely noticing any loading :+1:

Very interesting idea for an SEO workaround, here is another: https://github.com/twobin/react-lazyload/issues/109, both are fine ...

inyono:
We should definitely consider SEO. How bad is it, though, when Google doesn't see media content? Do they do anything interesting with what or do they look only at `alt` attributes anyway?

elbotho:
I really want serlo images to show up in the google image search, especially if you specify freely licensed images.
but not sure what they currently do to images while crawling.

elbotho:
([relevant line for workaround 1](https://github.com/aFarkas/lazysizes/blob/7bd825c55d96b02d18d55931af1afdfe323155fb/src/lazysizes-core.js#L291))

inyono:
Okay, that makes sense, didn't think about that. Maybe [Google's URL Inspection Tool](https://support.google.com/webmasters/answer/9012289) helps to answer the question?

elbotho:
the noscript version seems to be the cleaner and also recommended version for google. 
don't know about other bots, but I'll add that now.
@inyono good idea!

PR #372, opened 2020-07-07
Refactor InitialProps and prepare for i18n

Entkenntnis:
This PR contains a lot of (overdue) code maintenance work. There are still some refactorings to do, but because this PR is already quite big I would like to tackle the remaining changes in another PR and merge this after #369.

**Changes:**

- Complete and revised types for InitialProps in `src/data-types.ts`. This file defines all data that the client-side components use and unites them in one place. It also contains some commentary. (https://github.com/serlo/frontend/pull/372/files#diff-654162a61df67291a611f06b7320a968)
- All language-specific data are now describes as *InstanceData*. German-specific data are all stored in `src/data/de.ts` (includes header menu, footer menu and ui strings). This is a preparation for i18n. (https://github.com/serlo/frontend/pull/372/files#diff-df3c4d366ee4ac55d8f416e3c1125f20)
- While revising the client data model, I have moved business logic into the fetcher, especially meta-data generation, breadcrumbs shortening, secondary navigation active, course navigation active and next link, visibility of breadcrumbs, navigation, edit button, ...
- PrettyLinks are now directly resolved in the fetcher.
- Complete restructuring of topic (no need to hassle with TopicLinkList and TopicPurpose)
- New context for the origin to access the frontend domain.
- Fixed buggy behaviour of cookie bar.
- Moved components out of _app and [[...slug]]
- Added a separate error page.
- Moved getInitialProps into separate file
- Introduced the concept of *Category*, which are the different types visible in the taxonomy.

**Follow-Ups:**

- Check for regression.
- Now all types are defined, use them in the fetcher and remove TODOs.
- Add types for Injections.
- Check if all pretty links are properly resolved.


src/fetcher/query.ts
```diff
@@ -1,400 +1,549 @@
...
-        }
-        ... on TaxonomyTerm {
-          alias
-        }
-      }
-  `
-  )
-  return `{${query}}`
+// Keep this file in sync with the graphQL schema.
+// Maybe automate this one day.
```
inyono, line 2: Haven't used this yet, but https://github.com/apollographql/apollo-tooling seems to be able to generate TypeScript types from a GraphQL query.

src/fetcher/query.ts
```diff
@@ -1,400 +1,549 @@
...
-      ... on Entity {
-        alias
-      }
-      ... on Page {
-        alias
-      }
-      ... on TaxonomyTerm {
-        alias
-      }
+// This is one breadcrumb path. Why is only url optional?
```
inyono: ?

inyono:
Thanks for approaching this. Let me know if you need any input from API side.

src/fetcher/query.ts
Entkenntnis: grafik

Label is required, but url is not. This means we can have navigation nodes with label, but without url.

inyono: ~Basically: every navigation node has a required `label` and either `url` or `id`. `url` would be set if this is a custom route, e.g. `/auth/login`. `id` would be set if it is an Uuid (and `/<id>` resp. `/<alias>` could be used as an url).~

Scratch that, the API already polyfills the url for ids, see https://github.com/serlo/api.serlo.org/blob/master/src/graphql/data-sources/serlo.ts#L140. The only case were `url` is null if it is an Dropdown menu. So `url` should only be null if it also has children. I'm not sure if we have this case anymore since Wolfgang simplified the navigation (at least in Instance `de`).

Entkenntnis: I am a bit confused. As far as I understand, NavigationNode is an entry in the breadcrumbs, which has no Dropdown functionality? The secondary navigation, that contained nested elements, is stored separately in navigation.data (and is passed down as JSON)

So I think, in this case `url` can be made required??

inyono: The `NavigationNode` in breadcrumbs is just another perspective on the navigation & taxonomy term data (see below for a detailed explanation). So if the currently viewed page is a child of a dropdown, the dropdown element will appear in the breadcrumbs too, e.g.
```
Mathematik
  > Some Dropdown
     > Some Page
```

Entkenntnis: How is this related to the taxonomy? I thought that breadcrumbs are only using paths from the taxonomy (and please don't tell me that the taxonomy is just another perspective of the navigation :smile:)

inyono: Basically: the Taxonomy is part of the navigation and the breadcrumbs may contain information from both navigation and taxonomy tree. See e.g. for uuid 1855, the breadcrumbs are:
```
Mathematik (Navigation element that points to the math landing page /19767)
> Alle Themen (Navigation element that points to the math subject /5)
> Funktionen (TaxonomyTerm)
> Wichtige Funktionstypen und ihre Eigenschaften (TaxonomyTerm)
> Quadratische Funktionen - Parabeln (TaxonomyTerm)
> Funktionsterm einer quadratischen Funktion (TaxonomyTerm)
```
So basically, we glue together the navigation data to the subject and the taxonomy term path to the entity. The actual navigation data stops at "Alle Themen".

For pages, this is simpler (since they are not part of the navigation and therefore we only check if they are part of the navigation itself).

inyono: So related to the question you had above: the dropdown case should only occur for pages, since the subject taxonomy terms shouldn't be a child of a dropdown.

Entkenntnis: Thank you for the explaination, this makes much more sense now. So the takeaway is that the breadcrumbs are a combination of navigation *and* taxonomy - and the navigation nodes *could be* dropdowns without url.

inyono: Exactly, totally obvious /s 😛 (yes, we love our legacy system)

PR #373, opened 2020-07-09
Auth: implement logout workflow

inyono:
Depends on https://github.com/serlo/serlo.org/pull/456, marking as ready for review as soon as this is deployed to staging.


inyono:
Works as expected :)

PR #379, opened 2020-07-09
revert change that causes breadcrumbs to disappear

### LINKED ISSUE #377: Breadcrumbs are missing
Breadcrumbs are missing from articles. Strange, will investigate.

Entkenntnis:
fixes #377

### DIFF
--- a/src/fetcher/create-breadcrumbs.tsx
+++ b/src/fetcher/create-breadcrumbs.tsx
@@ -1,6 +1,5 @@
 // TODO: needs type declaration
 export function createBreadcrumbs(uuid: any) {
-  if (!uuid || !uuid.navigation || !uuid.navigation === undefined) return []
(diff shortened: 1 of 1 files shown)


src/fetcher/create-breadcrumbs.tsx
```diff
@@ -1,6 +1,5 @@
 // TODO: needs type declaration
 export function createBreadcrumbs(uuid: any) {
-  if (!uuid || !uuid.navigation || !uuid.navigation === undefined) return []
```
inyono, line 3: `!uuid.navigation === undefined` doesn't make sense since you already check `!uuid.navigation`. Also, the GraphqL API should return `null` for option values (not `undefined`)

Entkenntnis: That's the reason while I deleted this line, we check for navigation later on.

PR #384, opened 2020-07-13
Fix profile links

elbotho:
- currently frontend uses client side routing for profiles and fails to open them.
- might also solve #363

### DIFF
--- a/src/components/content/link.tsx
+++ b/src/components/content/link.tsx
@@ -55,8 +55,9 @@ export function Link({
     displayHref.startsWith('/user/profile/') ||
     displayHref.startsWith('user/profile/')
 
-  if (isExternal || (isAbsolute && prettyLink === undefined))
+  if (!displayHref || isExternal || (isAbsolute && prettyLink === undefined))
     return renderLink()
+
   if (!isLegacyLink || prettyLink) return renderClientSide()
 
   //fallback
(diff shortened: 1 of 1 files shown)


src/components/content/link.tsx
```diff
@@ -50,10 +50,14 @@ export function Link({
 
   const isAbsolute = href.indexOf('//') > -1
   const isExternal = isAbsolute && !href.includes('.serlo.org')
-  const isLegacyLink = legacyLinks.indexOf(displayHref) > -1
+  const isLegacyLink =
+    legacyLinks.indexOf(displayHref) > -1 ||
+    displayHref.startsWith('/user/profile/') ||
+    displayHref.startsWith('user/profile/')
```
inyono, line 56: Why could both of these cases occur?

inyono (commented):
Good catch, we recently fixed this in the Cloudflare Worker, too, https://github.com/serlo/serlo.org-cloudflare-worker/commit/83430fa86e671b9211240ed769bbd373bdaa398c (didn't think about CSR though). We definitely should sync the logic soon-ish.

src/components/content/link.tsx
elbotho: because authors… and currently we don't clean up the link for this check. 
since this is a temporary solution I'd keep it this way.

inyono: Welp, that wouldn't work correctly anyways with our alias system ^^'. But okay, if it's the actual href anyways, I'm fine with having this as a quick fix until we sync this.

inyono (approved):
I'm also fine with deploying this to production directly.

Entkenntnis (approved):

PR #387, opened 2020-07-13
Deploy staging to production (week 29)

Entkenntnis:
Merge scheduled next monday morning.


src/components/content/link.tsx
```diff
@@ -50,10 +50,14 @@ export function Link({
...
   const isAbsolute = href.indexOf('//') > -1
   const isExternal = isAbsolute && !href.includes('.serlo.org')
-  const isLegacyLink = legacyLinks.indexOf(displayHref) > -1
+  const isLegacyLink =
+    legacyLinks.indexOf(displayHref) > -1 ||
+    displayHref.startsWith('/user/profile/') ||
+    displayHref.startsWith('user/profile/')
 
-  if (isExternal || (isAbsolute && prettyLink === undefined))
+  if (!displayHref || isExternal || (isAbsolute && prettyLink === undefined))
```
inyono: This looks wrong since the check is after we already called `displayHref.startsWith` (which will fail if `displayHref` is not defined)

inyono (changes requested):

src/components/content/link.tsx
inyono: Also: if this might be falsy for some reason, please adapt the types so that we know what we need to check that case. Currently, `displayHref` seems to be always defined according to the types.

Entkenntnis: alright, the check as unnecessary, I have removed it.

inyono (approved):

PR #388, opened 2020-07-13
CSR and search

### LINKED ISSUE #350: Search stops working
when I use the search, then use client side routing (e.g. click in the menu) the searchbar does not respond any more.

**Steps:**
- navigate to any page (e.g. https://de.serlo.org/serlo)
- enter term in search bar (search results load)
- now click on e.g. the logo (client side routing to https://de.serlo.org/ happens)
- now I enter another term into the search bar
-> url changes, but results don't show

### LINKED ISSUE #351: Use client side routing for search results
We could catch clicks on the search results and then use next router to redirect

elbotho:
As expected the hackiest component get's some hacky fixes.

We had some problems with the search since we moved to client side routing. 
This PR addresses them (fixes #350) and also fixes history-navigation for search (mostly).

Also now the links in the search results use CSR as well (#351) and opens the results in the current environment (-> stays on localhost instead of navigating to de.serlo.org).

(downside: seems a bit slower, and there are more white flashes, but it's worth it I'd say :) )

### DIFF
--- a/src/components/navigation/search-input.tsx
+++ b/src/components/navigation/search-input.tsx
@@ -24,13 +24,13 @@ It's a bit hacky, but it's free and works quite well.
 export function SearchInput({ onSearchPage }: SearchInputProps) {
   const [searchLoaded, setSearchLoaded] = React.useState(false)
   const [searchActive, setSearchActive] = React.useState(false)
-  const [isSearchPage, setIsSearchPage] = React.useState(false)
+  // const [isSearchPage, setIsSearchPage] = React.useState(false)
   const { strings } = useInstanceData()
 
   React.useEffect(() => {
     // note: find a better way to tell search input that it should activate itself
     if (onSearchPage) {
-      setIsSearchPage(true)
+      // setIsSearchPage(true)
       activateSearch()
     }
     // I only want to run this the first time the page loads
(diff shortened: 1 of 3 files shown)


src/components/navigation/search-input.tsx
```diff
@@ -65,6 +65,32 @@ export function SearchInput({ onSearchPage }: SearchInputProps) {
...
+      if(!container || container === undefined) return;
+      var className = 'gs-title';
+      
+      document.addEventListener('click', function (e) {
+        
+        var link = null;
+        if( e.target.parentElement.classList.contains(className) ) link = e.target.parentElement;
+        if( e.target.classList.contains(className) ) link = e.target;
+
+        if(link){
```
Entkenntnis: it would be good to check that `link.dataset.ctorig` is defined (had some exceptions here)

src/components/navigation/search-input.tsx
```diff
@@ -65,6 +65,32 @@ export function SearchInput({ onSearchPage }: SearchInputProps) {
...
+      const script = document.createElement('script')
+      script.innerHTML = `
+    var resultsContainer = document.getElementById('gcs-results');
+    setupLinkCatcher(resultsContainer);
+
+    function setupLinkCatcher(container){
+      if(!container || container === undefined) return;
+      var className = 'gs-title';
+      
+      document.addEventListener('click', function (e) {
```
Entkenntnis: can we add the click-hanlder only to the container?

src/components/navigation/search-input.tsx
```diff
@@ -65,6 +65,32 @@ export function SearchInput({ onSearchPage }: SearchInputProps) {
       input.setAttribute('placeholder', strings.header.search)
       input.focus()
       setSearchActive(true)
+
+      const script = document.createElement('script')
```
Entkenntnis: have you tried to include your script directly (without the script tag?). I think, this *should* work.

Entkenntnis (changes requested):
Hacky components need hacky solutions, love this functionality, maybe we can be a little bit less hacky (see reviews)

src/components/navigation/search-input.tsx
elbotho: Thanks… forgot to change that back.

src/components/navigation/search-input.tsx
elbotho: oh wow. obviously way nicer.

elbotho:
@Entkenntnis thank you very much for the pointers! Looks a bit nicer now.

Entkenntnis (approved):
nice

PR #391, opened 2020-07-14
output solution markup only when active

### LINKED ISSUE #389: Lazy Load exercise solutions
In the current code, exercise solution are rendered, no matter whether the toggle is open or closed:

https://github.com/serlo/frontend/blob/7389ab2128fda4999994e620643bb6780657cda3/src/components/content/exercise.tsx#L110-L117

Adding the lazy component here will improve page performance.

elbotho:
fixes https://github.com/serlo/frontend/issues/389 

I figure solving this is even better than lazy loading. Cleaner code and it also solves a problem with the lazy-loaded components inside the solution (they used to show only after you scrolled once after opening the solution).

### DIFF
--- a/src/components/content/exercise.tsx
+++ b/src/components/content/exercise.tsx
@@ -90,7 +90,7 @@ export function Exercise(props: ExerciseProps) {
 
       {renderSolutionToggle()}
 
-      {renderSolutionBox()}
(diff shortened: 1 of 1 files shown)


Entkenntnis:
Thanks, another page speed improvement!

PR #392, opened 2020-07-14
Auth: refresh access tokens

inyono:
First PR of an upcoming sequence of PRs to implement notifications.

Adds POST route `/api/auth/refresh-token` that tries to refresh the token. Its intended to be used when an API request fails with an authentication error. Gonna add a helper for that in an upcoming PR.


Entkenntnis (approved):
that's fine

PR #393, opened 2020-07-15
Add auth-aware GraphQL playground

### LINKED ISSUE #385: RFC: Add graphql playground to frontend
I'd like to add a graphql playground (similarly to https://api.serlo.org/___graphql) to the frontend. This way, testing out user-specific queries using your token would be easier (Log in to frontend and open ___graphql).

inyono:
Second PR of an upcoming sequence of PRs to implement notifications, depends on #392.

Adds a GraphQL playground at `/___graphql` that also passes the user token when the user is logged in. This way, one can easily test user-specific queries without manually working with the tokens. Also: the playground automatically refreshes the (relatively short-lived) access tokens when a request fails because the token is expired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "npm-run-all": "^4.0.0",
     "prettier": "^2.0.0",
     "prettier-plugin-packagejson": "^2.0.0",
+    "prop-types": "^15.7.2",
     "react-is": "^16.7.0",
     "typescript": "^3.9.0"
   }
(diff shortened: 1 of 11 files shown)


inyono:
I marked #392 as the base branch for this PR for easier code review. Should be changed to `staging` before merging.

package.json
```diff
@@ -79,6 +81,7 @@
     "npm-run-all": "^4.0.0",
     "prettier": "^2.0.0",
     "prettier-plugin-packagejson": "^2.0.0",
+    "prop-types": "^15.7.2",
```
inyono, line 84:
```suggestion
    "prop-types": "^15.5.0",
```

PR #394, opened 2020-07-15
Notifications

inyono:
Third (and last for now) PR of sequence of PRs to implement notifications, depends on #393 (which in turn depends on #392).

- Adds `UnreadNotifications` component that shows the number of unread notifications in the menu bar (number is updated once per minute if it is in view, using https://swr.vercel.app/.).
- Adds route `/user/notifications` (see `NotificationsList`) that renders the actual notifications. @elbotho I just added an example that renders the 20 newest notifications (independing of unread state) which matches our current implementation in the legacy system. Feel free to design a nicer approach for that, we pretty much have access to all notifications created ever and the API also handles pagination (so we could also do something like infinite scrolling). When you have a concept, I can help you with the data fetching again.


src/components/navigation/unread-notifications.tsx
```diff
@@ -0,0 +1,23 @@
...
+          totalCount
+        }
+      }
+    `,
+    config: {
+      refreshInterval: 60 * 1000, // seconds
+    },
+  })
+  return data ? (
+    <>You have {data.notifications.totalCount} unread notifications</>
```
inyono: Could be a badge in the menu bar that shows how many unread notifications you have.

src/components/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+function Notification({
+  event,
+}: {
+  id: number
+  unread: boolean
+  event: NotificationEvent
+}) {
+  switch (event.type) {
+    case NotificationEventType.SetThreadState:
```
inyono: Basic boilerplate to render the actual notifications (covers all possible event types)

inyono:
Similarly to #393, the base branch should be set to `staging` before merging

inyono:
@inyono: Possible improvements in the current useSWR implementation:
- [x] https://swr.vercel.app/docs/arguments#multiple-arguments
- [x] https://swr.vercel.app/docs/conditional-fetching#conditional

src/components/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,207 @@
...
+              type
+              date
+              actor {
+                id
+                username
+              }
+              object {
+                id
+              }
+              payload
```
elbotho: @inyono could we get the title of the Entity/Page/Repository here as well as the id?
This would be needed to replicate current behaviour (which is nice).

elbotho: (my understanding is, that this would have to be changes in the api, right?)

inyono: See Slack PN, let's talk about that before I make any changes.

elbotho:
Getting there!

@inyono:

**Read/Unread**
I talked to Wolfgang and we decided, that we would keep the current logic of setting Notifications to read after displaying them once for the foreseeable future. For now we could maybe even let the api set them to read when after they get send for the first time?
Ah and I think atm the query in [`UnreadNotificationsCount`](https://github.com/serlo/frontend/pull/394/files#diff-5aba44afc7926d7d3b92eb5d7d124a19R14) returns the total count, not only the unread count. Is that a known issue?

**Mute feature prototype**
I added a function to mute notifications using the same approach as the current Abonnement page. Would this work?
Could you maybe look over if my code in ([`getSubscriptionId`](https://github.com/serlo/frontend/pull/394/files#diff-9f548a98473ce526df1c9e074440f553R64)) if the ids I picked are the ones that actually connected to the subscription?

---

image
image
image

Entkenntnis:
I'm going to merge this as upcoming features are depending on this PR and it's not breaking existing stuff.

PR #395, opened 2020-07-17
Types for /fetcher/

### LINKED ISSUE #382: Types for the fetcher
Finally, everything is ready to fully type the fetcher code: The incoming api data is typed as `QueryResponse` and the return value of the fetcher is `PageData`.

elbotho:
Okay, only warnings left in `/fetcher`. 🎉 
But boy did I not really know what else I was screwing up in the meantime 😅 .

@Entkenntnis sorry if I messed up your new type order. I'm especially unsure about the explicit types in `create-data`. Not sure if they are needed, but I used them to solve some issues in `serlo-api`. I think they should be PageData

@inyono Took some shortcuts to get through the strict linter, maybe we can discuss the TODO comments and worst hacks in a pair session soon?

for #382

### DIFF
--- a/src/auth/oauth2.ts
+++ b/src/auth/oauth2.ts
@@ -65,7 +65,7 @@ export function getLogoutUrl({
   state: string
   callback: string
 }) {
-  if (config === null) return null
+  if (config === null || HYDRA_HOST === undefined) return null
   const query = new URLSearchParams({
     id_token_hint: idToken,
     state: state,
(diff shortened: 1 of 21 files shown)


elbotho:
Down to:
```
yarn lint:eslint:strict
✖ 67 problems (34 errors, 33 warnings)
```

🎉 but I cheated a lot in the `convert-edtr-io-state.tsx`. @inyono do you think it makes sense to import the actual editor types already?

elbotho:
ready now.

inyono (approved):
Thanks :)

PR #398, opened 2020-07-20
Track visits to static pages

elbotho:


Entkenntnis (approved):
looks good, let's try this.

PR #420, opened 2020-07-28
Deploy staging to production (week 31-32)

Entkenntnis:
Contains some work-in-progress stuff.

I18n and Authenification needs to be enabled in the cloudflare worker. Small breaking change: redirections are not handled by the frontend anymore.

Deploy after https://github.com/serlo/serlo.org-cloudflare-worker/pull/37


src/fetcher/get-initial-props.ts
```diff
@@ -60,15 +60,23 @@ export async function getInitialProps(
...
+      return {
+        origin: '',
+        pageData: { kind: 'error', errorData: { code: 200 } },
+      }
     }
 
     if (fetchedData.error) {
-      props.res!.statusCode = 404
+      const code = fetchedData.error.includes(
+        "Cannot read property 'path' of null"
```
inyono: Not happy about that. Can we cover that in the fetcher itself (Namely checking if uuid / whatever is null)? The actual error message might change depending on which node version Vercel is running on.

inyono (changes requested):

src/fetcher/get-initial-props.ts
Entkenntnis: The request throws an exception with this message: 
grafik

Can the API avoid throwing this error and just pass uuid as null instead? Otherwise the frontend is not much able to tell what happened.

inyono: Ah, didn't know that this happens at API level. My bad, gonna look into this.

Entkenntnis: No problem, thanks for investigating this.

inyono:
Won't merge this today since I don't know if it's in a deployable state (especially considering i18n / auth).

inyono (changes requested):
Sorry, merging still blocked by Cloudflare Worker, see https://github.com/serlo/serlo.org-cloudflare-worker/issues/40

inyono (approved):
Okay, Cloudflare Worker is finally ready.

PR #430, opened 2020-07-30
Fix for #428

### LINKED ISSUE #428: Cookie Banner needs two clicks to disappear
Accepting the cookie banner takes two clicks.

elbotho:
Still can't really reproduce this behaviour. But I found at least one bug in the code that might have caused it. 
Let's try…


Entkenntnis (approved):
Thanks for fixing this.

PR #432, opened 2020-07-31
Keyboard Nav

### LINKED ISSUE #416: Ensure keyboard navigation
Make the frontend accessible with keyboard: Check that every interactive element has a tabindex and the action can be triggered with the keyboard (enter key down). Also make sure that outline is visible.

Issues:

- [x] User Tools are not accessible (mobile and desktop version)
- [ ] 'hide solution' button has no outline
- [x] check share menu
- [x] check all content types

elbotho:
Fixes #416 

Possible improvements for the future (after 100%?):
- user tools should be accessible earlier (currently after content on desktop)
- client side routing does not move focus up on courses (not a huge problem)
- applets are not working with keyboard nav (just a note, can't really change that)


Entkenntnis (approved):
Thanks for investigating and fixing this.

Issue #349, opened 2020-06-29
Long inline formula should break


elbotho:
e.g.
https://de.serlo.org/mathe/stochastik/grundbegriffe-methoden/ergebnisraeume/ergebnismenge

elbotho:
```
.mCrJk .katex .base {
    position: relative;
    display: inline-block;
    white-space: nowrap;
    width: min-content;
}

elbotho:
Build a test case here:
https://de.serlo.org/community/sandkasten/dev:-edge-cases

Quickfix: put into boxes

inyono:
Might want to look into the Edtr.io implementation since I'm pretty sure we handled that already in the legacy system. We even render block formulas as KaTeX inline formulas so that they are responsive (which we'll opt-out in the equations plugin because it doesn't work well for this use case).

Entkenntnis:
An interesting edge case, the frontend implementation is currently 1:1 to edtr.io, so this is a problem happens even with KaTeX inline formulas. The legacy system uses mathjax and breaks the formula correctly...

For the moment, the quickfix of making the formula a block-formula (instead of inline) would make the most sense, as changing KaTeX css could result in unexpected side effects.

Edit: It also looks like a buggy behaviour of KaTex, so maybe this can be solved by tweaking the latex code? Because it breaks after the equals sign, but not within the set.

elbotho:
@Entkenntnis so your conclusion is, that it's basically a content problem?
I added a fix in #433 that seems to work quite well. 
Could you verify that this works? :)

PR #433, opened 2020-07-31
Fix long katex

elbotho:
possible fix for #349


Entkenntnis (approved):
works fine, let's try it this way.

Issue #434, opened 2020-07-31
Complete Translations


Entkenntnis:
Every language has one file with all its strings (https://github.com/serlo/frontend/tree/staging/src/data). The strings are splitted into four types:

- loaded for every page of the instance (instance data)
- loaded only for the landing page (landing data)
- loaded only for logged in users (logged in data)
- loaded only on the server (server data)

For the current milestone, it would be enough if these files are translated. In the long-term, reintegration with crowdin (or similiar) would be great.

elbotho:
Branch is here: https://github.com/serlo/frontend/tree/crowdin-integration

It's not as pretty as I hoped it would be.
I had to move the menu data to a second file, because it is not structured the same way across languages.
we need upload those independently to use crowdin for those.

but now the rest of the strings can be edited and crowdin exports valid ts files back.

TODO:
- [x] There is a bug with german translations where umlauts get messed up. Waiting for the support to get back to me here, but actually it seems to work with the GitHub integration. Fingers crossed.
- [x] currently all targeted languages (53 of them) get exported and this make quite a mess. [Lots of commits](https://github.com/serlo/frontend/pull/437). Also this currently breaks lint:tsc and therefore the build

elbotho:
Okay, translations are done, but we keep coming up with new stuff so we also need translations for:

- [x] notifications (ready for translation now)
- [x] authed menus (ready for translation now)
- [x] comment system (ready for translation now)

elbotho:
Okay, I'll close this, because Crowdin is set up now and the workflow seems to work okay. 
Now we need to wait for the translations and need to keep i18n in mind for new features.
Follow up in #483.

PR #441, opened 2020-08-03
Crowdin integration

elbotho:
Okay! Looks good for now. 

- Menu data has to be edited via GitHub/manually for now.
- Rest of the strings should work automagically with Crowdin and their GitHub integration. 
- Missing strings fall back to `en` (which is the default file)
- I contacted _Jakob & the I18nalizers_ (cool bandname 🎵) so hopefully we will have full translation soon. 
- [Example PR from Crowdin](https://github.com/serlo/frontend/pull/440), looks good now. Only includes languages we use and actually builds 😅 

Ready for review, might work on the link encoding thing, but that is low priority right now.

relates to #434

ps: crowdin ux to set it up was meh, but their support was 👍


Entkenntnis:
We are lucky that crowdin now supports ts files :tada: The interface and the integration look fine, thanks for working on this.

Have you tested how we can change the structure of the translation? Would it be enough to edit `en/index.ts` or is it more involved? Just to make sure that we are not breaking stuff.

Entkenntnis (approved):

elbotho:
jup, editing `en/index.ts` should work fine.
it's also a good idea to hide strings in crowdin that translator should not change.

PR #445, opened 2020-08-05
API v0.8.0+ Notifications

inyono:
Uses api.serlo.org v0.8.0+ (which is deployed to dev at the moment).

## TODO

- [x] Use types from `@serlo/api`.
- [x] Use fancy refactored Notifications.


.env.local
```diff
@@ -1,6 +1,6 @@
 NEXT_PUBLIC_ENV=staging
 NEXT_PUBLIC_SERLO_DOMAIN_PRODUCTION=serlo.org
-NEXT_PUBLIC_SERLO_DOMAIN_STAGING=serlo-staging.dev
+NEXT_PUBLIC_SERLO_DOMAIN_STAGING=serlo-development.dev
```
inyono: @inyono Before merging this PR:
- Use staging API again:
  ```suggestion
  NEXT_PUBLIC_SERLO_DOMAIN_STAGING=serlo-staging.dev
  ```
- Deploy API v0.7.0+ to staging.
- Remove cache.

elbotho:
Okay notifications use fancy refactored data now and add a lot of info like entity titles.
@inyono code review is welcome. Also I'm pretty sure I'm not fetching all the data, maybe you could also scroll through your notifications-pile quickly.

As soon as we have an deployment preview of this, I'd also like to ask Wolfgang if the texts work for him like this.
~~Next up pagination?~~

**Pagination:**
I added a simple "load more" button + function. Maybe that's enough for now?

src/components/user/notification.tsx
```diff
@@ -62,152 +94,167 @@ export function Notification({
...
+  function renderText() {
+    switch (event.__typename) {
+      case 'SetThreadStateNotificationEvent':
         return (
           <>
-            <UserLink user={event.actor} /> hat Thread {event.thread.id}
+            {renderUser(event.actor)} hat einen {renderThread(event.thread.id)}{' '}
             {event.archived ? 'archiviert' : 'unarchiviert'}
           </>
         )
```
elbotho: I tried to make this easy to read, okay for now.

@inyono do you have an idea of how we could make this translatable?
My best guess so far is to write it like `"%USER% has %ACTION% a %THREAD%"` and write a little parser, which seems ridiculously complicating for what it is.

elbotho: (the render… functions return links as JSX elements)

inyono: That is usually what an i18n library would handle for us 😛 Otherwise, we'll probably need some parser for that, yeah.

inyono:
Looks good so far, haven't reviewed the code in detail. What I noticed:
- [x] We definitely need some loading indicator for the initial load and load more button.
- [x] Add placeholder for entities without title, e.g. replace `[Solution]` with something like `a solution`.

src/components/user/notification.tsx
elbotho: This is how it would work with next-i18next
``` jsx 
<Trans
  i18nKey="myKey"
  defaults="hello <italic>beautiful</italic> <bold>{{what}}</bold>" // optional defaultValue
  values={{ what: 'world'}}
  components={{ italic: <i />, bold: <strong /> }}
/>
```
[(source)](https://react.i18next.com/latest/trans-component#alternative-usage-v-11-6-0)

Wrote a little parser anyway, see next comment.

src/components/user/notification.tsx
```diff
@@ -124,141 +125,174 @@ export function Notification({
...
-          <>
-            {renderUser(event.actor)} hat {renderObject(event.child)} mit{' '}
-            {renderObject(event.parent)} verknüpft.
-          </>
-        )
+        return parseString('%actor% hat %child% mit %parent% verknüpft.', {
+          child: renderObject(event.child),
+          parent: renderObject(event.parent),
+        })
+
```
elbotho: this is the parser interface currently. 
a bit verbose (duplicate code for e.g. `child: renderObject(event.child)`) but clean and typescript is happy. 
should I move this logic into the parseString function? (possibly annoying because TS looses the relation to type, but less code overall)

elbotho: ping @inyono

PR #448, opened 2020-08-07
License icon, video description and other small stuff

### LINKED ISSUE #435: Don't show CC icon for license if entity is not creative commons
We currently always show a cc icon for the license, but we have many entities that are not creative commons. We need to show another icon in these cases.

### LINKED ISSUE #447: Do not show video description for injections
When you embed a video into other serlo content, the video description is not useful. Thus it should not be displayed. Example article with an embedded video: https://frontend.serlo.org/mathe/zahlen-groessen/bruchrechnen-dezimalzahlen/dezimalbrueche/addition-subtraktion-dezimalbruechen

elbotho:


Entkenntnis (approved):
nice solutions

PR #451, opened 2020-08-10
Deploy staging to production (week 33)

inyono:


inyono:
FYI: I'd like to delay Monday's release a week because we had a couple of breaking changes in the API and our Cache Worker isn't ready yet (which means we'd need to empty the cache in production). Is that okay for you @Entkenntnis?

PR #463, opened 2020-08-12
Plan B for comments

Entkenntnis:
A small prototype to load comments from legacy system by extracting the html.


Entkenntnis:
well, this is trash

PR #476, opened 2020-08-17
User profiles Draft

elbotho:
### example
https://frontend-git-user-profiles.serlo.vercel.app/user/profile/18981

### todo
- [ ] translate
- [ ] probably implement #479 here
- [ ] then consider adding to staging, without activating it in CSR or SW or only for non-authed users so this PR is not just hanging around

### status
- Uses mockup data for now (because descriptions are not part of the staging data (returns NULL)
- Waiting vor https://github.com/serlo/api.serlo.org/issues/69


elbotho:
``` ts
test('User', async () => {
  server.use(
    rest.post(endpoint, (req, res, ctx) => {
      return res(
        ctx.json({
          data: {
            uuid: {
              __typename: 'User',
              id: 1,
            },
          },
        })
      )
    })
  )
  const response = (await fetchPageData('/de/1')) as ErrorPage
  expect(response.kind).toEqual('error')
  expect(response.errorData.code).toEqual(200)
})
```

@Entkenntnis , @inyono who wrote this test and want's to explain me quickly how write a useful check for user profiles?

inyono:
Currently, it's basically:
- Copy-Paste the fetcher query into the staging api
- Copy-Paste the response into `res(ctx.json( ... ))` of the test

PR #482, opened 2020-08-17
Review assets

### LINKED ISSUE #453: Review assets
Importing from `public/_assets` doesn't make sense since imports will be handled by Webpack. So either add assets to `public/_assets` and use them via absolute url or import them from somewhere in `src/`

Also: especially small SVGs can be React components (so we don't have the overhead of another HTTP request), see e.g. https://github.com/gregberge/svgr.

elbotho:
Closes #453

Notes:
- Logo SVG is not embedded, because I want it to be accessible to download and accessible through image search
- All imported SVGs are moved into `@/assets-webkit` (verbose name to make it easy to understand)
- simplify landing page logic to remove some duplication
- small german landing page fix


inyono (approved):
Thanks 👍

PR #488, opened 2020-08-20
Finish input ex implementation

### LINKED ISSUE #261: complete input exercise implementation
Some bits are missing from input exercises:

- normalize input
- InputExpressionEqualMatchChallenge (algebra.js)

https://github.com/edtr-io/edtr-io/blob/master/packages/plugins/input-exercise/src/renderer.tsx

elbotho:
- closes #261
- implementation without `string` package, because it's not maintained any more. https://vocajs.com/ might be an alternative but I think the regex is okay for now.
- algebra js is also not maintained any more, but for now I keep using it.
- (def. needs more testing!)


src/components/content/input-exercise.tsx
```diff
@@ -35,24 +36,54 @@ export function InputExercise({ data }: InputExerciseProps) {
...
     } else {
       return <span>{strings.content.right}</span>
     }
   }
+
+  function normalize(value: string) {
+    const _value = collapseWhitespace(value)
+    switch (data.type) {
+      case 'input-number-exact-match-challenge':
+        return normalizeNumber(_value).replace(/( )?\/( )?/g, '/')
```
elbotho, line 76: @inyono this should be equivalent to the edtr implementation,
but what about `* ` ` *`, etc? Wouldn't it make more sense to remove all whitespaces for evaluation? Or is this another edge case?

Entkenntnis:
Thank you for tackling this. I've done some testing and fixed the handling of parse errors. I think this should suffice for now.

Issue #490, opened 2020-08-20
Load text exercises with new frontend  [cf-worker]


Entkenntnis:
On staging, currently all exercises are loaded with the legacy backend, e.g. https://de.serlo-staging.dev/mathe/funktionen/ableitung-funktionen/tangente-normale/aufgaben-tangente/13799

@kulla Could you please take a look at this and update the cloudflare worker? Thanks!

Entkenntnis:
Localhost is not affected, but #410 will not work on staging (yet).

elbotho:
should work because of CSR, but I'll create a PR in minute so we can test.

Entkenntnis:
This occurs if you are editing exercises: If you switch from legacy system (e.g. editor) to an exercise, the exercise is shown in the legacy system.

But you have a point: If we are using CSR, this will not happen.

kulla:
The type `Exercise` is not in the list of allowed types for the frontend. See https://github.com/serlo/serlo.org-cloudflare-worker/blob/master/wrangler.toml#L41

@Entkenntnis What is the current list of types which are supported?

Entkenntnis:
We can add Exercise and ExerciseGroup to the list of allowed types, they are fully supported now.

Maybe we can test it on staging?

elbotho:
a little aside, but I thought its interesting what other types are therefore not supported atm:
With your addition @Entkenntnis that leaves:
- Event (✔️ )
- GroupedExercise (✔️ )
- Solution (in progress ✔️ ))
- User _(#476 )_

and the revisions: AppletRevision, ArticleRevision, CourseRevision, CoursePageRevision, EventRevision, ExerciseRevision, ExerciseGroupRevision, GroupedExerciseRevision, PageRevision, SolutionRevision, VideoRevision

Entkenntnis:
Event and GroupedExercise will work, but are rarely link to. We can add them to the allowed types.

Solutions are not supported.

kulla:
Staging & dev: Event, Exercise, ExerciseGroup

kulla:
See https://github.com/serlo/serlo.org-cloudflare-worker/pull/48 for the fix. In Staging and Dev this is already deployed. See 

* Exercise: https://de.serlo-staging.dev/mathe/funktionen/ableitung-funktionen/tangente-normale/aufgaben-tangente/13799
* ExerciseGroup: https://de.serlo-staging.dev/mathe/analysis/ableitung/quotientenregel/6763

[...]

PR #515, opened 2020-08-24
add support for type Solution

### LINKED ISSUE #514: Implement `Solution`
As in #513 I discovered those via sentry.

https://de.serlo.org/mathe/universitaet/grundlagen-mathematik/relationen/aequivalenzrelationen
- injection /56990 _(▾ Verständnisfrage: Wie viele lineare Äquivalenzrelationen auf einer Grundmenge MM gibt es?)_
- injection /57015 _▾ Beweis: (Äquivalenzrelationen induzieren eine Zerlegung)_
- injection /57020 _▾ Beweis: (Jede Zerlegung induziert eine Äquivalenzrelation)_ 

All fail with `{"kind":"error","errorData":{"code":200}}`
e.g. https://frontend.serlo.org/api/frontend/de/57020

does happen in staging

---

~https://de.serlo.org/mathe/stochastik/grundbegriffe-methoden/baumdiagramm-vierfeldertafel/vierfeldertafel~
~injection /21332 (bottom _▾ Weitere Übungsaufgaben_)~
~error: https://frontend.serlo.org/api/frontend/de/21332~
~`{"kind":"error","errorData":{"code":500,"message":"Error while fetching data: Cannot read property 'text' of undefined"}}`~
~does not happen in staging~
(should be fixed in staging #464)

elbotho:
closes #514

### DIFF
--- a/src/components/content/exercises/exercise.tsx
+++ b/src/components/content/exercises/exercise.tsx
@@ -1,13 +1,6 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
 
-import {
-  makeMargin,
-  makeDefaultButton,
-  makePadding,
-  inputFontReset,
-} from '../../../helper/css'
-import { renderArticle } from '../../../schema/article-renderer'
 import { AuthorTools } from '../author-tools'
 import { LicenseNotice } from '../license-notice'
 import { ExerciseNumbering } from './exercise-numbering'
@@ -15,12 +8,14 @@ import { InputExercise } from './input-exercise'
 import { ScMcExercise } from './sc-mc-exercise'
 import { useAuth } from '@/auth/use-auth'
 import { useInstanceData } from '@/contexts/instance-context'
+import { FrontendExerciseNode } from '@/data-types'
 import {
-  FrontendContentNode,
-  FrontendExerciseNode,
-  LicenseData,
-  SolutionEdtrState,
-} from '@/data-types'
+  makeMargin,
+  makeDefaultButton,
+  inputFontReset,
+  makePadding,
+} from '@/helper/css'
+import { renderArticle } from '@/schema/article-renderer'
 
 export interface ExerciseProps {
   node: FrontendExerciseNode
@@ -55,13 +50,22 @@ export function Exercise({ node }: ExerciseProps) {
 
   function renderSolution() {
     return (
-      <Solution
-        solutionEdtrState={node.solutionEdtrState}
-        solutionLegacy={node.solutionLegacy}
-        license={node.solutionLicense}
-        authorTools={
-          loaded &&
-          auth.current && (
+      <SolutionBox>
+        {renderArticle(
+          [
+            {
+              type: 'solution',
+              solution: node.solution,
+              context: { id: node.context.solutionId! },
+            },
+          ],
+          false
+        )}
+        <SolutionTools>
+          {node.solution.license && (
+            <LicenseNotice minimal data={node.solution.license} />
+          )}
+          {loaded && auth.current && (
             <AuthorTools
               data={{
                 type: '_SolutionInline',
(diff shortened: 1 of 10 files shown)


src/components/content/exercises/exercise.tsx
```diff
@@ -162,6 +141,62 @@ export function Exercise({ node }: ExerciseProps) {
...
 }
 
+export interface SolutionProps {
+  license?: LicenseData
+  authorTools?: React.ReactNode
+  solutionEdtrState?: SolutionEdtrState
+  solutionLegacy?: FrontendContentNode[]
+}
+
+export function Solution({
```
Entkenntnis: Would be nice to move this to a separate file.

src/data-types.ts
```diff
@@ -396,6 +396,18 @@ export interface FrontendExerciseNode {
   href?: string
 }
 
+export interface FrontendSolutionNode {
```
Entkenntnis, line 405: There is now some duplication with FrontendExerciseNode, maybe add FrontendSolutionNode to FrontendExerciseNode?

src/components/content/exercises/exercise.tsx
```diff
@@ -44,10 +49,33 @@ export function Exercise({ node }: ExerciseProps) {
...
       {renderToolsAndLicense()}
 
-      {solutionVisible && renderSolutionBox()}
+      {solutionVisible && renderSolution()}
     </Wrapper>
   )
 
+  function renderSolution() {
+    return (
+      <Solution
```
Entkenntnis: You have alos added solution to articleRenderer, so it's probably good to use the articleRenderer here.

src/components/content/exercises/exercise.tsx
```diff
@@ -44,10 +49,33 @@ export function Exercise({ node }: ExerciseProps) {
...
+        solutionEdtrState={node.solutionEdtrState}
+        solutionLegacy={node.solutionLegacy}
+        license={node.solutionLicense}
+        authorTools={
```
Entkenntnis: Ok, we would need to copy auth and loaded to the new solution component ... this way, single view solutions will also have an edit box.

src/fetcher/create-exercises.ts
```diff
@@ -67,6 +68,34 @@ export function createExercise(
   }
 }
 
+export function createSolution(uuid: Solution): FrontendSolutionNode {
```
Entkenntnis: if FrontendSolutionNode is a property of FrontendExerciseNode, you can call this function from the method above.

src/fetcher/query.ts
```diff
@@ -279,6 +286,17 @@ export const dataQuery = gql`
         }
       }
 
+      ... on Solution {
```
Entkenntnis, line 290: It would be nice if this could be a fragment that exercise can reuse.

Entkenntnis (commented):
Hi @elbotho, you have updated the code in all relevant places. Before we merge this, I would like to improve this a little bit by using your new solution type also for existing exercises (see comments).

If you need help I can assist you or take over this PR :smile:

Entkenntnis:
@elbotho By the way, this PR would also be a good start for adding a test case for Solutions.

src/components/content/exercises/exercise.tsx
elbotho: I did not want to support that since the edit workflow we discussed (loading whole exes or even groups). 
I'd actually discourage the entity view of a solution. This is manly to support the injection edge case.

elbotho: How do we do that with events atm? Can I single view an event?

elbotho: Ah, we could redirect the single view of an id to the single view of the parent exercise –
that should work with the context id's right?

elbotho:
Thanks for the review. I guess I rushed this a bit since I see it as an edge case and wanted it to slip into the cf-worker update (#490). 
Will clean it up a bit 👍

kulla:
> Thanks for the review. I guess I rushed this a bit since I see it as an edge case and wanted it to slip into the cf-worker update (#490).

No need to rush -> I can easily add new allowed types in cloudflare. Just ping me at Slack :-)

inyono:
cc https://github.com/serlo/frontend/issues/514#issuecomment-682059357

Entkenntnis:
Sorry, we have a slight change of plan (see #514), but we can keep most of solution refactoring.

Entkenntnis (approved):

Issue #436, opened 2020-08-03
RFC: Just don't show default license for exercises and injections


elbotho:
### Suggestion:

**Default case:** We only show license information at the bottom of an entity.
![image](https://user-images.githubusercontent.com/1258870/89166946-121ff780-d57b-11ea-9c33-03ec90f9a0dc.png)

So we don't display it for embedded exercises, solutions, exercise groups, images and injections…
We might need to change the text to something like "Wenn nicht anders angegeben…"

---

**Exception:** If the license differs from CC BY-SA 4.0 we display a short text and an icon ([not the cc icon](#435) 😄) with title attribute and some other way for mobile (let's see)

![image](https://user-images.githubusercontent.com/1258870/89167092-42679600-d57b-11ea-84a7-08b9574766b9.png)

---
Reasoning: I don't see a benefit of showing all of those CC icons especially for exercise collections they create a lot of visual clutter atm. 

Any objections, ideas? Did I miss something?

ping @inyono @yolo-bird @simonkoehl 
related #269

Entkenntnis:
Previous discussion: https://github.com/serlo/frontend/issues/108

ghost:
So just to give you a few facts. Currently there are 2184 entities licensed not under the CC-BY-SA 4.0:
- 2160 instructions of text exercises and text-excercise groups (the solutions are all under  CC-BY-SA 4.0)
- 24 videos

In the unit begleitung we would like to reduce the number of entitites under our standard license. We expect to reduce the number of videos to 2, so here we could just write the separate license in the video description and we will not need some design solution imho.
If we succeed with the exercises we will need a way of communicating the "by" in the standard license clearly, if the exercise was created by somebody else. 

For exercises embedded via injection I would still show the license if it differs from our standard one.

elbotho:
@yolo-bird okay so we especially mark content that is not CC-BY-SA und also content where the license is CC-BY-SA but Serlo is not the source to be attributed. Good for you?

@simonkoehl there were no real objections so far, what do you think about this?

Entkenntnis:
If there is no objection, I would suggest that we start trying this with topic folders: In a "topicFolder" or "curriculumTopicFolder", we add a CC-BY-SA license notice to the bottom (equals our license id 1) and remove all other license notices with id 1 within the exercises. All other license will stay visible.

@elbotho This feels straight forward to implement, do you want to create a PR for it?

[...]

PR #517, opened 2020-08-25
Less license notices

elbotho:
for #436 

before we merge this into production we should add some notice to the default descriptions:
https://serlo.org/license/detail/1

See https://github.com/serlo/frontend/issues/436#issuecomment-680012757

Edit: The description in https://de.serlo.org/terms coudl suffice.


Entkenntnis (approved):
Thank you for cleaning of the license notices.

PR #530, opened 2020-08-30
rework error page

elbotho:
More user friendly error page
see e.g. /121199 for an unsupported type.


Entkenntnis (approved):
Thank you! Looks good.

PR #533, opened 2020-08-30
Support for revisions

elbotho:
## Status
- After discussing with @kathongi we will test the comparison to the current revision because it might actually be more useful, so there's no need to implement something else.
- all revision types are now supported.

## Supported revision types (all now)
- ArticleRevision: http://localhost:3000/en/160832
- ArticleRevision (Currently accepted version): http://localhost:3000/en/162707
- PageRevision: http://localhost:3000/en/147897
- VideoRevision: http://localhost:3000/en/99591
- CoursePageRevision: http://localhost:3000/en/62537
- EventRevision: http://localhost:3000/en/164902
- AppletRevision http://localhost:3000/en/141411
- CourseRevision -> http://localhost:3000/129611
- GroupedExerciseRevision -> http://localhost:3000/99372
- ExerciseRevision -> http://localhost:3000/54265
- ExerciseGroupRevision -> http://localhost:3000/51428
- SolutionRevision -> http://localhost:3000/124265

## Backlog
- [x] timeago implementation (easy once #543 is merged)

## Before merging into production
- SW should probably redirect /51428 to https://de.serlo.org/entity/repository/compare/51420/51428 as is currently the case
- Make sure new german strings have a translation


Entkenntnis:
According to #556, the query size has no ~~big~~ effect on the query time, so I think we are good to go here.

elbotho:
@Entkenntnis wow thanks for the benchmark!
it turned out to be not to be too much additional query.

PR #527, opened 2020-08-28
Improve fetcher tests, add edtr converter tests and types, refactor converter

elbotho:

PR #545, opened 2020-09-03
Import edtr io types

Entkenntnis:
The other part of #527


Entkenntnis:
Some notes:

https://github.com/serlo/frontend/pull/527#discussion_r482926047

https://github.com/serlo/frontend/pull/527#discussion_r482927883

https://github.com/serlo/frontend/pull/527#discussion_r482969315

elbotho:
Status:
Some open questions being discussed in https://github.com/edtr-io/edtr-io/issues/317

__tests__/schema/convert-edtr-io-state.ts
```diff
@@ -67,6 +67,7 @@ describe('edtr io plugins', () => {
     })
   })
 
+  //TODO: Not supported in editor any more, should we do something here? (like make it bold at least?)
```
inyono: Actually it is supported, see https://github.com/serlo/serlo.org/blob/6eca3d8c02994f5d0356d8bd73907a9afb3fd293/packages/private/edtr-io/src/plugins.tsx#L109 and https://github.com/serlo/serlo.org/blob/6eca3d8c02994f5d0356d8bd73907a9afb3fd293/packages/private/edtr-io/src/editor.tsx#L212.

package.json
```diff
@@ -82,6 +100,8 @@
     "@types/react-modal": "^3.0.0",
     "@types/react-notify-toast": "^0.5.0",
     "@types/simple-oauth2": "^4.0.0",
+    "@types/slate-html-serializer": "^0.6.4",
```
inyono: Do you need these types? Doesn't look like you import `slate-html-serializer` and the text plugin provides its own types.

package.json
```diff
@@ -82,6 +100,8 @@
...
+    "@types/slate-react": "^0.22.9",
```
inyono: Do you need these types? Doesn't look like you import `slate-react` and the text plugin provides its own types.

package.json
```diff
@@ -102,6 +122,8 @@
     "prettier": "^2.0.0",
     "prettier-plugin-packagejson": "^2.0.0",
     "prop-types": "^15.7.0",
+    "react-dnd": "^11.1.3",
```
inyono:
```suggestion
    "react-dnd": "^11.0.0",
```

package.json
```diff
@@ -102,6 +122,8 @@
...
+    "react-dnd-html5-backend": "^11.1.3",
```
inyono:
```suggestion
    "react-dnd-html5-backend": "^11.0.0",
```

package.json
inyono: We probably don't need this peer dependency at this point since this is only a runtime dependency for the editor itself.

package.json
inyono: We probably don't need this peer dependency at this point since this is only a runtime dependency for the editor itself.

src/schema/convert-edtr-io-state.tsx
```diff
@@ -10,91 +17,76 @@ import {
...
-        alt: (node.state as EditorStateDummy).alt!,
-        maxWidth: (node.state as EditorStateDummy).maxWidth,
+        src: node.state.src,
+        alt: node.state.alt,
+        maxWidth: node.state.maxWidth,
       },
     ]
   }
-  if (plugin === 'important') {
+  //Note: Not supported any more in edtr
```
inyono: Not true, see comment above. As equations, this is a Serlo-internal plugin.

inyono (commented):
Thanks, looks good. Noted some minor things.

__tests__/schema/convert-edtr-io-state.ts
elbotho: @inyono Thanks! Issue in #597

package.json
elbotho: removed

package.json
elbotho: removed

package.json
elbotho: removed, seems too work fine 👍

inyono (approved):
Thanks

PR #547, opened 2020-09-03
Hot fix Deployment

Entkenntnis:
I'll do some more testing and deploy this soon (scheduled on monday).


elbotho:
So it's a normal deployment right on schedule then 😸

elbotho:
500 that is not happening in production:
https://frontend-1fj74nsp6.vercel.app/mathe/funktionen/wichtige-funktionstypen-eigenschaften
vs
https://de.serlo.org/mathe/funktionen/wichtige-funktionstypen-eigenschaften

Entkenntnis:
@elbotho known bug, unfortunately (we have fixed the content, staging seems not to be up-to-date yet)

PR #564, opened 2020-09-04
Add multimedia component

### LINKED ISSUE #548: Image Magnific Plugin: Broken in new content
Images don't maximize on click
(reported by author hxwalther)

### LINKED ISSUE #550: Implement floating behaviour for multimedia-plugin
The implementation of the multimedia-plugin was I bit rushed by me, so it is currently missing support for floating (it actually just reuse the row node).

We should probably add a new node type `multimedia` and use the same implementation as edtr-io.

Example: https://de.serlo.org/community/sandkasten/politik/bundestag/abgeordnete

elbotho:
closes #550 
very bare-bones, but should be very similar to edtr-io implementation.

### DIFF
--- a/src/components/content/light-box.tsx
+++ b/src/components/content/light-box.tsx
@@ -1,0 +1,78 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Modal } from '../modal'
+import { makeMargin } from '@/helper/css'
+
+export interface LightBoxProps {
+  open: boolean
+  onClose: () => void
+  src: string
+  label?: string
+}
+
+export function LightBox(props: LightBoxProps) {
+  const { open, onClose, label, src } = props
+  if (!open) return null
+
+  return (
+    <StyledModal isOpen={open} onRequestClose={onClose}>
+      <_StyledImg onClick={onClose} src={src} alt={label || 'Bild'} />
+      <p>{label}</p>
+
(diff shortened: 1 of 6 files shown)


Entkenntnis (approved):
Thank you very much for finishing the multimedia plugin implementation!

src/components/content/light-box.tsx
```diff
@@ -0,0 +1,78 @@
...
+//   color: ${props => props.theme.colors.dark1};
+//   &:hover {
+//     background-color: ${props => props.theme.colors.brand};
+//     color: white;
+//   }
+//   width: 35px;
+//   height: 35px;
+//   border-radius: 30px;
+//   text-align: center;
+// `
```
kulla, line 78: Why is this code in a comment? Should this have been deleted? :smiley:

elbotho: sure could have deleted it… but since this is still in review I left it in.
don't know what the best approach is, but how would somebody other than me (also: me in 3 weeks) know that there was finished code for a close button at some point?

kulla: Possibilities:

- Add a commit with a close button and remove it afterwards -> we can access the code via the git history
- Add a comment to this PR with this code
- Add a comment to Slack or this code to a Slack thread

See https://github.com/labs42io/clean-code-typescript#dont-leave-commented-out-code-in-your-codebase

PR #566, opened 2020-09-05
Simplify button code

### LINKED ISSUE #532: Refactor buttons
there are some places where we use a very similar button style that would be worth putting into a small component by now

elbotho:
for #532


Entkenntnis (approved):
nice clean up

Issue #531, opened 2020-08-30
improve timeago i18n


elbotho:
Find a solution for react timeago to dynamically load only language info for current instance. At the moment I just import all the language files. ~and build into reusable component. Currently we use it for comments, notifications and revisions.~

elbotho:
@Entkenntnis do you have an idea how we should approach this?

Entkenntnis:
Not really, yet. The language file consists of js-function, so we can't pass it around as json. Maybe we can use the async `import()` statement to handle this? https://webpack.js.org/api/module-methods/#import-1

PR #581, opened 2020-09-08
Use dynamic import for TimeAgo languages

elbotho:
@Entkenntnis as per your suggestion in #531 here is a draft. 

Should work?
Downside: Regex in `webpackInclude` magic comment has to be maintained manually.

### examples
http://localhost:3000/160832
http://localhost:3000/en/160832


Entkenntnis:
Hm, it builds, but is still not working yet ...

Entkenntnis:
.Ok, now it's fine.

PR #598, opened 2020-10-07
Add support for blockquote plugin

### LINKED ISSUE #594: Add support for blockquote plugin
The block "Unsere Vision ist freie Bildung, die von einer offenen und unabhängigen Gemeinschaft [...]" is missing in the new frontend, probably a legacy node that we currently do not support.

If it's not too much work, we can add support for this.

elbotho:
fixes #594

### example
https://frontend-git-add-blockquote-support.serlo.vercel.app/community/sandkasten/dev:-edge-cases


inyono (approved):
LGTM

PR #599, opened 2020-10-07
add basic print stylesheet

### LINKED ISSUE #589: Optimize print stylesheet
Compared with the legacy system, the print stylesheet still needs some optimization (e.g. don't show navigation, don't show hroizon, ...)

elbotho:
- for #589
- quite basic
- good enough for now?


inyono:
- [x] Cookie banner should be removed, too (is shown if you don't scroll to the end of the site before printing):
Screen Shot 2020-10-09 at 09 45 26
- [x] Horizon should be removed (is shown if you scroll to the end of the site before printing).
- [x] Font size looks quite large for a printed document.
- [x] We probably should remove the icon that shows the type of the entity (e.g. the article icon), too.
- [x] Btw: I only checked articles, other content types might have additional issues.
- [ ] (fine as follow-up issue later): Maybe we can have a look at [`onbeforeprint`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint). We can at least move the warning to scroll to there (instead of adding it to the print) and maybe even automate the scrolling / loading there (not sure how feasible this is because I'm not sure how the current implementation works).
- [ ] (fine as follow-up issue later): The inline URLs make stuff hard to read. I'd prefer to display them as a footnote or at the end of the document instead.

@metzgaria do you have any additional feedback?

elbotho:
@inyono thank you for the thorough review!

#### Other content types
I checked exercise collections and courses yesterday (tbh not very thoroughly).
Work ok, courses can only be printed one page at a time, but we could change that once we change how we load courses in general.

#### Follow ups
In the [issue](https://github.com/serlo/frontend/issues/589) I mentioned some ["research"](https://stackoverflow.com/questions/43707762/how-to-lazy-load-images-and-make-them-available-for-print) I did.
Even `onbeforeprint` doesn't look to promising, we could probably get some browsers to work, but not all. 
Same with the footnotes, we need to check if we could run some JS before the print dialog is open and the print content is set.
Should we then just not show the URLs then for now? (Like we do currently)

If we change the lazy load implementation to set a state one all of them loaded we could make the UX a bit nicer.

Will open an issue.

#### Font size
Will actually print a sample to test :)

inyono:
Okay, thanks. Haven't look too much into the issue so far. So I'm fine with giving feedback and letting you decide what we do before the first release and what we do later (or never, depending on urgency ^^) 👍

inyono (approved):

elbotho:
Output of my local environment:
```
yarn lint:prettier
yarn run v1.22.5
$ yarn _prettier --check
$ prettier . --check
Checking formatting...
All matched files use Prettier code style!
✨  Done in 5.04s.
```

All good but fails here. Any ideas @inyono or @Entkenntnis ?

inyono:
There was probably some outdated prettier somewhere in the node_modules. In such cases, `rm -rf node_modules && yarn` helps.

elbotho:
@inyono thank you for fixing this!
follow up is in #602.
Will wait a little bit for feedback by @metzgaria before merging.

PR #600, opened 2020-10-08
Link minimal license notice to internal page

### LINKED ISSUE #596: Add support for free licenses compatible to the serlo license
We have licenses like https://de.serlo.org/license/detail/7 which are compatible to the serlo CC-BY-SA license but where we need to attribute the original author. In order to support those licenses the following should be implemented:

### 1. Link to license details

Whenever a non serlo license is shown the icon shall link to the license detail page at `/license/detail/:license-id`:

2020-10-06-151001_1920x1080_scrot

### 2. Other symbol

For those licenses (currently the license with `id=7`) which are not the serlo license but which are compatible to the serlo license another symbol should be shown (example: https://commons.wikimedia.org/wiki/File:Cib-creative-commons_(CoreUI_Icons_v1.0.0).svg instead of the non-cc-symbol). The list of those serlo compatible licenses can be hardcoded in the frontend repo (they do not change often).

(ping @yolo-bird )

elbotho:
for #596
~need to check if it works for license id 7 after title is changed~

check https://de.serlo.org/license/detail/7 after merge.

### DIFF
--- a/src/components/content/license-notice.tsx
+++ b/src/components/content/license-notice.tsx
@@ -73,10 +73,9 @@ export function LicenseNotice({ data, minimal, type }: LicenseNoticeProps) {
 
   function renderMinimalNotice() {
     const typeString = translateTypeString()
-
     return (
       <MinimalLink
-        href={data.url}
+        href={`/license/detail/${data.id}`}
         title={typeString ? typeString + ': ' + data.title : data.title}
         noExternalIcon
       >


Entkenntnis:
https://frontend-git-license-notice-changes.serlo.vercel.app/license/detail/7 differs from https://de.serlo.org/license/detail/7 (the version in the frontend is much shorter)

Looks like that we have two different versions of this license? (Other licenses, like id 5 or 6, have the same content, maybe there was a recent change somewhere...)

elbotho:
It has a different name in staging that's why I want to test it in staging.

Entkenntnis (approved):

PR #606, opened 2020-10-14
Update API

inyono:
Prepares the repo to use [api.serlo.org@10.0.0](https://github.com/serlo/api.serlo.org/releases/tag/v0.10.0).


elbotho (approved):
thank you (also for fixing type stuff on the way :)

PR #622, opened 2020-11-02
Fix spelling

xnumad:
https://de.wiktionary.org/wiki/lizenzieren


Entkenntnis:
@xnumad Thank you very much for this hint. We are handling our translations here: https://crowdin.com/project/serlo, you can join the project and make your correction there (it will be automatically deployed)

PR #625, opened 2020-11-03
Subject icons for internat. landing pages

### LINKED ISSUE #604: Topic Matemáticas on es.serlo.org starting page with icon.
It would be amazing, if we could have the topic Matemáticas on the es.serlo.org-starting-page (with the same icon as in de.serlo.org), so the starting page would look similar to the german one.

elbotho:
Closes #604


elbotho:
Hey @jakobwes,
could you check [this preview](https://frontend-git-icons-for-other-locales.serlo.vercel.app)? Does that work for you?
(Had to guess the title 😄)

With this change we can now add/change/remove subject icons for the instances quite easily.

Entkenntnis (approved):
Looks good!

jakobwes:
Also just had a look at it and got some feedback from the spanish-LC. We both think it looks really good! Thanks for implementing @elbotho!

PR #627, opened 2020-11-04
use trash with csrf from cookie

Entkenntnis:


Entkenntnis:
needs testing on staging

PR #650, opened 2020-11-22
Unrevised revisions

### LINKED ISSUE #595: Show number of unrevised revisions
Follow up when https://github.com/serlo/api.serlo.org/pull/87 is merged: The number of unrevised revisions can be shown for all types implementing `AbstractRepository` (type of `AbstractEntity` like `Article`, `Video`, ... and `Page`):

grafik

The api query will be:

```graphql
query {
  uuid(...) {
    ...
    revisions(unrevised: true) {
      totalCount
    }
  }
}
```

Entkenntnis:
Change button from edit to history:

grafik

Fixes: #595 

Because of caching, unrevised revisions take a long time to show up.

Live example: https://frontend-git-unrevised-revisions.serlo.now.sh/180846


kulla:
Can we make it clearer that an author needs to see the history because there are unrevised revisions?

Entkenntnis:
What do you think about changing the label to "Work in progress" ("Inhalt in Arbeit")?

kulla:
> What do you think about changing the label to "Work in progress" ("Inhalt in Arbeit")?

How about "Show unrevised revisions"?

PR #674, opened 2020-11-27
add canonical tag

### LINKED ISSUE #672: Add canonical tag
So that query parameters aren't indexed by Google.

elbotho:
@Entkenntnis is there an easier way of constructing a canonical url than:

```jsx
<link rel="canonical" href={`https://${lang}.serlo.org${router.asPath}`} />
```

closes #672

### DIFF
--- a/src/components/head-tags.tsx
+++ b/src/components/head-tags.tsx
@@ -1,6 +1,7 @@
 import Head from 'next/head'
 import { useRouter } from 'next/router'
 
+import { useInstanceData } from '@/contexts/instance-context'
 import { useOrigin } from '@/contexts/origin-context'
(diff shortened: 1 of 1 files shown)


Entkenntnis:
We should actually use the alias from the api response to construct the path, e.g. for a SingleEntityPage, we should add the alias to the fetcher and add the tag in the page. Puh, we would probably have to craft this for every page type ...

inyono:
Nah, not needed. CF Worker is responsible for redirecting to the correct canonical page. The only important thing is that the query params (looking at content API) get removed.

inyono (approved):
LGTM

PR #692, opened 2020-12-02
Toast notice rework and improved content restore UX

elbotho:
- toast notification now called toast notice to avoid confusion with our Notifications
- using queue for notifications so we can display more than one
Also:
- using fetch for the restore content call and add feedback via loading bar and toast notice on success or fail
this is an experiment, but since we hopefully use fetch + a mutation in the future this is how we could handle calls like this.
@Entkenntnis what do you think?

**More options:**
- If the call is successful we could also add a simple refresh: `window.location.reload()` or some local overwrite to display the changes that happened.
- ~we could also use fetch to do the post call (instead of that form) until we get a real mutation :)~ try in the PR

PS: Hard to test in preview and locally. Locally you can at least trigger the loading bar and error message. e.g.
http://localhost:3000/1567

### DIFF
--- a/src/components/navigation/author-tools-hover-menu.tsx
+++ b/src/components/navigation/author-tools-hover-menu.tsx
@@ -301,7 +301,7 @@ export function AuthorToolsHoverMenu({ data }: AuthorToolsHoverMenuProps) {
 
   function trash(id = data.id) {
     // todo: use graphql mutation
-    if (!data.trashed) {
+    if (data.trashed) {
       return (
         <Li>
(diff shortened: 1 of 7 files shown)


elbotho:
experiments in staging: using fetch for the simple calls to legacy works and would improve UX.
Might need some investigating to get the CSRF part working (for "trash") but there is some problem with this function atm. anyway (might be related to unstable staging server).

Entkenntnis (approved):
Nice, lets try this on staging.

PR #694, opened 2020-12-04
new hook to trigger getInitialProps and delete cache partially

elbotho:


elbotho:
Will merge this for better testing.
Should only affect logged in users.

PR #696, opened 2020-12-07
use embed.serlo.org for preview images

elbotho:
can be merged anytime,
but preferably when https://github.com/serlo/serlo.org-cloudflare-worker/issues/66 is a bit further along.


Entkenntnis (approved):
Looks nice

PR #717, opened 2020-12-23
Rework user tools

### LINKED ISSUE #575: Show User Tools on medium devices
Currenty, we show the "Inhalt überarbeiten"-Button and "Weitere Tools" only on large devices. This could be problematic for some medium devices, like tablets or chromebooks, or a split screen on desktops.

I would suggest to show theses two buttons also on medium devices and use the same breakpoint as the main menu.

elbotho:
- Closes #575
- add breakpoints as ints to theme
- add breakpointMax (bp -1px) to theme (to make sure we don't run into randomly unstyled components using media queries with both min and max on the same breakpoint)
- remove unused CookieBar


elbotho:
@Entkenntnis merge if you deem it worthy :)

Entkenntnis (approved):
Thanks for improving the media queries :)

PR #726, opened 2021-01-01
Load exercise on solution id

### LINKED ISSUE #569: Redirection to solution while editing
If you are editing a solution, e.g. https://de.serlo-staging.dev/entity/repository/add-revision/168352, and you save it, you will get redirected to a page with a solution entity. As the frontend is not supporting solutions, this leads to a 404.

Some ideas:

- The cf-worker redirects to the corresponding exercise
- The legacy system is redirection to the exercise directly
- The frontend supports solutions, anyway.

New:

- Don't allow editing an single solution. (remove entry from menu or point to to parent)

Entkenntnis:
Fixes #569

Similiar to course/coursepage, the frontend just loads to exercise of a solution instead of an bare solution. Better than a 404.

### DIFF
--- a/src/fetcher/fetch-page-data.ts
+++ b/src/fetcher/fetch-page-data.ts
@@ -87,6 +87,10 @@ async function apiRequest(
     }
   }
 
+  if (uuid.__typename === 'Solution') {
+    return await apiRequest(`/${uuid.exercise.id}`, instance)
+  }
+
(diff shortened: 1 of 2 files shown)


inyono:
Is this a redirect or duplicate content?

elbotho:
(as discussed in dev-meeting: probably duplicate content, redirect e.g. in cf-worker would be better)

Issue #472, opened 2020-08-13
Author Menu: Show entries depending on role [api]


Entkenntnis:
The Author Menus are complete now (all possible entries are visible).

The next step is to hide all entries (also in user tools) that are not accessible to less priviledged users. The only additional information necessary is the role of the user.

elbotho:
Menu in the comments also needs this info.

Entkenntnis:
Let's collect this information in a document:

https://docs.google.com/document/d/1WVES9v9rlWjMB-NXAeO4ZOxJ3EbEbt1H254D4bpkLrE/edit?usp=sharing

elbotho:
@Entkenntnis could you list what needs to happen to make this work?

Entkenntnis:
Serlo.org issue: https://github.com/serlo/serlo.org/issues/451

@inyono can probably give more information about this.

elbotho:
User roles will be provided via auth (@inyono will add this) but we can get started on the code to adapt the menus.

elbotho:
possible roles, would suggest following strings for now:
``` js
['guest','login','moderator','reviewer','taxonomy-manager','page-builder','admin','sys-admin']
```

note: multiple roles possible!

inyono:
Looks good at first glance. Please add an enum for that so we can easily change that later.

Entkenntnis:
We plan to add an API query to get the list of permissions
todo: link related issue

Entkenntnis:
Another update: We plan to use a npm package that provides a interface to query permissions. This way, the frontend doesn't need to know the actually implementations.

Entkenntnis:
All permissions are available now.

PR #732, opened 2021-01-04
Implement roles (mocked for now)

elbotho:
(for #472 but no fix until it's implemented in hydra https://github.com/serlo/serlo.org/issues/451)


Entkenntnis (approved):
Very nice! One step closer to the milestone :)

PR #736, opened 2021-01-05
add notices when revision hashes present

### LINKED ISSUE #730: Use new edtr redirect hashes to display notices [legacy]
~If https://github.com/serlo/serlo.org/issues/553 is merged we can use the hashes to display notices in the frontend and maybe purge the cache and reload.~

**Status**: There seems to be a bug in staging prevents us from testing this.

> Closes #553
> 
> This PR implements (ping @elbotho ):
> 
> Adds #revision-saved when a revision is saved but needs to be reviewed
> Adds #revision-accepted when a revision was successfully checkout out / accepted
> Adds #revision-rejected when a revision was successfully rejected
> Adds #revision-saved-accepted when a revision was successfully saved and accepted

elbotho:
need i18n and testing, maybe an well timed refresh.
but staging is pretty slow atm so I'll try another time.

### DIFF
--- a/src/components/toast-notice.tsx
+++ b/src/components/toast-notice.tsx
@@ -11,16 +11,43 @@ export function ToastNotice() {
     background: theme.colors.brand,
     text: '#fff',
   }
-  const showTime = 2000
+  const showTime = 4000
+
+  function removeHash() {
+    history.replaceState(null, '', window.location.href.split('#')[0])
+  }
+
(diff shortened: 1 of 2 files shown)


inyono:
Staging should work better again (see also slack).

Entkenntnis:
Still can't save on staging: server just timeouts

PR #740, opened 2021-01-06
check if ids are valid signed 32 bit ints

### LINKED ISSUE #731: Make proper 32-int check for ids
Follow up of #720.

Only prettify ids that are valid 32-bit signed ints.

elbotho:
> I'd change that logic so that it is future-proof (i.e. instead of checking the length, I'd check if this is a 32bit signed integer).

closes #731


Entkenntnis (approved):
Let's see if we someday surpass this number :)

elbotho:
#yologoals

PR #754, opened 2021-01-12
Move instance data out of initial props

Entkenntnis:
Finally found a good way to load translation data into the client. The trick was to think out of the box, e.g. out of react.

The instance data is not anymore part of the page props. Instead, the data is added to the html output in `_document.tsx`. This way, its only loaded on full page load, and not while csr.

@elbotho This is the second step of the refactorings. Can you take a look it?


elbotho (approved):
It sure feels a bit weird to do it this way :D

Entkenntnis:
It felt a bit obvious as I saw this approach here: https://stackoverflow.com/questions/62236258/hydrating-ssr-react-app-with-react-i18next-causes-flicker

PR #755, opened 2021-01-12
Use staticProps and avoid usage of getInitialProps

### LINKED ISSUE #743: Replace getInitialProps with newer methods
I will try to move away from getInitialProps and use newer methods like getServerSideProps, getStaticPaths, etc.

### LINKED ISSUE #746: Investigate chunk optimization
next.js has changed their chunk optimization in the latest release. But an analysis of our chunks revealed that they are not optimal.

The reason could be that we are loading all pages, static and dynamic, from one catch-all page. Maybe this is irritating the optimizer. Idea: Use separate pages for static pages to give optimizer better hints.

Entkenntnis:
Edit3: This PR is getting ready. Changes:

* All pages are using getStaticProps (instead of getInitialProps)
* Use next.js i18n routing feature for subpath routing (instead of homebrew code)
* Map special pages to files, optimizing chunk sizes (first load js for [...slug] down to 150kb (from 170kb))

Will need some more testing, though.


elbotho:
Maybe we should add the `prefetch={false}` to the `<NextLink>` for now :)

Entkenntnis:
Yeah, such extensive prefetching is probably not what we need right now, but it's cool to see what's the performance limit for page navigation (and this is probably the fastest you can get).

elbotho:
we could do some fancy manual prefetch on hover in the future:
https://nextjs.org/docs/api-reference/next/router#routerprefetch

And maybe prefetch some important routes.

src/pages/_document.tsx
```diff
@@ -43,16 +43,11 @@ export default class MyDocument extends Document {
...
+      : undefined
     return (
-      <Html
-        lang={
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          (this.props.__NEXT_DATA__.props.pageProps?.instanceData
-            ?.lang as string) ?? 'de'
-        }
-      >
+      <Html>
```
elbotho, line 50: nice bonus 👍

src/pages/_app.tsx
```diff
@@ -55,16 +58,53 @@ Router.events.on('routeChangeStart', () => {
...
-        <ToastNotice />
-      </ThemeProvider>
-    </React.StrictMode>
+const FrontendClientBase = dynamic<FrontendClientBaseProps>(() =>
+  import('@/components/frontend-client-base').then(
+    (mod) => mod.FrontendClientBase
   )
+)
+
+// a very specific monkey patch for the next.js router usage of history api
```
elbotho: okay, no idea what happens here. 
why do we need this?

src/pages/spenden.tsx
```diff
@@ -0,0 +1,20 @@
+import { GetStaticProps } from 'next'
```
elbotho, line 1: did splitting up improve the chunks?

src/pages/[...slug].tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  if (page.kind === 'donation') {
+    return null
+  } else {
+    // all other kinds are using basic layout
+    // render it together to avoid remounting
+    return (
+      <HeaderFooter page={page}>
+        {(() => {
+          if (page.kind === 'landing') {
```
Comment (elbotho) [inline src/pages/[...slug].tsx]: can we get rid of the `return null` options?

elbotho:
@Entkenntnis looks very promising! Thanks for all the work 👍

src/pages/_app.tsx
Entkenntnis: added a comment

Entkenntnis:
Interesting fact: even with prefetch={false}, the json is fetched on hover already :+1:

PR #757, opened 2021-01-13
Load injection from static props

### LINKED ISSUE #756: Let injections use static props
Injection can load the page props from `/_next/data` as json so they can share the same cache.

Entkenntnis:
Example page: https://frontend-git-load-injections-from-static-props.serlo.now.sh/mathe/27801/addition-und-subtraktion-von-dezimalbrüchen


elbotho:
if stuff is in vercel cache it's spooky how fast this is now!
(injections but also navigation in general)

PR #761, opened 2021-01-14
Optimize Font loading and reduce FOUT

Entkenntnis:
Following the recommandations here: https://kirazhang.com/posts/nextjs-custom-fonts


Entkenntnis:
Very smooth in chrome, stupid FF is not supporting preloading, haven't tested safari yet.

elbotho:
Nice, thanks for working on this!

Sidenote: While testing with simulated "Slow 3G" I saw that we should probably add a min-height to our main container.

elbotho (approved):
👍 LGTM

PR #764, opened 2021-01-14
use correct embed... domain in staging and dev

elbotho:


elbotho:
when https://github.com/serlo/serlo.org-cloudflare-worker/pull/128 is in production also, we can think about a hotfix to production also.

src/helper/cloudflare-worker-domain.ts
```diff
@@ -0,0 +1,4 @@
+export const cloudflareWorkerDomain =
```
Entkenntnis: This value is exactly the same as `serloDomain`, is a separate cloudflareWorkerDomain necessary here?

Entkenntnis (changes requested):
Would

src/helper/cloudflare-worker-domain.ts
elbotho: Thanks, that's why I did not commit it directly.
I searched for that but didn't find it…
yeah, serloDomain should be fine :)

Entkenntnis:
As a matter of fact, the current production code already uses `embed.serlo.org`, so it will get the changes as soon as cf worker updates.

PR #772, opened 2021-01-16
Markup links in comments and make sure they line-break

### LINKED ISSUE #770: long link in comment breaks layout on mobile
example:
https://de.serlo.org/mathe/29577/gesetz-der-gro%C3%9Fen-zahlen 

(we should probably shorten links via js, and make sure that they can't break the layout with css)

elbotho:
closes #770

Not sure about the need to escape.
Chose not to use `<Link/>` to keep it simple. 
(downside: serlo links will be treated like external links, but's still better than link-stings? 😄)


Entkenntnis (approved):
Nice, it's fine to treat all links the same.

PR #774, opened 2021-01-16
Use SWR for isSubscribed check and comments

### LINKED ISSUE #737: Client-side cache for comment queries
Add a session storage cache for client-side comments to reduce server load. (probably with swr for authenticated users)

elbotho:
Trying it, works fine so far.


Entkenntnis (approved):
Nice work. So the only change for production is that comments are fetched with useSWR?

elbotho:
> Nice work. So the only change for production is that comments are fetched with useSWR?

Should be, yes. And maybe that it's not well tested :)

Issue #800, opened 2021-01-22
Show single comment?


Entkenntnis:
Just noticed that urls like https://de.serlo.org/191548 are not implemented at all.

Is this something that we want to add - or just let legacy handle this?

Entkenntnis:
ups

PR #803, opened 2021-01-22
support scrolling to comment with anchor link

elbotho:
[example](https://frontend-git-scroll-to-comment.serlo.now.sh/mathe/1543/kathete#comment-124814)

notes:
- a bit complicated because of the lazy loading… (ping #800)
- also works with CSR
- hint: does not work on reload

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,8 @@
     "styled-components": "^5.0.0",
     "swr": "^0.3.0",
     "timeago-react": "^3.0.0",
-    "timeago.js": "^4.0.0"
+    "timeago.js": "^4.0.0",
+    "tippy.js": "^6.2.7"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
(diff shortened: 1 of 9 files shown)


Entkenntnis:
Super cool feature. For me, it's working on reload.

To make this feature complete, we could:

- add entry to comment hover menu like "Link zu diesem Kommentar"
- add logic to cf worker that redirects ids to the proper path/hash link

Entkenntnis (approved):
Nice! Will merge after deployment. It took me some time to understand how "Link to this comment" works, though.

elbotho:
> Nice! Will merge after deployment. It took me some time to understand how "Link to this comment" works, though.

yeah, maybe we need to rename that function… It was the easiest way to implement this for now. 
Or we do it like the function here on GitHub ("Copy Link")?

Entkenntnis:
The github implementation looks quite straight forward.

elbotho:
@Entkenntnis changed behaviour, also menu now closes after action.
(more annoying to implement than It should be ^^ be that's something I'll probably add to all menus now that we will rely on mutations more)

PR #805, opened 2021-01-23
Refactor pages

elbotho:
Most important change:  Moved `FrontendClientBase` and `HeaderFooter` to `_app`. 
This way we can always access i18n strings and have less renders again.

@Entkenntnis any objections? This way it should not screw with packets, right?

This is how a simple page looks now, I like ^^
``` tsx
import React from 'react'

import { Search } from '@/components/pages/search'

export default function Page() {
  return <Search />
}

```

### DIFF
--- a/src/components/content/entity.tsx
+++ b/src/components/content/entity.tsx
@@ -78,7 +78,7 @@ export function Entity({ data }: EntityProps) {
     if (!data.title) return null
 
     return (
-      <StyledH1 extraMarginTop itemProp="name">
+      <StyledH1 spaceAbove itemProp="name">
         {data.title}
         {renderEntityIcon()}
       </StyledH1>
--- a/src/components/content/page-title.tsx
+++ b/src/components/content/page-title.tsx
@@ -1,0 +1,22 @@
+import Head from 'next/head'
+import React from 'react'
+
+import { StyledH1 } from '../tags/styled-h1'
+
+type PageTitleProps = React.PropsWithChildren<{
+  headTitle?: boolean
+  title: string
+}>
+
+export function PageTitle({ headTitle, title }: PageTitleProps) {
+  return (
+    <>
+      {headTitle && (
+        <Head>
+          <title>{title}</title>
+        </Head>
+      )}
+      <StyledH1 extraSpaceAbove>{title}</StyledH1>
+    </>
+  )
+}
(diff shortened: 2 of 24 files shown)


Entkenntnis:
This is something I have been experimenting with a lot. My conclusion in the end was to keep `_app` as slim as possible and let pages add `FrontendClientBase` explicitly. It was a very close decision (and I also really like how simple pages can be), but the main arguments were:

- It's very difficult to opt-out of code within `_app`, as you may have noticed here: https://github.com/serlo/frontend/blob/8dfcd77f6f3c11066b6cbf602969707e0b31fbbd/src/pages/_app.tsx#L66
- It breaks pages like https://frontend-git-refactor-pages.serlo.now.sh/___graphql
- Renderwise, there is no difference - as far as I can tell.

But I have an idea on the middle-ground: We can make FrontendClientBase more complete by adding HeaderFooter, MaxWidthDiv and RelativeContainer to it (and props to disable them) - so a page will look like this:

```jsx
<FrontendClientBase>
    <Search />
</FrontendClientBase>
```

or

```jsx
<FrontendClientBase noHeaderFooter>
    <Donations />
</FrontendClientBase>
```

elbotho:
Hmm, thanks for the explanation!

What I was shooting for:

``` tsx
export default function Page() {
  const data = someFetcher()
  const auth = useAuth()
  
  const output = auth.current === null ? <PleaseLogIn /> : !data ? <Loading /> : <NotificationsList />
  
  return (
    <>
      <PageTitle>{ /* uses  i18n */} </PageTitle>
      {output}
    </>
}
```

Then the components like NotificationsList could really be components.

Otherwise I will probably mirror the structure in `pages` and `components/pages` and do the same there.

Entkenntnis:
Ah, I see. Maybe not as neat, but avoiding the mirroring between `pages` and `components/pages`:

```jsx
export default function Page() {
  return <FrontendClientBase>
    <Content />
  </FrontendClientBase>
}

function Content() {
  const data = someFetcher()
  const auth = useAuth()
  
  const output = auth.current === null ? <PleaseLogIn /> : !data ? <Loading /> : <NotificationsList />
  
  return (
    <>
      <PageTitle>{ /* uses  i18n */} </PageTitle>
      {output}
    </>
}
```

elbotho:
thanks @Entkenntnis , that's good enough!

Issue #809, opened 2021-01-24
Use separate page for user profiles


Entkenntnis:
We can add a page like `/user/[id]/[name].tsx` to handle user profiles and moving them out of slug to improve bundle size optimization.

PR #816, opened 2021-01-25
Revision path

### LINKED ISSUE #812: Revision path and separate page
revisions usually use links like `/entity/repository/compare/{entityId}/{revisionId}`. 
Those links do not work atm.

Direct links to `/{revisionId}` should probably be handled by the CFWorker?

Like #809 we could put revisions into their separate page, to improve bundle sizes and simplify the main query.

elbotho:
- support for eg. `entity/repository/compare/1639/112398`
- remove revision stuff from query and fetcher
- requests to `/{revisionId}` get redirected for now


elbotho:
@Entkenntnis if that is what you had in mind I will do the same changes for users #809

Entkenntnis:
Yes! This is nice - and I really like the fact that we can move revision-related stuff out of our main query :+1:

Entkenntnis (approved):
This was fast :) Looks fine, so let's test it on staging together with all the other pages.

PR #819, opened 2021-01-25
Separate user profile

### LINKED ISSUE #809: Use separate page for user profiles

elbotho:
okay, nice, slug is quite manageable now.

weirdly the first load js did not change?

``` diff
-  ├ ● /[...slug]                                                 1.26 kB         148 kB
+  ├ ● /[...slug]                                                 1.19 kB         148 kB
```


Entkenntnis:
This is normal, because the user profile is a dynamic component. (and the number is a bit "better than reality")

Entkenntnis (approved):
Thanks

PR #822, opened 2021-01-26
Improve code-splitting and dynamic imports

### LINKED ISSUE #810: Investigate dynamic imports and only make non-trival modules dynamic
Dynamic imports have their overhead I I think we are overusing them in some places ...

Entkenntnis:
Still a long way to go.

### DIFF
--- a/src/components/author/revision.tsx
+++ b/src/components/author/revision.tsx
@@ -1,14 +1,13 @@
 import { faList } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import dynamic from 'next/dynamic'
 import React from 'react'
 import ReactDiffViewer from 'react-diff-viewer'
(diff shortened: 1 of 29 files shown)


Entkenntnis:
First two candidates: Math and Comments, brings bundle down to 203kb.

Entkenntnis:
Extracting logged in components to separate bundle, coming down to 185kb. This is the state of [...slug]:

grafik


Next steps: Modal + algebra.js (or tippy?)

Something buggy with fontawesome: We don't have separate bundles for icons, so we can only load all icons we use at once. This is taking up 3 of the bundles.

Entkenntnis:
Just dynamically loading tippy brings us down to 172kb.

Entkenntnis:
It looks good again:

grafik

I will merge this PR now, because I want to avoid that it becomes to big, there will be same following clean up issues - and a lot to communicate about how to handle dynamic imports.

PR #830, opened 2021-01-26
New mutations

elbotho:
- createThread, createComment and setThreadArchived work quite nicely!
- comments need a refactoring… (bit of a mess}
- setState mutations don't work (https://github.com/serlo/api.serlo.org/issues/213)

Atm. I do no real cache-mutation but query all thread data again after api call. Surprisingly fast on staging.


elbotho:
@Entkenntnis parts of it are still WIP but merge if you like
(so that we run into big conflicts with all the refactoring at the moment :)

Entkenntnis:
Yeah, let's merge this into staging to avoid conflicts.

PR #832, opened 2021-01-27
Readd persistent layouts

### LINKED ISSUE #823: Persistent layouts with static property on page
Ups, after reading this https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/ (again) I think the best solution for us would be Option 3 or 4 with a static property on the component.

### LINKED ISSUE #829: rework data-types
Instead of initialprops, we now have a lot of different routes (with different props). Adapt the data types to reflect the new architecture.

Entkenntnis:
Will fix #823, #829.

Some interesting properties arise.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -117,7 +117,6 @@ module.exports = {
     'react/jsx-fragments': ['error', 'syntax'],
     'react/prop-types': 'off',
     'react/react-in-jsx-scope': 'off',
-    'react/display-name': ['error', { allow: ['@/pages/**'] }],
 
     // eslint-plugin-react-hooks
     'react-hooks/exhaustive-deps': 'warn',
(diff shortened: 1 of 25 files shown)


Entkenntnis:
@elbotho This on is ready now. The main change is the function `renderedPageNoHooks` that takes a render function. Everything within this function can be persisted between pages. It's quite simple and types are working fine. The downside is that you can not use hooks in the function.

elbotho (approved):
Looks very nice,
and thanks for cleaning up the types!

inyono:
>  The main change is the function `renderedPageNoHooks` that takes a render function. Everything within this function can be persisted between pages. It's quite simple and types are working fine. The downside is that you can not use hooks in the function.

What is the reason for passing the argument as a function (instead of a React component which would allow hooks etc.)?

Entkenntnis:
Every page is a different component. Rendering with this component will remount everything within that component, including header + footer.

Example: Type something into the search input on the landing page and navigate to /mathe -> the search input is reset

Entkenntnis:
You can still easily use hooks if you need to:

grafik

PR #837, opened 2021-01-27
Mutations follow up

elbotho:


Entkenntnis (approved):
Nice, was just about to implement entityId myself

PR #849, opened 2021-01-30
New jsx transform

### LINKED ISSUE #848: Default React import not needed any more
According to [this](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) we could kick some `import React from 'react'` if we update some of our deps and config.

We might get `performance improvements and simplifications` 🚀 💥 

---

Also in future versions there will not be a default export any more, so changing to `import * as React from 'react'` or the (preferred) `import { useState } from 'react'` now is encouraged.

elbotho:
closes #848

@Entkenntnis any objections? (Reasoning in #848)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,9 @@
 {
   "presets": ["next/babel"],
-  "plugins": [["styled-components", { "ssr": true }]]
+  "plugins": [["styled-components", { "ssr": true }],[
+      "@babel/plugin-transform-react-jsx",
+      {
+        "runtime": "automatic"
+      }
+    ]]
 }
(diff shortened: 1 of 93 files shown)


.babelrc
```diff
@@ -1,4 +1,12 @@
 {
   "presets": ["next/babel"],
-  "plugins": [["styled-components", { "ssr": true }]]
+  "plugins": [
+    ["styled-components", { "ssr": true }],
+    [
+      "@babel/plugin-transform-react-jsx",
```
inyono: Not needed, see https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#nextjs.

elbotho: Thanks!

PR #850, opened 2021-01-30
move toast-notice to module, remove context

elbotho:
Note: @Entkenntnis Last commit moves the instance data to a global variable instead of a JSON. 
This way we can access it pretty nicely with `window.__CLIENT_INSTANCE_DATA__` in the frontent-client-base.
Looks a bit neater, but can also revert if you prefer


Entkenntnis (changes requested):
I'll wait for this after current deployment. Bundle size for slug went up by 26kb -> isServer seems to break server-side detection.

elbotho:
> isServer seems to break server-side detection

I knew there had to be a downside, looked to pretty ;)
Will remove it again.

elbotho:
PS: Thanks for keeping an eye on the bundle sizes!
 Could we maybe add this to CI to see how PRs will change bundles?

elbotho:
(part of #852 now)

Issue #851, opened 2021-01-30
Review auth


inyono:
- [ ] When using multiple tabs, the actually used token may be stale (e.g. if it is was refreshed in another tab earlier). We should probably sync it with the current cookie value before doing the request.
- [ ] The login state itself should be a state instead of a ref (so that changes to it trigger a React update)

Entkenntnis:
List of (imho minor) problems:

- If I log in from a tab, other tabs will sync after reload or navigation, same for logout. I haven't encounter strange states.
- Mutations (e.g. create comment) will fail, but show a message that you need to login

Nothing dramatic. I am not really sure how to fix these cases without actively polling the cookie: For the user, the mutation will fail anyway, the difference would be only if the client is handling it or the server response ...

Opinions here?

elbotho:
Currently If I log in in one tab, the other tab just hides the user menu completely. 
Reload fixes it, but navigation not. Logout actually works better.
![image](https://user-images.githubusercontent.com/1258870/117970050-7e106700-b328-11eb-9239-abf5f24b763c.png)
Maybe we could solve that particular issue instead, but atm. it's … strange :)

inyono:
Can look into that after the db layer bug hunting.

Entkenntnis:
@elbotho Can confirm, this is happening in production - really strange. I have opened a separate issue for this and will close this general discussion.

(Feel free to open new issues for specific UX bugs)

elbotho:
There is also a bug, where I'm logged in but the frontend does not show it. 
Not sure how to reproduce it yet.

PR #852, opened 2021-01-30
Simplify mutations usage / Error Handling

### LINKED ISSUE #459: Use mutations to trash/restore content [api]
Currently, the trash menu entry is a form that makes a POST request. Should we replace this button with a graphql mutation?

### LINKED ISSUE #729: Add functionality to comment admin menu [api]
The UI is ready, but the functionality for trash / restore /  archive comments is not implemented yet.

### LINKED ISSUE #834: Overlong title is overflowing
On mobile, this page https://de.serlo-staging.dev/community/145470/veranstaltungs%C3%BCbersicht has a title that overflows the page.

### LINKED ISSUE #854: Add option to unarchive threads

elbotho:
**Goals:**

-> Easy mutations usage in components:
``` tsx 
const setUuidState = useSetUuidStateMutation()
setUuidState({})
void setUuidState({ id: [data.id], trashed: !data.trashed })
```

-> Readable mutation hooks: (check [mutations.ts helper](https://github.com/serlo/frontend/pull/852/files#diff-3ac9da7467e0e47e721d487456ac175b28d3fb61a67a6969d9bf1cb5f658b855R53-R77))

-> Error handling that provides user-friendly feedback.


src/helper/mutations.ts
```diff
@@ -228,44 +159,108 @@ export function useCreateThreadMutation() {
...
+type ApiErrorType = 'UNAUTHENTICATED' | 'FORBIDDEN'
+type ErrorType = ApiErrorType | 'UNKNOWN'
+
+export interface ApiError extends GraphQLError {
+  extensions: {
+    code: ApiErrorType
+  }
+}
+
+export async function mutationFetch(
```
elbotho, line 234: @inyono its is the relevant function (that I use in place of `createAuthAwareGraphqlFetch` atm.)

src/components/comments/dropdown-menu.tsx
```diff
@@ -88,14 +88,14 @@ export function DropdownMenu({
   function onDelete() {
     onAnyClick()
     if (!isParent) {
-      void setCommentState({ id, trashed: true })
+      void setCommentState({ id: [id], trashed: true })
```
inyono, line 99: Is this necessary? Shouldn't GraphQL accept a single id, too?

src/components/frontend-client-base.tsx
```diff
@@ -40,13 +44,11 @@ export function FrontendClientBase({
     if (typeof window === 'undefined') {
       // load instance data for server side rendering
       // Note: using require to avoid webpack bundling it
+      // eslint-disable-next-line
```
inyono: Why? You should at least be specific here and only disable the rules that aren't useful for this special case.

src/components/user-tools/author-tools.tsx
```diff
@@ -382,7 +380,7 @@ export function AuthorTools({ tools, entityId, data }: AuthorToolsProps) {
             }
 
             setTimeout(() => {
-              refreshFromAPI()
+              csrReload()
```
inyono, line 383: That looks like a hack. Don't we know the state after the GraphQL query anyway and can update that w/o forcing a refresh?

src/helper/mutations.ts
```diff
@@ -228,44 +159,108 @@ export function useCreateThreadMutation() {
...
+            Authorization: `Bearer ${auth.current.token}`,
+          }
+        : {},
+    })
+    // … okay for now
+    return (
+      JSON.stringify(await client.request(query, { input })).indexOf(
+        '"success":true'
+      ) > -1
+    )
```
inyono: Not a fan. If you treat all the mutations the same way, checking the `error` property is IMHO much safer.

src/helper/mutations.ts
elbotho: (https://github.com/serlo/frontend/blob/937faf1c80f1decf1f719b6ed3313ae5325d5d0c/src/helper/mutations.ts#L203)

src/helper/mutations.ts
```diff
@@ -228,44 +159,108 @@ export function useCreateThreadMutation() {
...
+    }
+
+    return handleError(type)
+  }
+
+  async function executeQuery(): Promise<boolean> {
+    const client = new GraphQLClient(endpoint, {
+      headers: auth.current
+        ? {
+            'access-control-allow-origin': '*',
```
inyono: Does this even do anything? AFAIK access-control-allow-origin is only relevant as a response header (i.e. for the server, not for the client)

elbotho:
https://github.com/serlo/frontend/blob/937faf1c80f1decf1f719b6ed3313ae5325d5d0c/src/helper/mutations.ts#L239
Irgendwie nach einigem überlegen die einfachste Lösung, vetos? 😸

src/helper/mutations.ts
inyono: Why do we even mutate the response? shouldn't the actual query decide what it wants to have returned?

src/helper/mutations.ts
```diff
@@ -228,44 +159,108 @@ export function useCreateThreadMutation() {
...
+      : 'Ein unbekannter Fehler…'
+
+  if (type == 'UNKNOWN') {
+    console.log(e)
+    triggerSentry({ message: 'Unknown API error' })
+  }
+
+  if (type == 'UNAUTHENTICATED') {
+    //make sure correct auth state is showing
+    csrReload()
```
inyono, line 287: Ping #851. Not a fan of this hacks tbh, since they allow us to be sloppy about the state handling (instead of fixing the actual problem why this doesn't work right now).

src/components/comments/dropdown-menu.tsx
elbotho: It accepts single id's but the type does not. Maybe a problem we can solve with `regenerate-types`?

src/components/frontend-client-base.tsx
elbotho: I thought that might be a bit verbose… :) 
```
// eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```

src/components/user-tools/author-tools.tsx
elbotho: ~Follow up in https://github.com/serlo/frontend/issues/851~
sorry wrong comment. 

Jup it's a hack. Cache management and mutations are still WIP.

src/components/frontend-client-base.tsx
inyono: Hm. Well, that's an annoying case then. I'd still prefer the verbose variant to document which rules fail here (so that e.g. a newly introduced rule still triggers here possibly and the next one that touches the code can possibly review that).

elbotho: Check.

src/helper/mutations.ts
elbotho: It's specifically a mutation helper and currently we only use mutations in this way: Only querying success.
And errors should be handled already, so I could just return true instead?

src/helper/mutations.ts
elbotho: Should probably remove it in `createAuthAwareGraphqlFetch` as well then.

src/components/comments/dropdown-menu.tsx
inyono: Interesting. That's kinda annoying. Feel free to create an issue in api.serlo.org for that. Not sure whether there is an easy fix though since we just extract that from the GraphQL schema (so possibly a feature request for the library that does the extraction).

src/helper/mutations.ts
inyono: Well, I kinda get that but I also dislike those kind of implicit assumptions (Hey, every mutation needs to return "success": true or frontend possibly fails). When you handle the errors anyways, why not just return nothing or `return true` if you prefer that.

src/helper/mutations.ts
elbotho: Okay, we leave it (because it's better than the current state) and try to find a fix in #851.

src/components/comments/dropdown-menu.tsx
elbotho: Jup, I kind of wish I knew that before I [advocated this](https://graphql-rules.com/rules/mutation-batch-changes).

elbotho: follow up in https://github.com/serlo/api.serlo.org/issues/230

elbotho:
@Entkenntnis  (no rush, review after deployment is fine)

elbotho:
This PR will close all but one issue with our current milestone 😸 
https://github.com/serlo/frontend/milestone/4

Entkenntnis:
I have reverted the instance-data change. The reason is that I want to stay consistent with how next.js handles its data.

Entkenntnis (approved):
Thank you for this work. We are almost at our next milestone. (I hope my conflict solving has not caused problems?)

elbotho:
Thanks fore reviewing and merging!

PR #861, opened 2021-02-04
Notification page changes

### LINKED ISSUE #858: Notifications: Handle unread and read differently
currently we often encounter old unread notifications. 
this results in users expecting new notifications but only seeing read ones.

elbotho:
We now have tabs to switch between unread and read notifications.

closes #858


Entkenntnis (approved):
Thank you!

PR #862, opened 2021-02-04
Comments: add invitation to log in or start thread

### LINKED ISSUE #845: Add message: Log in to write commet
Add a message to the commets that only logged in users can write comments, like "Melde dich an um einen Kommentar zu schreiben".

Also show the heading "Kommentare" if no comments are present.

elbotho:
- Always shows either form to start a new thread or the invitation to log in to do so.
- Uses `<PleaseLogIn>` component that uses a generic "log in to use this function" that we could make more specific in the future.

closes #845


Entkenntnis (approved):
Nice!

PR #863, opened 2021-02-04
add prop to act on loaded

Entkenntnis:
Fix for issue when you search directly and haven't given consent yet.

@elbotho Can you verify this works for you?


elbotho:
@Entkenntnis jup! Works for me.

PR #867, opened 2021-02-05
add filter to subscription manager

elbotho:


Entkenntnis (approved):
Merging into staging to make this better testable.

PR #868, opened 2021-02-06
add reject and accept feature

elbotho:
@Entkenntnis the auth check and extra component is probably not enough for `<CheckoutRejectButtons>` not to get bundled, right?


Entkenntnis:
Right, the modal is bundled. To change this, a dynamic component for `StyledModal` should solve the issue. Do you want to try this? In theory, the bundle size should go down to ~170kb...

elbotho:
@Entkenntnis okay I used dynamic for the whole button component (since we won't load that for users that are not allowed to decide). 
Bundle size down again.

PR #871, opened 2021-02-06
lazy: adapt placeholder height, increase offset

elbotho:
closes #871


Entkenntnis:
Thanks! Works nicely now.

PR #879, opened 2021-02-07
make license notice detailed for non cc licenses

elbotho:
My draft for #636, #870

- Default license: no notice
- Other Licenses: `[icon] [typename]: Different License` all linked to license page with additional hover info for desktops.
- Icon changes depending on CC or non CC

This way we inform users very clearly that the license for this content (relation with easy to understand typename) is different from the usual. Imho this is enough as a first step. For more info ppl have to click / hover. 

Non-cc license under video: http://localhost:3000/1515
CC-License in exercise injection: http://localhost:3000/1565

image
image

Did not find a non default solution yet.
 We could change `Aufgabe` to `Aufgabenstellung` to make that relation even more clear. But only if it's @inyono wish :)

PR #881, opened 2021-02-08
introducing concept of "inline license" and add converter for it

### LINKED ISSUE #873: Let video node handle license info
Currently, the license of a video is handles by injections - this prevents us from building some finer grained solutions. To solve this, we can add the license information to the video directly (in the fetcher, similiar to exercises). This way, injections don't need to handle licenses in general - and videos can show a more clearer notice.

Entkenntnis:
fixes #873


Entkenntnis:
will be merged together with #879

PR #899, opened 2021-02-17
Revision fixes

elbotho:
adds some support for exercises in revisions.
solutions, link to solutions etc. still missing.

@Entkenntnis you are welcome to take this over if you have time :)


Entkenntnis:
We are almost compatible to legacy with the exception of #919 - which I will probably add in another PR.

PR #909, opened 2021-02-22
Handle /user/profile/... as legacy links

### LINKED ISSUE #898: When clicking on profile-links on the /team-page I get an error
On here https://de.serlo.org/team I get an error "This page (https://de.serlo.org/de/user/profile/Adrian_B) is currently offline. However, because the site uses Cloudflare's Always Online™ technology you can continue to surf a snapshot of the site." and get rerouted to the old frontend. If I click on "Retry for a live version" then it works, but I only get the old frontend.

kulla:
Fixes https://github.com/serlo/frontend/issues/898

### DIFF
--- a/src/components/content/link.tsx
+++ b/src/components/content/link.tsx
@@ -33,7 +33,6 @@ const legacyLinks = [
 export function isLegacyLink(_href: string) {
   // compat: this is a special frontend route or force frontend use
   if (_href == '/user/notifications') return false
-  if (_href.startsWith('/user/profile/')) return false
   if (_href.startsWith('/entity/repository/history')) return false
   if (_href.startsWith('/entity/repository/compare')) return false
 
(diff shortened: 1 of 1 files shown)


inyono:
Shouldn't `/user/profile` be a non-legacy link (i.e. CSR-handled link) since the route is handled by the frontend?

Entkenntnis:
What's unclear for me right now is which link should be the "canonical" user profile link? If it's not /user/profile, then it can make sense to exclude it from csr and allow the cf worker to redirect it.

kulla:
> Shouldn't `/user/profile` be a non-legacy link (i.e. CSR-handled link) since the route is handled by the frontend?

You mean since it can be passed into `uuid(alias: ...)` as an alias an get resolved there?

Entkenntnis:
It makes sense to disable csr for /user/profile. In staging and production, the frontend will pass the url to cf worker and will get a proper redirect. In development, #912 makes sure that the page is still rendered.

A full fix would extend our prettify-logic to also transform these links.

PR #912, opened 2021-02-24
user page accept any alias

### LINKED ISSUE #898: When clicking on profile-links on the /team-page I get an error

Entkenntnis:
support alias like `/user/profile/<name>`

Fixes #898

### DIFF
--- a/src/fetcher/user/query.ts
+++ b/src/fetcher/user/query.ts
@@ -1,8 +1,8 @@
 import { gql } from 'graphql-request'
 
 export const userQuery = gql`
-  query userUuid($id: Int) {
-    uuid(id: $id) {
+  query userUuid($path: String!, $instance: Instance!) {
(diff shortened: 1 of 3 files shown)


elbotho:
thanks!
looks good so far. some cleanup and then this should be done - nice! 🎉 

one small thing: `/user/{id}/{usernameOrRandomString}` works fine but `/user/{id}` does not.
not strictly necessary but probably easy to add.

Entkenntnis:
`/user/<id>` is not handled by the API :eyes: so I suppose it's not officially supported?

Issue #911, opened 2021-02-23
Bug: Input Exercise does not accept right answer


ghost:
https://de.serlo.org/188902 Teilaufgabe d)

In the legacy system the answer gets accepted
![Screenshot_2021-02-23 Die Länge](https://user-images.githubusercontent.com/46370375/108816913-54f8f980-75b7-11eb-90e6-536b452dd1a8.png)

The now frontend does not
![Screenshot_2021-02-23 Die Länge - lernen mit Serlo ](https://user-images.githubusercontent.com/46370375/108816932-5f1af800-75b7-11eb-94bc-c85bd1cd26e1.png)

Entkenntnis:
Yes, confirmed.

elbotho:
@yolo-bird thanks for reporting this. 
there are answers for both "2,5" and "2,50" which is:
- unnecessary
- confuses our implementation
but obviously:  
- should not confuse our implementation ;)
- will be fixed soon.

Entkenntnis:
is fixed: https://de.serlo-staging.dev/mathe/188902/188902

PR #913, opened 2021-02-24
fix for #911

elbotho:
flying pretty blind without tests here but this solves the bug (but also might open new ones).
using edtr code (at least the logic) for this would be a great first case :)


Entkenntnis:
Fix looks reasonalbe, thanks!

PR #927, opened 2021-03-01
Rename of subject in fr.serlo.org

kulla:
See https://github.com/serlo/serlo.org/pull/597 (ping @jakobwes )


Entkenntnis:
Thanks!

PR #934, opened 2021-03-08
Query always by alias

kulla:
API supports querys of the form `/:id` and thus there is no need to have
different ways to query for a resource.


elbotho (approved):
LGTM! Thanks.

Entkenntnis:
Thanks, sorry for the late merge

PR #943, opened 2021-03-17
Updated with correct menu-links for es.serlo.org

jakobwes:
(not a 100% about the meaning of participationHref, so if someone could double-check that, that would be amazing)


elbotho:
`participationHref` is used for the participation link in the footer. Looks good!

PR #955, opened 2021-04-09
RFC Authorization

inyono:
WIP, still needs work and thorough testing.

Open questions:
- [ ] `useAuthentication` and `useAuthorization` are orthogonal in this implementation. We could also combine those into a single `useAuth`. Don't have a strong opinion either way, though it feels to me like these aren't used together, usually. (You'd use `useAuthorization` if you want the current username / check if a user is logged in. You'd want `useAuthentication` if you want to check if the current user - logged in or guest - may do something. *Note: terms mixed up*

Open TODOS, apart from the comments:
- [ ] Deprecate roles stuff.
- [x] Of course I mixed up `authentication` and `authorization`. So review those ^^


src/auth/use-authentication.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { useGraphqlSwrWithAuth } from '@/api/use-graphql-swr'
+import { useInstanceData } from '@/contexts/instance-context'
+import { Instance } from '@serlo/api'
+
+export function useAuthentication() {
+  // TODO: this should also work if authorization === null
+  const instance = useInstanceData()
+  const scope = instanceToScope(instance.lang as Instance)
+  const { data } = useGraphqlSwrWithAuth<{
```
inyono: TODO: `useGraphqlSwrWithAuth` throws an error if user is not authenticated. Should probably add an option like `requireAuth` that falls back to an request without token.

src/auth/use-authentication.ts
```diff
@@ -0,0 +1,31 @@
+import { AuthorizationGuard, AuthorizationPayload, instanceToScope, Thread } from '@serlo/authorization'
+import { gql } from 'graphql-request'
+
+import { useGraphqlSwrWithAuth } from '@/api/use-graphql-swr'
+import { useInstanceData } from '@/contexts/instance-context'
+import { Instance } from '@serlo/api'
+
+export function useAuthentication() {
```
inyono: This is the hook that may be used somewhere. Usage:
```tsx
import { Thread } from '@serlo/authentication'

function SomeComponent() {
  const mayDo = useAuthentication()
  if (mayDo(Thread.createThread)) {
    return <button>Start thread</button>
  } else {
    return 'Please log in to start a thread'
  }
}
```

inyono:
Might also depend on #851 so that `authorizationPayload` is correctly refetched when login state changes.

src/auth/use-authentication.ts
```diff
@@ -0,0 +1,31 @@
+import { AuthorizationGuard, AuthorizationPayload, instanceToScope, Thread } from '@serlo/authorization'
+import { gql } from 'graphql-request'
+
+import { useGraphqlSwrWithAuth } from '@/api/use-graphql-swr'
+import { useInstanceData } from '@/contexts/instance-context'
+import { Instance } from '@serlo/api'
```
inyono: TODO: `@serlo/api` leaks quite some authentication types because one of the `export *` now also contains some stuff that is not required for the API. We should probably fix those.

inyono:
Open questions:
- [ ] `useAuthentication` and `useAuthorization` are orthogonal in this implementation. We could also combine those into a single `useAuth`. Don't have a strong opinion either way, though it feels to me like these aren't used together, usually. (You'd use `useAuthorization` if you want the current username / check if a user is logged in. You'd want `useAuthentication` if you want to check if the current user - logged in or guest - may do something.

Open TODOS, apart from the comments:
- [ ] Deprecate roles stuff.

src/auth/use-authentication.ts
inyono: Resolved via https://github.com/serlo/api.serlo.org/pull/302.

PR #959, opened 2021-04-12
Tailwind

Entkenntnis:
Adding tailwind to the frontend. This allows for a very incremental approach to move styles over.

Major companies are doing the move and sharing their experiences (e.g. https://www.netlify.com/blog/2021/03/23/from-semantic-css-to-tailwind-refactoring-the-netlify-ui-codebase/).

Requirement: I'll do the move after checking for visual regressions and trying to fix them.


package.json
```diff
@@ -63,6 +65,7 @@
     "simple-oauth2": "^4.0.0",
     "styled-components": "^5.0.0",
     "swr": "^0.4.0",
+    "tailwindcss": "^2.1.1",
```
Entkenntnis: there will be a 3.0 release soon

Entkenntnis:
This looks quite promising so far. Integration is quite done with minimal to no visual differences. The new workflow:

- Within `./src`, use css classes *only*.
- Need some special css? Extend the theme or create a new utility within `tailwind.config.js`.  This will be the only place in the frontend where we actual write bare css.
- Avoid using styled-components. tailwind and next.js are providing all features we need, so this package is a duplicate.
- A rule of thumb: No more than 5 classes per line. Use `clsx` (tiny helper function) to split class names. Can also be used for conditional applying classes, too
- IntelliSense for VSCode is *very* helpful.

Some gotchas:

- There is no concept of extending styles anymore - for the good. You can create components in tailwind instead, which can be overridden with utilities. No need to track down a long inheritance chains.
- Because all css classes are created in `tailwind.config.js`, there is no concept of theme variables (no more `${props => props.theme.colors.brand}` anymore :grin:). Create a new specialised utility instead or use existing ones like `text-brand` or `bg-brand`.
- For edge cases, there is still `serlo-tailwind.css` for global styles

Things to do:

- This PR sets up the foundation and has already some edge cases converted.
- Refactoring every step by step, one component at a time and adding new utilities - completely non-critical, because the two systems coexists nicely

PR #965, opened 2021-04-15
Some refactorings for the TimeAgo component

kulla:


src/components/time-ago.tsx
```diff
@@ -1,22 +1,14 @@
...
-  opts?: timeago.Opts
-  className?: string
+interface TimeAgoProps extends Pick<ReactTimeAgoProps, 'datetime'> {
   dateAsTitle?: boolean
 }
 
-export function TimeAgo({
-  datetime,
-  opts,
-  className,
```
elbotho, line 17: Please check if removing `className` breaks custom styled components (e.g. in notification.tsx we use a `<StyledTimeAgo/>`). This is the reason it was added.

elbotho (commented):
see comment. rest looks a lot nicer, thanks!

src/components/time-ago.tsx
kulla: Is there a way to try the login feature locally?

elbotho: sure, if you don't change the default setup, you should be able to login (via redirect to staging) using the normal login button.

kulla: @elbotho Yes, you are right. Parameter `className` is needed. I have added it again as a paramater.

kulla:
@elbotho Suggestions implemented. Ready for merge.

PR #968, opened 2021-04-16
fix: Fix error in /user/notifications

### LINKED ISSUE #967: Error on /user/notifications: Object as react children
After I logged in as `kathongi` I get the following error on http://localhost:3000/user/notifications for the current staging branch:

2021-04-16-144341_958x964_scrot

kulla:
Fixes https://github.com/serlo/frontend/issues/967


elbotho (approved):
Weird, did not occur before.
Thank you for the fix 👍

PR #990, opened 2021-04-29
Update dependencies

inyono:
- Update CI and type definitions to Node v14 since Vercel (resp. AWS) finally supports the current LTS.
- Update Edtr.io to v2.0.0.
- Update other dependencies. Apart from postcss-nested, which has a fixed version for some reason, all dependencies are up-to-date.


Entkenntnis:
Thank you! postcsss-nested has some trouble with postcss 8, but I hope this will resolve soon.

PR #991, opened 2021-05-01
Demo Quickbar

Entkenntnis:


tailwind.config.js
```diff
@@ -59,6 +70,8 @@ module.exports = {
       },
       fontFamily: {
         serlo: 'Karmilla, sans-serif',
+        homepage: 'Karla',
```
elbotho, line 73: @Entkenntnis Karla used to be unusable, because there were important glyphs missing. There is an update now (finally). I'll investigate if we should switch.

Entkenntnis:
Will be built into the new homepage.

PR #1000, opened 2021-05-10
canonical url: construct without parameters

### LINKED ISSUE #998: Remove query parameters from canonical urls
`head-tags.tsx` L 19

elbotho:
construct from slug instead of filtering string, should be more robust.
fixes #998 

🥳  💯*10


Entkenntnis (approved):
Thanks! Nice solution. And it handles language subpaths :)

PR #1007, opened 2021-05-10
Use new subscription query and mutations

### LINKED ISSUE #999: New API currentUserHasSubscribed()
In staging the new API is deployed:

```graphql
extend type Query {
  subscription: SubscriptionQuery!
}

type SubscriptionQuery {
  currentUserHasSubscribed(id: Int!): Boolean!
}
```

This might be used to shorten some requests in the frontend. See also 
2021-05-10-101102_896x572_scrot

elbotho:
- we now have separate requests for the subscription overview and the author-menu
- uses mutations now 🎉 
- swr keys are a mess, but we can solve that in a followup
- closes #999


Entkenntnis (approved):
Thank you!

PR #1010, opened 2021-05-11
fix license undefined error

elbotho:
for some types (or on an inconsistency server side, not sure) license info can be undefined.

### DIFF
--- a/src/fetcher/create-inline-license.tsx
+++ b/src/fetcher/create-inline-license.tsx
@@ -12,10 +12,7 @@ const shortTitles: { [key: number]: string } = {
   19: 'schule-bw.de',
 }
 
-export function createInlineLicense(
-  license?: License
-): LicenseData | undefined {
-  if (!license) return license
+export function createInlineLicense(license: License): LicenseData {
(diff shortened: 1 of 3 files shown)


inyono:
Can you give some example UUIDs? IMHO, this shouldn't happen...

elbotho:
@inyono Well ok then I'll fix it properly ;) 
We just did not fetch the license data for `GroupedExerciseRevision` because we usually don't display license info on revisions.

elbotho:
example url https://frontend-git-fix-license-error-serlo.vercel.app/entity/repository/compare/0/3936
The `0` in the url should be gone in staging/production I think.

inyono:
> example url https://frontend-git-fix-license-error-serlo.vercel.app/entity/repository/compare/0/3936
> The `0` in the url should be gone in staging/production I think.

That's an independent frontend issue. Not sure where you get the `0` from but this only happens on frontend.serlo.org (which isn't handled via CF Worker redirects).

Entkenntnis:
Thank!

PR #1012, opened 2021-05-11
Revert quickfix: readd .env.local

inyono:
Don't merge until we have a response to our support ticket. Just here to make sure we don't forget to revert this change :)


inyono:
Okay. Apparently .vercelignore is not supposed to work with the Git integration (although it did work at some point...). I changed to build command on Vercel so that .env.local is removed before the build.

PR #1023, opened 2021-05-13
fetch solution revisions

### LINKED ISSUE #930: History view for solutions
The history view is currently not implemented for solutions: https://frontend.serlo.org/entity/repository/history/195107

Entkenntnis:
fixes #930


src/pages/entity/repository/history/[id].tsx
```diff
@@ -237,6 +240,21 @@ const query = gql`
           }
         }
       }
+      ... on Solution {
+        currentRevision {
+          id
+        }
+        solutionRevisions: revisions {
```
Comment (Entkenntnis) [inline src/pages/entity/repository/history/[id].tsx:247]: I needed to add a label, because otherwise, I get an error:

grafik

@kulla:Do you have an idea why this is only happening for revisions of solutions?

PR #1042, opened 2021-05-21
Set notification to read manually, cache revalidation

elbotho:
~basically done, but atm. revalidation (or redraw) after setting notification to `read` only works on the first "page" (10 entries).~

This changes the logic of notifications in the frontend so that users can (and have to) set them to read manually.
@kulla and @inyono preferred this behaviour and also it's a lot more stable that the automatic approach.

Also we now revalidate the local swr cache immediately after the mutation.
I opted for a very optimistic UI, since a failure (single notification is still marked unread after refresh) is not a real problem here. Seems pretty fast this way ⚡ (no delay).
The button to set all visible notifications to read and the notification count are not optimistic but rely on the (slightly delayed) revalidation.

(preview not possible in PR because of auth)


Entkenntnis (approved):
Thank you for working on this! Logic sounds good let's merge this to allow testing on staging

Issue #935, opened 2021-03-08
Implement new mockup for user profiles


kulla:
# Todos:

## Frontend
- [x] Add motivation: 
    - [x] Add edit button
    - [x] Add edit button also when user has no motivation text
- [x] Implement profile pic: 
- [x] Edit link: https://community.serlo.org/account/profile
- [x] Add link to chat
- [x] Fix link for not logged in user
- [x] Check whether user has a rocket chat link 
- [x] Add "logined since"
- [x] Add header
- [x] Link Badges to: https://de.serlo.org/community/202923/rollen-der-serlo-community
- [x] Link Roles to: https://de.serlo.org/community/202923/rollen-der-serlo-community

- [x] Add activity diagram
- [x] Add roles
- [x] Add events
- [x] Add last login
- [x] remove comments
- [x] Suggestion from Simon for better profile structuring

## Frontend: Refactoring

- [x] Refactoring CSS
- [x] Fetch motivation + imageUrl + chatUrl from API
- [x] Add semantic HTML elements for header

## Api

- [x] Add `User { imageUrl }`
- [x] Add `User { chatUrl }`
- [x] Add `User { roles }`
- [x] Add `User { events }`
- [x] add `User { motivation }`
- [x] Add statistic data for a user

## Database layer

- [x] Query for user roles
- [x] Query for events
- [x] Query for user activity 

## Infrastructure

- [x] Add changes in module-api
- [x] Add changes in env-staging
- [ ] Add changes in env-production

## Planning needed / Other
- [x] How shall we transition from the old description?
- [x] Analyze: Who has an image in its description / in the community chat? How to retrieve it?
- [x] How shall we retrieve the statistical data?
- [x] Check: Is the element "motivation" well designed?
- [x] How shall roles be displayed?

# Follow Ups:

* Upload of profile picture in Serlo.
* Add diagram for subjects: 
![2021-03-08-183645_812x391_scrot](https://user-images.githubusercontent.com/1327215/110359593-f82d2280-803d-11eb-8c6a-6d7076fea4f2.png)
* Redirect for Rocket.Chat for not logged in user possible?

PR #1046, opened 2021-05-25
user-profile: Add modal "how to edit profile picture" 

kulla:
See https://github.com/serlo/frontend/issues/935

Edit-Button: 
2021-05-25-171317_288x384_scrot

Modal: 
2021-05-25-171325_890x711_scrot

@elbotho: Ready for styling feedback. The `focus` style for the button I do not like.... And what do you think about the modal? If it is okay, I would create a common modal object with title and close button.


elbotho:
> And what do you think about the modal? If it is okay, I would create a common modal object with title and close button.

Yeah makes sense. We actually use the `ShareModal` in two other places already.
Maybe also think about at the dynamic import used e.g. here: 
https://github.com/serlo/frontend/blob/d55928704975b3c74696935b200ac13dc37f256c/src/components/content/entity.tsx#L29

src/components/pages/user/profile.tsx
```diff
@@ -29,6 +37,7 @@ export const Profile: NextPage<ProfileProps> = ({ userData }) => {
   const isOwnProfile = auth.current?.username === username
   const lastLoginDate = lastLogin ? new Date(lastLogin) : undefined
   const registerDate = new Date(date)
+  const [howToEditImage, setHowToEditImage] = React.useState(false)
```
elbotho: I'd prefer something like `showImageModal` because it's a bit more descriptive.

src/components/pages/user/profile.tsx
```diff
@@ -117,12 +127,86 @@ export const Profile: NextPage<ProfileProps> = ({ userData }) => {
...
+    outline: 2px dotted ${(props) => props.theme.colors.brand};
+  }
+`
+
+const ProfileImageCage = styled.figure`
+  width: 150px;
+  height: 150px;
+  contain: content;
+
+  & > * {
```
elbotho: Small thing: We try to style children directly where possible.

elbotho:
Thank you! Looks great. I changed the focus style to match the hover style, looks and works well enough. Usually we depend on the browser default but for some reason it does not show in Firefox in the case (webkit/chrome works fine). Not sure why.

kulla:
@elbotho Suggestions implementing. Regarding the `StyledModal` I would handle it in a separate PR (follow up issue in https://github.com/serlo/frontend/issues/1049 ). This PR is IMHO ready to be merged.

elbotho:
is live on staging:
https://de.serlo-staging.dev/user/26217/kulla

PR #1054, opened 2021-05-28
fixes #1044

### LINKED ISSUE #1044: Authorization: Login can't create new content
In Taxonomy users should be able to create a new content (like an article).
Not sure if we have to solve this here or in the API.

elbotho:
@Entkenntnis could you check that? First time using `authorization`.
(explanation in #1044)


Entkenntnis (approved):
Thanks.

src/components/user-tools/author-tools.tsx
```diff
@@ -309,7 +309,6 @@ export function AuthorTools({ tools, entityId, data }: AuthorToolsProps) {
   }
 
   function renderNewEntity() {
-    console.log('renderNewEntity')
```
elbotho: thanks!

PR #1063, opened 2021-05-31
Improvements for user profiles from user research

kulla:
See #935 This PR includes the following improvement for the user profiles from a little user research:

* The person first wanted to click "edit profile" in the user menu to update the profile image -> remove item "edit profile" from user menu (can be accessed from the user profile") and change item "Public profile" -> "Own profile".
* The person wondered why the direct chat button is active in her account -> change colors so that the button seem to be deactivated

Also:

* Add `shouldReturnFocusAfterClose={false}` to our modal so that the button which opens the modal does lose it's focus after the modal is closed.

### DIFF
--- a/src/components/modal-with-close-button.tsx
+++ b/src/components/modal-with-close-button.tsx
@@ -30,7 +30,11 @@ export function ModalWithCloseButton({
   const { strings } = useInstanceData()
 
   return (
-    <Modal isOpen={isOpen} onRequestClose={onCloseClick}>
+    <Modal
+      isOpen={isOpen}
+      onRequestClose={onCloseClick}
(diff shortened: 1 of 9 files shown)


elbotho (approved):
👍 thank you.
Optional: We could add a title to the direct message button that explains that we use the community chat for that function.

kulla:
> Optional: We could add a title to the direct message button that explains that we use the community chat for that function.

@elbotho You mean like `<ChatButton title="You can chat with yourself like in Slack or WhatsApp">...</ChatButton>`?

elbotho:
Ah, I thought of the default case. Something like: 
> "You can send %username% a message using our chat (community.serlo.org)

PR #1064, opened 2021-05-31
Feature: Activate Revision History and Revision View

elbotho:
Now that the mutations are live (thank you @kulla and @inyono for the super fast implementation) [Revision History](https://frontend-git-feature-revisions-serlo.vercel.app/entity/repository/history/1565) and [Revision Preview](https://frontend-git-feature-revisions-serlo.vercel.app/entity/repository/compare/1565/162707) are ready for testing imo.

@Entkenntnis what do you think?

Notes:
- I checked some different content types, but not very thoroughly. 
- The Revision View always compares to the currently accepted revision now.

### DIFF
--- a/src/components/content/link.tsx
+++ b/src/components/content/link.tsx
@@ -33,8 +33,8 @@ const legacyLinks = [
 export function isLegacyLink(_href: string) {
   // compat: this is a special frontend route or force frontend use
   if (_href == '/user/notifications') return false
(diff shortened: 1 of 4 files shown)


Entkenntnis:
@elbotho Generally, I would like to include revision history and revision view into the current release.

I would wait this week to see if there are bigger issue with the current feature set and then add these features.

Entkenntnis:
Let's get this into our beta testing

PR #1076, opened 2021-06-03
More tailwind conversion

elbotho:


tailwind.config.js
```diff
@@ -109,6 +109,18 @@ module.exports = {
...
+          lineHeight: '1.22',
+        },
+        '.special-text-h4': {
+          fontSize: '1.1875rem',
+          lineHeight: '1.22',
+        },
+        '.special-text-2x': {
+          fontSize: '2rem',
+          lineHeight: '1.22',
+        },
```
Entkenntnis: see https://tailwindcss.com/docs/font-size#customizing

src/schema/article-renderer.tsx
```diff
@@ -226,8 +209,14 @@ function renderElement(props: RenderElementProps): React.ReactNode {
     return <StyledP>{children}</StyledP>
   }
   if (element.type === 'h') {
-    const Comp = StyledHx[element.level]
-    return <Comp id={element.id}>{children}</Comp>
+    return React.createElement(
+      `h${element.level}`,
+      {
+        className: `serlo-h${element.level}`,
```
Entkenntnis: This is a bit problematic together with purging: https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html

Adding an object like `{1: 'serlo-h1', 2: 'serlo-h2', ...}` should do the trick.

Entkenntnis:
Thank you @elbotho Overall, it already looks very good. just 2 small comments.

tailwind.config.js
elbotho: much to learn I have

elbotho: Should be solved!

Entkenntnis:
If you have no pending changes, you can merge https://github.com/serlo/frontend/pull/1077 into this branch

src/components/tags/styled-td.tsx
```diff
@@ -1,7 +1,5 @@
-import styled from 'styled-components'
+import { PropsWithChildren, FC } from 'react'
 
-export const StyledTd = styled.td`
-  padding: 10px;
-  vertical-align: top;
-  border: 3px solid ${(props) => props.theme.colors.lightBlueBackground};
-`
+export const StyledTd: FC<PropsWithChildren<{ className?: string }>> = (
```
elbotho: @Entkenntnis Experiment. Basically just shortens `<td className="serlo-td">` to `<StyledTd>`

tailwind.config.js
```diff
@@ -110,6 +110,9 @@ module.exports = {
         '.special-increment-list-counter': {
           'counter-increment': 'list-counter',
         },
+        '.special-content-space': {
```
elbotho, line 119: @Entkenntnis any Idea why this is not working?

src/components/tags/styled-td.tsx
Entkenntnis: Styled Components is doing a lot of hidden magic to make this approach work and make the component behave like a regular dom element, e.g. ref forwarding, display name, props forwarding, ...

While this is doable, I do not yet know if this is worth the effort of building such a wrapper.

tailwind.config.js
Entkenntnis: `content: '" "'` is doing the trick (and restart dev server for me)

src/components/tags/styled-td.tsx
elbotho: Okay, removed for now.

tailwind.config.js
elbotho: Thank you!

tailwind.config.js
```diff
@@ -69,17 +75,25 @@ module.exports = {
...
     },
     screens: {
       mobile: '500px',
       sm: '800px',
       md: '1024px',
       lg: '1216px',
+      'max-mobile': { max: '499px' },
+      'max-sm': { max: '799px' },
+      'max-md': { max: '1023px' },
+      'max-lg': { max: '1215px' },
```
Entkenntnis: See comment above, shouldn't be necessary.

src/components/pages/donations.tsx
```diff
@@ -98,18 +96,26 @@ export function Donations() {
...
               <FontAwesomeIcon icon={faPaypal} size="1x" /> Über PayPal spenden
-            </Button>
+            </a>
           </p>
         </DonationCol>
-        <ContactCol>
+        <div
+          className={clsx(
+            'max-md:pt-16 max-md:mt-11 max-md:border-t-8 max-md:border-brand-150',
+            'md:max-w-xs md:absolute md:right-0 md:top-48 md:pr-4'
```
Entkenntnis: Generally, tailwind is using the mobile first approach, e.g. no variant = mobile, with variant = desktop, so I would suggest this equivalent form:

grafik

Entkenntnis (approved):
Thank you very much for this work! I have changed one small place (max responsive variants) to make it more consistent.

src/components/pages/donations.tsx
```diff
@@ -154,136 +164,39 @@ export function Donations() {
...
-  margin-right: auto;
-  margin-left: auto;
-
-  @media (min-width: ${(props) =>
-      props.theme.breakpoints.md}) AND (max-width: 1320px) {
-    margin-left: inherit;
-  }
-
   iframe {
     margin-left: -20px;
```
Entkenntnis: @elbotho This line is causing the donation form to be clipped:

grafik

So I have removed it. Any thoughts?

src/components/pages/donations.tsx
elbotho: please search `@media (max-width` in our code base. If you are confident we can (and should ^^) replace all of those: fine. But that patterns sometimes saves a lot of code/complexity imo.
https://tailwindcss.com/docs/breakpoints#max-width-breakpoints

src/components/pages/donations.tsx
elbotho: Thanks, better this way.

src/components/pages/donations.tsx
Entkenntnis: I have actually looked through all cases ^^

In most cases, either there are just a few attributes affected or we are already setting the same attribut for other screen sizes.

The worst case is the meta menu, which is behaving as two completely different components -> looking at the code, it may be better to just split it into two different components and change the display property. (but this is for another day)

PR #1085, opened 2021-06-06
Community wall

Entkenntnis:
Integrating new community wall into landing page.

Early draft: https://frontend-git-community-wall-serlo.vercel.app/


elbotho:
@Entkenntnis first draft done. Looks pretty good imo.
Not tested in other browsers etc.

It's a mess of Tailwind, CSS and Styled Components now. 
But for this level on complexity (`:before` pseudo on hover animations and `vw` units inside of media queries) we have to rely a lot on `serlo-tailwind.css`. Maybe we can discuss sometimes how we want to handle this when the styles are only important for a single page / component.

Entkenntnis:
@elbotho Thank you! Wow, hover effects are looking nice :+1: 

Sorry for the inconvenience here with tailwind - next.js has built-in support for css-modules (https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css) and I think this could be good solution here ... but I can take care of this from here.

Entkenntnis:
@elbotho Are there any downsides of simply using an img instead of `:before`? In combination with tailwind's group-feature, I'm getting very similiar results (pr is wip atm, still playing around)

elbotho:
yeah, it's semantically wrong. 
it's decoration and therefore should be css. 

>Where possible, decorative images should be provided using CSS background images instead.

https://www.w3.org/WAI/tutorials/images/decorative/ (also for more infos on how to declare img that are not informative)

elbotho:
> Sorry for the inconvenience here with tailwind - next.js has built-in support for css-modules (https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css) and I think this could be good solution here 

Obviously not your fault. 
Yes maybe css components are a solution here. They also support sass which is a plus.
I could def. also imagine sticking with StyledC for the tricky stuff. For me the DX is pretty amazing.

Entkenntnis:
So, two decorations are implemented as backgroud images using tw - the rest will probably follow tomorrow

Entkenntnis:
@elbotho I have converted everything to tailwind and it looks fine to me. maybe you can check it again.

(found a super interesting discussion here: https://github.com/tailwindlabs/tailwindcss/issues/603)

src/components/pages/landing-de.tsx
```diff
@@ -1,203 +1,250 @@
...
+          className={clsx(
+            'text-center text-4xl text-truegray-700 font-bold',
+            'leading-cozy tracking-tight',
+            'max-w-2xl mt-40 mx-auto relative z-10'
+          )}
+        >
+          Wir sind eine große, ehrenamtliche Community und gestalten Serlo
+          <p className="text-brand italic font-handwritten">gemeinsam.</p>
+        </h3>
+        <div className="relative z-0 h-0 w-full mt-1">
```
elbotho, line 135: Tbh: not a fan of html markup for a decorative element. For me that's hacky.

src/components/pages/landing-de.tsx
```diff
@@ -1,203 +1,250 @@
...
+      <section>
+        <h3
+          className={clsx(
+            'text-center text-4xl text-truegray-700 font-bold',
+            'leading-cozy tracking-tight',
+            'max-w-2xl mt-40 mx-auto relative z-10'
+          )}
+        >
+          Wir sind eine große, ehrenamtliche Community und gestalten Serlo
+          <p className="text-brand italic font-handwritten">gemeinsam.</p>
```
elbotho: `gemeinsam` is too small. 
I suggest `font size: 2.7rem;` and maybe `margin-top: -4rem;´ on the background to match the design more closely.

src/helper/shuffle-array.ts
```diff
@@ -0,0 +1,7 @@
+export function shuffleArray<T>(array: T[]) {
```
elbotho, line 1: Nice, that's handy.

elbotho (changes requested):
Nice result 👍 
Requested small change.
And sorry, but as you can tell I remain sceptical of the 100% tailwind approach.

src/components/pages/landing-de.tsx
Entkenntnis: I have increased the font-size and it looks fine (for reference: the design is using 2.5rem / 3rem)

src/components/pages/landing-de.tsx
Entkenntnis: I can understand your idea behind this statement, but the "evidence" points in another direction:

Counter example: https://big.tc/ is using simple plain html elements for the arrows (and is not even setting them as background - just straight img tags).

According to https://www.w3.org/WAI/tutorials/images/decorative/ I'm already using css background for this.

A google search wasn't that supportive of that claim: https://www.google.de/search?q=don't+use+html+for+decorative+elements

So I think, by using css background we are already a bit better than the "default" and it feels ok to me to build it that way. At least it's simple to understand whats happening.

src/components/pages/landing-de.tsx
elbotho: Looks good 👍

src/helper/shuffle-array.ts
Entkenntnis: :)

elbotho:
this is now part of #1121

PR #1106, opened 2021-06-09
Add eslint rule "no-console"

kulla:
Thus we can easier see whether there are leftover console.log() statements (see https://eslint.org/docs/rules/no-console ). Please check whether the remained console.log() and console.warn() statements should be in the code.


elbotho (approved):
lgtm!

PS: you sure like the word thus :)

Entkenntnis:
thank you!

PR #1115, opened 2021-06-11
Update texts for rejecting revisions

kulla:
"not accepting" is not so harsh as "rejecting". See also https://community.serlo.org/channel/software-features-and-bugs?msg=PGDEtgM4ZbNYcwv4q


Entkenntnis:
thank you

PR #1121, opened 2021-06-14
Landing page redesign

elbotho:
So, this is the draft for the new landing page.
based on (#1085)

### Status

- This is a pretty complete hack.
- The quiz feature is not quite ready conceptually and is therefore missing.
- The code is a mess of Tailwind and StyledC and duplication. But it works and imo is okay for now.

### Preview
https://frontend-git-landing-page-serlo.vercel.app

### TODO (not complete)

- [x] don't load new font from google
- [x] use bigger font weights as in the drafts
- [x] test in anything other than chrome dektop
- [x] 1-2 Icons differ from the draft (but I think that's okay for now)
- [x] Split up huge file
- [x] Test on mobile


src/components/landing/rework/landing-subjects-new.tsx
```diff
@@ -0,0 +1,317 @@
...
+import SustainabilitySVG from '@/assets-webkit/img/landing/subjects-sustainability.svg'
+import { Link } from '@/components/content/link'
+import { LandingSubjectLink, LandingSubjectsData } from '@/data-types'
+import { makeResponsivePadding, makeTransparentButton } from '@/helper/css'
+
+interface LandingSubjectsProps {
+  data: LandingSubjectsData
+}
+
+export function LandingSubjectsNew({ data }: LandingSubjectsProps) {
```
elbotho, line 26: This is almost identical to the current one (we still use on e.g. es.serlo.org). 
Probably one of the first files where we can get rid of the code duplication if we move forward with this.

src/components/landing/rework/menu-new.tsx
```diff
@@ -0,0 +1,296 @@
...
+  user: faUser,
+  notifications: faBell,
+}
+
+export interface MenuProps {
+  data: HeaderData
+  auth: AuthenticationPayload
+}
+
+export function MenuNew(props: MenuProps) {
```
elbotho, line 36: Also very close to the current code.

elbotho: Current code would enjoy some refactoring :)

src/components/landing/rework/footer-new.tsx
```diff
@@ -15,39 +17,45 @@ export function FooterNew() {
...
   )
 
   function renderFooterLine() {
+    const items = R.concat(
+      R.map(
+        R.assoc('mdHidden', false),
+        footerData.footerNavigation[2].children
+      ),
+      R.map(R.assoc('mdHidden', true), footerData.footerNavigation[1].children)
+    )
```
elbotho: Hmm to be honest this is quite unreadable for me since I never really used ramda.
Atm. we only use ramda twice in the frontend. 
We should make a conscious decision if we want to rely on it more. 

My personal preference would be to use it sparingly, since it's another library new people might have to learn when they want to contribute.
 And we managed to be pretty functional (as in functional programming) without it.

elbotho: Not saying that the previous iteration was very elegant ^^
We could def. merge the two maps or move them to a function to avoid the duplication.

Entkenntnis: As soon as you understand that ramda supports currying and partial application, this code actually makes sense.

(side note below)

The basic idea is good, why not define the function as

```jsx
function renderFooterLineItems(
  items: {title: string; url: string}[]
  mdHidden: boolean 
) {
 return items.map(...)
}
```
and call it like

```jsx
<>
  {renderFooterLineItems(footerData.footerNavigation[2].children, false)}
  {renderFooterLineItems(footerData.footerNavigation[1].children, true)}
</>
```

src/components/landing/rework/landing-subjects-new.tsx
```diff
@@ -0,0 +1,312 @@
...
+  return (
+    <>
+      <SubjectsWrapper>{data.subjects.map(renderSubject)}</SubjectsWrapper>
+
+      {data.additionalLinks.length > 0 && (
```
kulla: @elbotho Are there any additional links in the new landing page design?

elbotho: Not at this point, but I opted to keep the code since the design is not fixed atm.
If you want you can mark it with a comment to remove it when we remove the "old" landing page completely.

elbotho: Or remove it now.

src/components/pages/landing-de.tsx
```diff
@@ -1,203 +1,225 @@
...
+          className={clsx(
+            'hidden md:inline-block mx-auto mt-12',
+            'font-bold text-xl rounded-lg text-truegray-700 ',
+            'px-8 py-4 tracking-tight border-truegray-700 border-solid border-2',
+            'hover:border-brand-light hover:no-underline hover:text-brand-light'
+          )}
+          href="/partner"
+        >
+          Alle Partner ansehen
+        </ButtonWithWink>
```
kulla: This button is not visible in mobile nor desktop version due to the class "hidden". Can it be deleted?

elbotho: `hidden` uses `display: none` which is then overwritten by `md:inline-block`.

kulla (approved):
Okay, I have finished my refactoring. There are some follow ups for long rainy days (see https://github.com/serlo/frontend/issues/1143 ) but nothing importing. Can be IMHO merged.

elbotho:
> Okay, I have finished my refactoring. There are some follow ups for long rainy days (see #1143 ) but nothing importing. Can be IMHO merged.

Thank you very much @kulla.

PR #1127, opened 2021-06-17
next 11, next lint, updates

elbotho:
Testing next 11 and `next lint`.

It seems we need the new static image feature disabled to make it work with webpack5 for now.
 `disableStaticImages: true`


tsconfig.json
```diff
@@ -21,5 +21,5 @@
     }
   },
   "exclude": ["node_modules"],
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"]
+  "include": ["custom.d.ts", "next-env.d.ts", "**/*.ts", "**/*.tsx"]
```
elbotho, line 24: needed to svgr to work.

.eslintrc-next.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable import/no-commonjs */
+module.exports = {
+  extends: ['next'],
```
inyono: Seems like this is not the recommended way, see https://nextjs.org/docs/basic-features/eslint#migrating-existing-config.

elbotho: Thanks!

PR #1128, opened 2021-06-18
Auth

### LINKED ISSUE #1020: User menu is hidden if logged in from another tab
Currently If I log in in one tab, the other tab just hides the user menu completely. 
Reload fixes it, but navigation not. Logout actually works better.
image
Maybe we could solve that particular issue instead, but atm. it's … strange :)

_Originally posted by @elbotho in https://github.com/serlo/frontend/issues/851#issuecomment-839707196_

### LINKED ISSUE #1090: Log out after 2h
If logged in in the frontend, you will get logged out after 2h.

Needs investigation.

### LINKED ISSUE #1100: Not able to relogin as another user
Sometimes it happens that after logout, clicking login is automatically using the last user. This makes it impossible to login as another account.

### LINKED ISSUE #1108: Make fetching authorization more robust
Currently, there is following bug:

If the token is stale, there is a auth-refresh, and notifications are refetched, but authorization is not (it actually should) - this could help with #1090.

inyono:
Fixes #1020
Fixes #1090
Fixes #1108 
Fixes #1100


src/api/endpoint.ts
```diff
@@ -1,3 +1,6 @@
 import { serloDomain } from '@/helper/serlo-domain'
 
-export const endpoint = `https://api.${serloDomain}/graphql`
+export const endpoint =
+  process.env.NEXT_PUBLIC_ENV === 'local'
+    ? 'http://localhost:3001/graphql'
```
inyono, line 5: We should permanently change up the port of API or frontend so that they don't overlap. Ignoring that for now.

scripts/tailwind-status.js
```diff
@@ -133,7 +133,7 @@ if (fs.existsSync('./.next/BUILD_ID')) {
   }
 
   if (unknownClasses.some((cls) => !knownExceptions.includes(cls))) {
-    throw 'Unknown class name!'
+    throw new Error('Unknown class name!')
```
inyono, line 136: Was throwing a string intended?

Entkenntnis: This is ok, more a sloppyness from my side

src/components/frontend-client-base.tsx
```diff
@@ -85,37 +80,20 @@ export function FrontendClientBase({
...
-          query: 'query{authorization\nnotifications(unread:true){totalCount}}',
-        })
-      )
-        .then((value: { authorization: AuthorizationPayload }) => {
-          setAuthorizationPayload(value.authorization)
-        })
-        .catch(() => {})
-    }
-  }, [loggedInComponents, auth])
-
```
Entkenntnis, line 109: @inyono see slack for explaination

inyono:
@Entkenntnis I addressed the two issues discussed on Slack. See https://github.com/serlo/frontend/pull/1128/commits/1d79a57964742e3924dc95de6de9815c06316216 and https://github.com/serlo/frontend/pull/1128/commits/bfac3b6015a52159bec5a9f1d325c36a3f6d1444. Bundle size doesn't increase anymore compared to staging.

PR #1133, opened 2021-06-26
Revision compare improvements

elbotho:
This PR implements most of the changes in #1132 so we can test and refine them.
ping @yolo-bird: feedback appreciated.

[closes #1132]

PR #1140, opened 2021-06-28
add rel tags to profile links to discourage spam

elbotho:
Sorry, I thought we have this for profiles already, but it was only on comments. 
This should in theory discourage spammers since there is no more SEO advantage to be gained. (But since they also use the chat now… idk 😩)

@kulla or @Entkenntnis could one of you review this? 
Unfortunately we can't really test it in staging since the descriptions are missing.

### DIFF
--- a/src/schema/article-renderer.tsx
+++ b/src/schema/article-renderer.tsx
@@ -171,8 +171,10 @@ function renderElement(props: RenderElementProps): React.ReactNode {
   const { element, children, path } = props
 
   if (element.type === 'a') {
+    const isOnProfile =
+      path && typeof path[0] === 'string' && path[0].startsWith('profile')
     return (
-      <Link href={element.href} path={path}>
+      <Link href={element.href} path={path} unreviewed={isOnProfile}>
         {children}
       </Link>
     )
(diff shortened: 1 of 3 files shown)


Entkenntnis:
Thank you, looks good for me.

Entkenntnis (approved):

src/schema/article-renderer.tsx
```diff
@@ -171,8 +171,10 @@ function renderElement(props: RenderElementProps): React.ReactNode {
   const { element, children, path } = props
 
   if (element.type === 'a') {
+    const isOnProfile =
+      path && typeof path[0] === 'string' && path[0].startsWith('profile')
```
kulla, line 175: Out of curiosity: Where is path set?

kulla (approved):
LGTM and works :-)

kulla:
> Unfortunately we can't really test it in staging since the descriptions are missing.

I tested it locally, it works

src/schema/article-renderer.tsx
elbotho: https://github.com/serlo/frontend/blob/staging/src/components/pages/user/profile.tsx#L90
It's a parameter of `renderArticle`

elbotho:
> I tested it locally, it works

How? :)

src/schema/article-renderer.tsx
elbotho: quite handy feature, used it a lot in #1133

elbotho:
Thank you!

PR #1168, opened 2021-07-07
check permissions before showing edit button

### LINKED ISSUE #1154: check permissions for edit button
Permissions are not checked for the edit button yet, e.g. createEntityRevision / editTaxonomy / pageRevision .... This way, you could easily get a "Zugriff verboten" warning.

Entkenntnis:
fixes #1154


elbotho:
Thanks for fixing this 🎉 

`AuthorToolsData[type]` should probably not be typed as `string` but concrete values (might simplify the `else ifs`). But that's unrelated.

PR #1182, opened 2021-07-14
More improvements for revision view

elbotho:
- adds revision help link to tool menu
- adds note for grouped exercises (what grouped ex does this revision belong to?)

unrelated
- refactor entity-base (move comments from taxonomy and entity into base)
- min height for content (looks better on very short content)


Entkenntnis (approved):
looks good so far.

PR #1184, opened 2021-07-15
graphql config v3

elbotho:
@Entkenntnis you also use VSCode right?
I updated and renamed the config file (now using [`graphql-config v3`](https://graphql-config.com/migration)) and it seems to work better with `graphql.vscode-graphql` now.

Merge if it works for you too.


Entkenntnis:
Thank you for the hint, wasn't using the plugin yet, but good to know what works best here.

If you like to, you can add the plugin to our wiki https://github.com/serlo/frontend/wiki/VS-Code.

PR #1197, opened 2021-07-25
Deployment

Entkenntnis:
Events + Profile - Updates / Revision View Updates


elbotho (changes requested):
Please don't merge yet. There are pending API bugs and at least one feature that should not be activated yet.

elbotho (approved):

PR #1200, opened 2021-07-26
Printmode & PDF Export feature

### LINKED ISSUE #1213: Feature: offer pdf download

elbotho:
### Print modes
- `print--preview` [test](https://frontend-git-pdf-print-test-serlo.vercel.app/1565#print--preview)
- `print--now` [open this](https://frontend-git-pdf-print-test-serlo.vercel.app/1565) and use browsers print dialog to test (you should be redirected in chrome and firefox and all components/images should display)

Print preview may become the base for a improvised pdf export

### Example
https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?url=https%3A%2F%2Ffrontend-git-pdf-print-test-serlo.vercel.app%2F1565 or in the tool menu for logged in users (locally)

### Todo
- [x] Find a solution for direct PDF export? ([puppeteer docker](https://blog.risingstack.com/pdf-from-html-node-js-puppeteer/) or pimp [html2pdf](https://github.com/eKoopmans/html2pdf.js))
- [x] Add time and Link to original content
- [x] format date string for locale
- [x] Automatically extend spoilers, exercises and solutions
- [x] Add to menu
- [x] Remove Puppeteer before merge


Entkenntnis:
Thank you for testing this out! Print-preview already looks very nice!

Pretty amazing that you can run a headless browser on vercel serverless :grinning: But the test link above is not picking up the print preview page :thinking: and something like `https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?url=https%3A%2F%2Ffrontend-git-pdf-print-test-serlo.vercel.app%2F1565` is also not working ...

Another "problem": puppeteer is just huge. It doubles my node_modules folder size (from ~360mb to ~800mb) - maybe we can also use the compressed chromium version locally?

elbotho:
Thank you for the review!

> headless browser on vercel serverless

Jup, pretty happy with that :)  Took quite a bit of research.

~Test links don't work, because I restricted it to domain `serlo.org` (maybe to restrictive).
serlo in the domain somewhere is probalby enough. (but it should work once merged)~
-> Should work now in [preview](https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?url=https%3A%2F%2Ffrontend-git-pdf-print-test-serlo.vercel.app%2F1565) and locally.

Good point with the node_modules size. The compressed version is optimized for AWS (some linux version) so using that locally won't work. But we could either: use a locally installed chome (should acutally work ^^) or just remove puppeteer locally and only add it to branches if we actually need to work on the pdf export (which is hopefully almost never).

elbotho:
@Entkenntnis ah yes also: Do you have an idea how we could avoid the css duplication? 
It's suprisingly hard to share code between a print media query and a class. 
Had no luck with `postcss-extend`.

src/assets-webkit/styles/serlo-tailwind.css
```diff
@@ -255,26 +250,57 @@
...
+    img,
+    table {
+      page-break-inside: avoid;
+    }
+    .serlo-exercise-wrapper {
+      border-bottom: 1px solid black;
+    }
   }
   .header-breadcrumbs,
   a,
```
Entkenntnis: global styles are a bit tricky generally - you might try to add the class `print:serlo-print-style ` to the body tag (print is the media selector we have activated in the config) - this should avoid the duplication.

elbotho: ah nice, I'll try that.

.eslintrc.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
     // eslint
     'no-duplicate-imports': 'error',
     'no-unused-vars': 'off',
-    'no-console': 'error',
+    'no-console': 'warn',
```
elbotho, line 24: personally I think the warnings are promiment enough :) and this sometimes annoyed me while developing.

Entkenntnis: agree, this is fine :+1:

src/pages/api/pdf.ts
```diff
@@ -0,0 +1,101 @@
...
+    padding-left:40px;
+    color: #666;
+    `
+
+export default async function createPdf(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const noSolutions = req.query.noSolutions !== undefined
+  const urlString = decodeURIComponent(req.query.url.toString())
```
Entkenntnis: Passing the full url to the route feels a bit awkward. Can we refactor this and only pass in the ID of the content?

You can get the absolute url of the deployment with a package like https://github.com/jakeburden/next-absolute-url

(for the language, we can probably also pass it in as query parameter ...)

Entkenntnis (changes requested):

src/pages/api/pdf.ts
elbotho: We could do it this way. But tbh. I don't share your awkward feeling :)
We use the same style for fetching thumbnails:
https://embed.serlo.org/thumbnail?url=https%3A%2F%2Fwww.youtube-nocookie.com%2Fembed%2FKtV2wlp9Ts4%3Fautoplay%3D1%26html5%3D1
Also all the share links work the same way.

But I can be convinced :) 
Why do you prefer to use ids?

elbotho: okay, it now uses an id parameter and the absolute id.

elbotho: (also: it's not working any more -_-)
please don't merge

Entkenntnis: This is working: https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?id=1565

elbotho: super weird. 
this is not: https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?id=1555
but it seems to be related to this page.
… okay, need to test more entities again. so weird because I'm sure this one worked without problems before.
(and works locally)

Entkenntnis: But revisions are not working: https://frontend-git-pdf-print-test-serlo.vercel.app/api/pdf?id=212696

elbotho: thats fine. we don't offer that.

elbotho: still not sure what happens on 1555…

elbotho: okay, just to document here:
1555 results in a >22MB PDF locally. Which is too big for AWS (and too big in general).
Probably something with the images, maybe they convert it from webp to something else and don't mind the size?

elbotho:
```
[ERROR] [1628665754926] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 413.
[ERROR] [1628665754951] LAMBDA_RUNTIME Failed to get next invocation. Http Response code: 403
Unexpected Top Level Error: Error: Failed to get next invocation, error 403
[ERROR] [1628665754963] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 403.
RequestId: 3a2fb12c-2a39-4d24-a2f5-493e5ba16099 Error: Runtime exited with error: exit status 129
Runtime.ExitError
```

Entkenntnis: interesting! The pdf-size of 1565 is <1MB and looks fine ... and the images for 1555 aren't that big, either.

elbotho: jup, im researching that now. there are bugs relating to images in the pdf creation, but they should be solved by now…

elbotho:
okay, pretty annoying that I just found this problem so late in the progress. but it seems I might have found a solution.

### summary:
- aws has file size limits 
- pages with some images result in pretty large pdfs
- bad times -> no return from the serverless function
- I did not find any options to set image quality or convert them so far
- (hack with page size and scaling did some good, but not enough so … nevermind)

### why?
so yeah, it's basically worst case.
- our images can be huge because we don't reduce the image dimensions
- we get nice `webp`s from cloudflare to decrease file size
- Unfortunately chromium converts those `webp`s to pngs and so we end up with high resolution unoptimized pngs in our pdf. yay 🥇 

### solution
I overwrite the default `accept` header to not include `image/webp`. Now the still to large but somewhat optimized jpgs get loaded.
22.3MB PDF is now 4MB

### future improvements:
- shrink images dimentions on upload to fit limits
- add width/height info for images in content (via edtr.io) so we can use `next/image` and let vercel provide optimized (and smaller) images for us

Entkenntnis (approved):

Entkenntnis:
@elbotho Thank you for investigating this and working on the pdf export.

These finding are helpful, also for the general usage, e.g. image optimizations, hopefully we can find a good solution for this in the future.

Feel free to merge this into staging when you are ready with it :+1:

PR #1201, opened 2021-07-26
add custom style for mini components

elbotho:
@Entkenntnis what do you think of this solution for styles that repeat but only inside one component?

### DIFF
--- a/src/components/pages/revision-history.tsx
+++ b/src/components/pages/revision-history.tsx
@@ -6,7 +6,11 @@ import { UserLink } from '../user/user-link'
 import { Link } from '@/components/content/link'
 import { TimeAgo } from '@/components/time-ago'
 import { useInstanceData } from '@/contexts/instance-context'
-import type { CompBaseProps, HistoryRevisionsData } from '@/data-types'
+import type {
+  CompBaseProps,
+  HistoryRevisionData,
+  HistoryRevisionsData,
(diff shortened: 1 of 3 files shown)


Entkenntnis:
basically we have two approaches:

1. renderXYZ(arguments, ...)
2. Mini-Component

I have preferred (1.), because it's straight forward for simple cases. (2.) is fine for me, too, if it's only used within one component. In general, a wrapper is quite complex to write, styled components has put a lot of "black magic" into it, so if (1.) is working, I'd probably go with that (like in this case).

src/components/pages/revision-history.tsx
```diff
@@ -98,17 +97,13 @@ export function RevisionHistory({ data }: RevisionHistoryProps) {
   }
 }
 
-const Th = ({ text }: { text: string }) => (
-  <th className="serlo-th sticky top-0 bg-white border-0">{text}</th>
+const Th: CompBaseProps = ({ children }) => (
+  <th className="serlo-th sticky top-0 bg-white border-0">{children}</th>
 )
```
Entkenntnis: consistent with other places:

```
function renderTh(text: string) {
  return <th className="serlo-th sticky top-0 bg-white border-0">{text}</th>
}
```

Entkenntnis (changes requested):

PR #1202, opened 2021-07-26
Include sustainabilty in en-navigation

jakobwes:
The subject is now big enough to be featured there.


elbotho:
@jakobwes thank you, will probably go online next week.

PR #1210, opened 2021-07-28
Author menus: also show unrevised instead of edit

elbotho:
For submenus (courses and exercises/exgroups/solutions) editing on top of existing unrevised revision could happen easily.

@Entkenntnis I hope you don't mind, I renamed `lic` to `loggedInComponents` across the project, to make it easier to understand for newcomers (and me). Also I renamed Comp to the Components Name when it was not right below the definition.


Entkenntnis:
@elbotho Thank you for renaming! The naming was indeed confusing.

Entkenntnis (approved):

PR #1212, opened 2021-07-30
prototype edtr.io in comments

### LINKED ISSUE #1175: More editing options for comments
Mandatory options:
- latex-formulas
- links
- bold, italic

Nice to haves
- colours
- lists
- ping other authors (more complex and would need more planning)
- Emojis

Btw. can we easily transfer the editing options for the text plug-in of edtr.io to comments?

elbotho:
### example
https://frontend-git-rich-comments-serlo.vercel.app/193389

### includes:
- working prototype
- some css-hacks to change the style of edtr.io children (might even be okay now that I just accepted the floating toolbar)

### TODO
- [x] resolve conflicts
- [ ] one type error I don't understand
- [ ] all strings in english, we could experiment with react-i18n-next integration
- [x] ~conversion should happen on server side -> extract fetch login into separate api route~ converter is reasonable small to be loaded on the client
- [x] only load edtr io when form is activated? -> comment form is dynamically loaded only for logged in user
- [x] fix tests
- [ ] problem with math input if using `strg-m` to insert element


Entkenntnis:
Tested in locally right now - and it works :) ... but yes, there are some minor bugs, e.g. math formula rendering or missing translations.

Entkenntnis:
(even auto-detect of pasted urls is working)

elbotho:
> missing translations

Ah, good catch! Will have to find out how I can tell edtr what language to use.

elbotho:
could you give me details/an example for the math formula rendering problem?
Seems to work for me.

Entkenntnis:
@elbotho It was the bug when the css is missing and formulas are appearing twice - but I can't reproduce! Looks like a one-time quirk, I think we can ignore at atm ...

elbotho:
- [ ] check how we can load this on demand only

Entkenntnis:
the whole comment part of the application is already lazy loaded.

elbotho:
> the whole comment part of the application is already lazy loaded.

true. My thinking: Most people won't acutally comment so they don't need to load the edtr.

Entkenntnis:
Ah, right - we indeed need another level here

Entkenntnis:
No work planned on this PR for the next 3-4 months, will focus on #1329 instead. (keep branch)

PR #1218, opened 2021-08-06
add support for slate inline code

elbotho:
@Entkenntnis  do you remember why we did not support this before or did I overlook something? 😅 

### example
https://frontend-git-text-code-serlo.vercel.app/221693


Entkenntnis (approved):
Thanks! I can remember that we asked if this feature is needed and the response was "no", so we didn't implement it - but maybe I misremember or misunderstood. In any case, I think it's good to implement it now.

PR #1219, opened 2021-08-06
use new mutation to accept page revision

### LINKED ISSUE #1205: `/page/revision/revisions/{id}` not supported
`/page/revision/revisions/{id}` route is not supported.

The good new is that `/entity/repository/history/` handles pages alright so we should be able to use that instead.
(atm. the accept /reject mutation [does not support pages](https://github.com/serlo/api.serlo.org/issues/457))

elbotho:
fixes #1205

### DIFF
--- a/src/components/author/revision/checkout-reject-buttons.tsx
+++ b/src/components/author/revision/checkout-reject-buttons.tsx
@@ -13,6 +13,7 @@ export interface CheckoutRejectButtonsProps {
   repositoryAlias: string
   isRejected: boolean
   isCurrent: boolean
+  isPage: boolean
   buttonStyle?: string
 }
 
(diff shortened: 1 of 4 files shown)


src/components/author/revision/checkout-reject-buttons.tsx
```diff
@@ -58,7 +65,7 @@ export function CheckoutRejectButtons({
         <FontAwesomeIcon icon={faCheck} className="lg:mr-0.5" />{' '}
         {strings.revisions.checkout.action}
       </button>
-      {!isRejected && (
+      {!isRejected && !isPage && (
```
elbotho, line 68: pages are not part of the revision process and therefore don't need to be rejected

Entkenntnis (approved):
thank you for the fix

PR #1220, opened 2021-08-09
Unrevised Revisions

elbotho:
**Upsides**
- behaviour of the legacy page is almost copied.
- loads super fast (let's calculate that speed difference 🎉 )
- adds quicklinks and total numbers
- offers lighter overview with option to display all revisions
- adds labels for `wip` and `new`

**Downside**
- ~~atm we don't display edits from new users differently~~

### Demo
https://frontend-git-unrevised-revisions-serlo.vercel.app/entity/unrevised


Entkenntnis (approved):
Looks very good!

elbotho:
this is through the first feedback and can be merged imo.
(we need to work with @kulla to update the api before deploying to production)

PR #1226, opened 2021-08-11
new page `___content`

elbotho:
since the begining we successfully averted having storybook ^^
maybe now it's time for a first step.
based on https://github.com/serlo/frontend/wiki/Schema I created a minimal helper page at `___content` to access different entity types directly. 
This probably comes a bit late, but I think it still might be helpful.
more links can be added as needed.


Entkenntnis (approved):
Love the ultra minimal approach :+1:

PR #1257, opened 2021-08-20
schema: add new semantic article plugin

inyono:
Super hacky, probably need to talk about how this should be actually done. Still gonna merge so that we can proceed with the article plugin.

cc @elbotho @Entkenntnis just for the notification

### DIFF
--- a/src/schema/convert-edtr-io-state.tsx
+++ b/src/schema/convert-edtr-io-state.tsx
@@ -11,9 +11,11 @@ import {
 import { sanitizeLatex } from './sanitize-latex'
 import {
   FrontendContentNode,
+  FrontendLiNode,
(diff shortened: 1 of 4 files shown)


src/schema/convert-edtr-io-state.tsx
```diff
@@ -72,8 +200,8 @@ function convertPlugin(node: EdtrState) {
     return [
       {
         type: 'img',
-        src: node.state.src,
-        alt: node.state.alt,
+        src: node.state.src as string,
+        alt: node.state.alt || '',
```
inyono, line 204: Here was a mismatch between types and actual implementation. Feel free to revert that change and instead make `alt`. optional

PR #1270, opened 2021-09-10
equations: Use border-left for transform step

kulla:
Result: 
2021-09-10-133152_404x301_scrot

See at `/mathe/220515/220515`


Entkenntnis (approved):
Thank you

PR #1275, opened 2021-09-13
Article plugin refactor

elbotho:
Moves new code to a component.
This is how we do it for the other plugins right now.
Will need to be refactored when we integrate Edtr more closely.


inyono:
LGTM, will go over some minor details / i18n before merging.

elbotho:
just to make sure: german (an other langs) i18n is done in staging via crowdin.

PR #1296, opened 2021-09-24
add max-width to all content

elbotho:
@Entkenntnis this is bit crude, but seems effective. Any objections? :)

### DIFF
--- a/src/components/header-footer.tsx
+++ b/src/components/header-footer.tsx
@@ -11,7 +11,9 @@ export function HeaderFooter({ children }: HeaderFooterProps) {
   return (
     <>
       <Header />
-      <div className="min-h-[68vh] max-w-full overflow-x-hidden">
+      <div
+        className="min-h-[68vh] max-w-full overflow-x-hidden lg:overflow-visible" /* needed for sticky menu*/
+      >
         {children}
       </div>
       <Footer />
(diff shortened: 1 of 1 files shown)


Entkenntnis:
unfortunately, this breaks the sticky placement of the share-toolbox.

Entkenntnis:
maybe we can find another way to make the sticky placement possible & hide overflow

elbotho:
> unfortunately, this breaks the sticky placement of the share-toolbox.

😿 
nice catch.

src/components/header-footer.tsx
```diff
@@ -11,7 +11,9 @@ export function HeaderFooter({ children }: HeaderFooterProps) {
   return (
     <>
       <Header />
-      <div className="min-h-[68vh] max-w-full overflow-x-hidden">
+      <div
+        className="min-h-[68vh] max-w-full overflow-x-hidden lg:overflow-visible" /* needed for sticky menu*/
```
Entkenntnis, line 15: @elbotho THIS could be the final approach :)

Entkenntnis: only thing that breaks is the sticky header for e.g. https://de.serlo.org/entity/repository/compare/224015/224311 - but that would be fine.

elbotho: since it's working on larger screens I can live with that.

PR #1318, opened 2021-10-08
preview route for authors

Entkenntnis:
experiment with disabling cache for authors

checks for auth are stubs to allow testing


Entkenntnis:
alternative approach in #1319

PR #1323, opened 2021-10-08
increase revalidate to 15 minutes

Entkenntnis:
experiment to reduce server load


elbotho:
Is there a good way to quantify this without waiting 30 days? ;)

Entkenntnis:
There should be a day-to-day resource tab on vercel, so we should see a decrease if we compare weeks

PR #1326, opened 2021-10-09
simplify mouse events, named imports if possible

elbotho:
does not change bundle sizes, so tree shaking seems to work ;)


Entkenntnis (approved):
thank you for the cleanup

PR #1353, opened 2021-10-26
next 12, upgrades, …

elbotho:


Entkenntnis:
bundle size increased to 153kb  ... it looks like _app is always bundled but I haven't investigated it yet.

elbotho:
> bundle size increased to 153kb 

As always, thanks for keeping an eye on that.

elbotho:
this branch: `153kb`
without `swcMinify`: `148kb`
current main: `144kb`

Entkenntnis:
related: https://github.com/vercel/next.js/discussions/30237

elbotho:
with `12.02` and `swcMinify: false` we are at `147kb`
the `3kb` might just be new code I guess.
@Entkenntnis should we merge it or wait for further improvements?

Entkenntnis:
looks fine, we can investigate bundle size increase later

PR #1358, opened 2021-10-29
user: Same behavior of shoing description in staging und production

kulla:
After we export the description also in staging both production and
staging shall have the same behavior

* with description: https://frontend-git-description-staging-serlo.vercel.app/user/profile/kulla
* without description: https://frontend-git-description-staging-serlo.vercel.app/user/profile/peter_serlo1

Is it normal that https://frontend-git-description-staging-serlo.vercel.app/1 results in an error?

### DIFF
--- a/src/components/pages/user/profile.tsx
+++ b/src/components/pages/user/profile.tsx
@@ -130,7 +130,6 @@ export const Profile: NextPage<ProfileProps> = ({ userData }) => {
 
   function renderDescription() {
     if (!description) return null
-    console.log(description)
     return (
       <section>
         <h2 className="serlo-h2">{strings.profiles.aboutMe}</h2>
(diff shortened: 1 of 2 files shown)


elbotho (approved):
Thank you, looks good.

elbotho:
> Is it normal that https://frontend-git-description-staging-serlo.vercel.app/1 results in an error?

Well… it's expected. CF-Worker handles the redirect of user uuids for us on staging and production.
We could handle it in the frontend, but it's complexity and we usually don`t need it.

kulla:
> Well… it's expected. CF-Worker handles the redirect of user uuids for us on staging and production. We could handle it in the frontend, but it's complexity and we usually don`t need it.

Ah, okay :-) I just missed it...

PR #1359, opened 2021-10-30
fix: remove react-virtual

Knorrke:
~This PR adds a resizeobserver to react-virtual as shown in this issue: https://github.com/tannerlinsley/react-virtual/issues/167 to account for dynamic size changes of text-exercises.~

**EDIT**: Our implementation of react-virtual currently isn't working (All elements are rendered anyway) and causing some UI issues (Not listening to dimension changes by spoilers, misplacing the plugin overlay). Easiest solution for now is to just revert https://github.com/serlo/serlo.org/pull/656 which this PR does.

## Fixes
It fixes the issue reported by @kathongi that text-exercise groups couldn't be edited in some cases, because the solution where overflowing the next exercise. E. g. https://de.serlo.org/entity/repository/add-revision/36869

I hope performance is not impacted to heavily, by the recalculations during editing. Is there a huge text-exercise-group to test this on?

## Before
without-resizeObserver
## After
resizeObserver


Entkenntnis:
thank you @Knorrke for submitting a fix :)

tested it with your example (https://frontend-git-text-exercise-observe-resize-serlo.vercel.app/entity/repository/add-revision/36869) and works fine.

Is this one big enough? https://frontend-git-text-exercise-observe-resize-serlo.vercel.app/entity/repository/add-revision/2373

Knorrke:
thanks, performance looks fine for me.

Got a different issue though, with the plugin overlay `position: fixed`. The `transform` of react-virtual is messing with that ([transformed parents act as a containing block for fixed descendants](https://www.w3.org/TR/css-transforms-1/)). Not sure how to fix that, got any ideas?

elbotho:
@Knorrke thanks! 

> Got a different issue though, with the plugin overlay position: fixed. The transform of react-virtual is messing with that (transformed parents act as a containing block for fixed descendants). Not sure how to fix that, got any ideas?

What overlay do you mean? Could you explain the problem? I did not find it.
Because this is an importat fix we will probably also deploy over at [legacy](https://github.com/serlo/serlo.org). 
Do you want to do that to earn yet another hacktoberfest PR or should I do it? :)

Knorrke:
The plugin overlay inside of a solution. It should be positioned over the full page, which was done with `position: fixed; width:100vw; height: 100vh;`, but because of the transform of react-virtual it now looks like this:
overlay-bug

I'll update legacy (Hacktoberfest is already over, but I got there with this PR 🎉😅)

elbotho:
> I'll update legacy (Hacktoberfest is already over, but I got there with this PR 🎉😅)

Congrats :) and thanks.
Ah yeah, now I see the problem 😅

elbotho:
@Knorrke I'm not at all deep into this problem but using position relative and top:… seems to work for me.
Could you check that?

``` tsx
<div
  key={virtualRow.index}
  ref={virtualRow.measureRef}
  style={{
    position: 'relative',
    top: `${virtualRow.start}px`,
    left: 0,
    width: '100%',
  }}
>
```

Knorrke:
Thanks for taking a look, However I'm losing some in the long document `/entity/repository/add-revision/2373`, when I try that  (exercises (a) - (f) are missing).

Actually, now that I looked at it again, I think the virtualization isn't even working right now... all the exercises are rendered anyway. It "works" if you change the hight to a fixed size here https://github.com/serlo/frontend/blob/7a55f1b0e66df51a63e562c7eb0d6c54c2695f0b/src/edtr-io/plugins/types/text-exercise-group.tsx#L148
But that results in a scroll pane.

There is an open PR with some suggestions on how to use the window scroll instead (see [this gist](https://gist.github.com/mogelbrod/7029848de62016bf0f0b38fb56f8da3e)), I'll take a closer look later. If I can't get it to work right away, I'll probably just remove react-virtual for now to fix the issues and then we have time to reimplement it properly.

elbotho:
> I'll probably just remove react-virtual for now to fix the issues and then we have time to reimplement it properly.

I was thinking the same thing.
Maybe we could just start with all the solutions folded in the text exercise group.
That way it should render quickly **and** give users a better overview.

kulla:
> I'll probably just remove react-virtual for now to fix the issues and then we have time to reimplement it properly.

For background information: @inyono has introduced it because long exercise groups couldn't be edited well due to too many DOM rerenderings. One should check whether editing long exercises groups like https://de.serlo.org/mathe/225387/225387 or https://de.serlo.org/mathe/2513/2513 works well (ping also @inyono )

Knorrke:
@kulla Thing is, that it currently still renders all the elements, so we don't lose performance by removing it. I agree, that we could improve performance in long documents if we implement it correctly. The default implementations of `react-virtual` is for usage in inner scroll handles, and we would need it to use the window scroll instead, so it's a bit more work.

> Maybe we could just start with all the solutions folded in the text exercise group.
> That way it should render quickly **and** give users a better overview.

@elbotho The nesting makes this quite complicated. I could collapse all solutions by default when editing? Don't know if that would be annoying...

elbotho:
Hmm, for Groups I think it's better.
And I'd say it's okay for single exercises. 
Try it and see if the community freaks out ;)

elbotho (approved):
👍 thank you!

PR #1362, opened 2021-11-01
Remove "Physik" from navigation "Fächer"

kulla:
See https://community.serlo.org/channel/physik?msg=gSZs4pQ4nLTKFAt5J


elbotho (commented):
I would wait with this until Physik is in here:
https://de.serlo.org/community/106081/neue-f%C3%A4cher-und-themen

Otherwise it's really hard to find.

PR #1363, opened 2021-11-02
show empty courses with warning

### LINKED ISSUE #1327: broken course
fails with "without message":
https://de.serlo.org/informatik/138931/erweiterungskurs-messen-und-bewegen

### LINKED ISSUE #1337: 404 for course in /entity/unrevised with first course page has no current revision yet
In `/entity/unrevised` there is a link to the course https://de.serlo.org/informatik/227404/versionskontrolle-mit-git which has only one course page without any revised revision yet

```graphql
{
  uuid(id: 227404) {
    __typename
    ... on Course {
      pages {
        id
        currentRevision {
          id
        }
        revisions {
          nodes {
            id
          }
        }
      }
    }
  }
}
```

Result:

```json
{
  "data": {
    "uuid": {
      "__typename": "Course",
      "pages": [
        {
          "id": 227406,
          "currentRevision": null,
          "revisions": {
            "nodes": [
              {
                "id": 227408
              }
            ]
          }
        }
      ]
    }
  }
}
```

Click on the link in `/entity/unrevised` results in a 404 error page. @elbotho Do you have any clue what the desired behavior shall be? Shall I filter out `Course` entities in `unrevisedEntities` and only return `CoursePage`?! Shall it be handled in the frontend?

elbotho:
fixes #1337 and #1327.

should give users at least some feedback (instead of 404) when:
- the course has only trashed pages
- only unrevised pages (even with links for reviewers)
- no page at all

also those pages are hidden from search engines.

### DIFF
--- a/src/components/content/entity.tsx
+++ b/src/components/content/entity.tsx
@@ -1,6 +1,13 @@
-import { faTools, faTrash } from '@fortawesome/free-solid-svg-icons'
+import { IconProp } from '@fortawesome/fontawesome-svg-core'
+import {
+  faExclamationCircle,
+  faTools,
+  faTrash,
+} from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import clsx from 'clsx'
 import dynamic from 'next/dynamic'
+import Head from 'next/head'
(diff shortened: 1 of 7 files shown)


Entkenntnis (approved):
thank you - works fine

PR #1367, opened 2021-11-03
refactor static info notices into component

### LINKED ISSUE #1369: should we hide trashed content from robots?
They are usually trashed for a reason, so it should make sense to hide them on e.g. google, right? @yolo-bird @kathongi

elbotho:


Entkenntnis (approved):
Thank you!

Issue #1351, opened 2021-10-26
Investigate: Are there legacy interactive exercises left?


elbotho:

elbotho:
> 2635 / 2970 Aufgaben konvertiert
> 1784 / 1790 Aufgabengruppen konvertiert
> 950/ 1796 Artikel konvertiert

last convo update I could find (end of september)

inyono:
https://are-we-edtr-io-yet.serlo.org

Entkenntnis:
I'm 100% sure that frontend + api are not supporting legacy exercises. (we haven't even got a `__typename` for them)

@kulla And I think we are also filtering them out in the api? Maybe there are some active uuids left over ...

elbotho:
those numbers don't differenciate between interactive and and non-interactive exercises.

elbotho:
Note: Could not find any Exercises that did not work with our implementation. If we find exercises that need that code (0cc1b7b8e53d337a9bc221c3ef5b7b7f0fa6217f) later, I'll suggest adding a warning and redirect to legacy.

PR #1377, opened 2021-11-04
Todo hunting

elbotho:
fixed or ignored all but ~two~ one issue:

- ~freeze/focus issue in equations plugin~
- ~unconverted exercise code in `editor-response-to-state.ts`~ (related issue: #1351)

Note: Could not find any Exercises that did not work with our implementation. If we find exercises that need that code later, I'll suggest adding a warning and redirect to legacy.

### DIFF
--- a/src/edtr-io/plugins/equations/editor.tsx
+++ b/src/edtr-io/plugins/equations/editor.tsx
@@ -85,11 +85,12 @@ export function EquationsEditor(props: EquationsProps) {
 
   React.useEffect(() => {
     if (nestedFocus) {
-      // TODO: fixing freeze, investigate why this is deadlooping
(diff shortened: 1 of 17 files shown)


src/edtr-io/plugins/equations/editor.tsx
```diff
@@ -198,13 +199,12 @@ export function EquationsEditor(props: EquationsProps) {
 
                       return (
                         <ExplanationTr
-                        /* TODO: investigate why this is not working */
-                        /*onFocus={() =>
```
elbotho, line 202: @Entkenntnis could you check if this works for you now?
I'm not 💯 % sure I understood the problem :)

Entkenntnis: thank you, it's working fine

Issue #1401, opened 2021-11-10
Refactor Error handling: use next approach


elbotho:
Follow up of #1391:

We should probably use the new ['notFound'](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) return and [next's custom error pages](https://nextjs.org/docs/advanced-features/custom-error-page#500-page)
 and maybe implement something like [this](https://victorthadeu-66076.medium.com/how-to-return-404-while-using-fallback-blocking-in-next-js-6fda63aad818).
 
Why:
- We should really not return status code 200 for 404 errors
- easier to understand if we follow the documented approach

Entkenntnis:
I will look into this and see if we can use some of the approaches.

Entkenntnis:
So my idea would be to remove the error-type from page data -> throw error instead and use nextjs error page. Hopefully this works ... (not sure yet how to pass the message)

Entkenntnis:
Proposed new behaviour in detail (wip):

- `404`: return `notFound: true`, generic message that page is not found (e.g. no custom message)
- `5xx`: throw error in getStaticProps instead of returning, nextjs will catch it, custom error component extracts error code, sets it to response and renders error page, no caching. In development: dev-server will show overlay.
- removing `kind:'error'` from data-types, because its not needed anymore.

PR #1402, opened 2021-11-10
1s revalidation time for 50x status codes

### LINKED ISSUE #1391: don't cache 500 pages
I think we are currently generating static pages even for 500 results - better set it to error and don't cache them.

elbotho:
- closes #1391
- bonus: noindex for our error pages (upsi)

follow up: #1401


Entkenntnis (approved):
thank you

PR #1404, opened 2021-11-10
Editor improvements

### LINKED ISSUE #1403: Feedback: editor integration
[Source](https://serlo.slack.com/archives/CFF72UKJL/p1636456208004000):
- [x] save/undo/redo toolbar should be fixed (Note: that is actually the expected behaviour already, let's reinvestigate when we have examples where this does not work)
- [x] equations plugin: visual difference between editor and rendering (arrow color…?)
- [x] how are quotes styles in editor vs frontend?
- [x] drop downs should be marked more clearly (e.g. answer type of input ex)
- [x] link not marked here:
image

elbotho:
closes #1403

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -53,5 +53,5 @@ module.exports = withBundleAnalyzer({
     localeDetection: false,
   },
   reactStrictMode: true,
-  productionBrowserSourceMaps: true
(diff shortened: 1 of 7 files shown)


Entkenntnis (approved):
thank you!

PR #1406, opened 2021-11-10
use window width to conditionally load mobile menu

elbotho:
- this way we don't load tippy on mobile on first load. -> might also help with #1405
- also: report failed tippy chunk load to sentry directly


elbotho:
closed for now.

PR #1411, opened 2021-11-12
Editor box plugin

elbotho:
[add plugin here](https://frontend-git-editor-box-plugin-serlo.vercel.app/en/entity/repository/add-revision/193587)

- needs two click on title or content to start editing -> known upstream issue
- creates a random id on creation that is saved and can be used to anchor link to box (safer than relying on title that might be empty or might change)
- only allows text, images, equations and tables in content


Entkenntnis:
really nice demo :100:

Entkenntnis:
@elbotho Plugin itself looks good. for full support, we still need to add it to the converter (https://github.com/serlo/frontend/blob/staging/src/schema/convert-edtr-io-state.tsx), data types (https://github.com/serlo/frontend/blob/staging/src/data-types.ts) and renderer (https://github.com/serlo/frontend/blob/staging/src/schema/article-renderer.tsx).

this duplicated work is not optimal, but the current state of the frontend (and now with the migration there is some potential to refactor this)

elbotho:
@Entkenntnis thanks!
Have the frontend part on my list, but wanted to finalize the editor plugin first.

PR #1416, opened 2021-11-15
spam: lock profile desc for new users

elbotho:


Entkenntnis (commented):
thank you, hopefully this discourage spammers

PR #1423, opened 2021-11-22
Deployment

elbotho:
- **injections** get a border to better seperate them from each other other
- experimental feature: editor box plugin
- fix [problem](https://github.com/serlo/frontend/issues/1390): don`t load comment area with undefined ids any more
- add JSON-LD metadata


elbotho:
okay box plugin should be hidden in production for everybody now. 
can be merged imho.

elbotho:
okay, should be ready now.
box plugin is hidden in production.

PR #1427, opened 2021-11-26
Editor box plugin: frontend

elbotho:
[226222](https://frontend-git-editor-box-plugin-frontend-serlo.vercel.app/226222)

### DIFF
--- a/src/edtr-io/plugins/box/renderer.tsx
+++ b/src/edtr-io/plugins/box/renderer.tsx
@@ -14,7 +14,7 @@ import { useInstanceData } from '@/contexts/instance-context'
 import { useLoggedInData } from '@/contexts/logged-in-data-context'
 import { hasOwnPropertyTs } from '@/helper/has-own-property-ts'
 
-const boxTypeStyle = {
+export const boxTypeStyle = {
   blank: {},
   example: {},
   quote: { icon: faQuoteRight },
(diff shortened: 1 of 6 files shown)


Entkenntnis:
bundlesize again -_-

increased to 180kb, some of the imports is pulling in a lot of dependencies into the [...slug]

src/edtr-io/plugins/box/renderer.tsx
```diff
@@ -14,7 +14,7 @@ import { useInstanceData } from '@/contexts/instance-context'
 import { useLoggedInData } from '@/contexts/logged-in-data-context'
 import { hasOwnPropertyTs } from '@/helper/has-own-property-ts'
 
-const boxTypeStyle = {
+export const boxTypeStyle = {
   blank: {},
   example: {},
   quote: { icon: faQuoteRight },
```
Entkenntnis: maybe moving this out of this file into the frontend will do the trick (and import from there)

elbotho: ah damn, was afraid that might happen. we really have to take care when we try to split editor plugins into a render and editor part in the future.

elbotho:
okay, we are at `+2kb` compared to staging now.

Entkenntnis:
thank you 👍

PR #1433, opened 2021-12-09
Add SingleSignOn with Keycloak

hugotiburtino:


hugotiburtino:
Bug: in the PR's endpoint, Firefox disables the keycloak cookies after a while, so that the website will behave after a while as if the user were logged out.
```
Cookie "kcToken" will be soon rejected because it has the "SameSite" attribute set to "None" or an invalid value, without the "secure" attribute.
```

~~Possible solution: maybe it will be automatically solved when deploying in the same domain of keycloak. ~~

Clarification: it happens only when user sets strict cookies policies

inyono:
Branch can be deleted now since we updated the redirects in CF Worker.

PR #1444, opened 2022-01-17
user: Hide edit buttons for newly registered user

kulla:
Most newly registered users who want to edit their profile are currently
spam accounts. This PR hides edit buttons for profile picture and
motivation text for the first hour after registration. This might help
that most spam accounts change the description which we can easily
detect and which does not effect other authors (most people will not see
these texts currently)


elbotho:
I'll merge this as a test run. 
There are UX problems here (Currently new users can't change their avatar and might not know why).

In the long term I'd prefere to allow all changes to the profile but only show them after some time has passed. This way moderators can identify spammers, but users don't see the spam.

PR #1457, opened 2022-02-03
experimental feature: old design

elbotho:
with this user can switch to the old design in the "exerimental features" section on their profile. I thought that's a fitting place in the future so power users can discover it.


Entkenntnis (approved):
lgtm

PR #1458, opened 2022-02-07
use new mutation to update profile from frontend

elbotho:
- testing needs authed user so test locally for now.
- http://localhost:3000/user/settings


inyono:
Is the save modal supposed to look like this? (macOS, Chrome)
image

inyono (commented):
Apart from the question regarding the modal, it works well :)

elbotho:
> Apart from the question regarding the modal, it works well :)

Thanks! Good catch.

PR #1466, opened 2022-02-17
Yarn v3

inyono:


inyono:
Okay, `yarn --immutable --immutable-cache` fails because of SWC somehow. Removed that checks and seems to work otherwise.

elbotho:
Fast, but the output is quite noisy 😅

Not sure about `yarn --immutable --immutable-cache`. Where was the check before?
Seems to run locally for me.

inyono:
`yarn --immutable --immutable-cache` was part of the CI before, but failed because some SWC linux stuff was not part of the cache. I presume that yarn is not caching native modules per platform or something, idk. Removed the checks and just did a normal `yarn`.

PR #1487, opened 2022-03-10
experiments: simpler legacyDesign, add legacyEditor

elbotho:
should be deployed with https://github.com/serlo/serlo.org-cloudflare-worker/pull/270


Entkenntnis (approved):
ok, code looks fine! will do some testing on staging

elbotho:
Thanks!

Issue #1493, opened 2022-03-14
Migrate Content API


Entkenntnis:
we need to support (parts of) https://github.com/serlo/serlo.org/wiki/Content-API for edtr-io integration in the frontend

Idea: use CF to redirect to a separate `/embed/[...slug]` route

Entkenntnis:
most important two parameters atm are `contentOnly` and `hideTopbar`

elbotho:
https://github.com/serlo/serlo.org-cloudflare-worker/pull/280

elbotho:
We decided on a **feature set**:
only `?contentOnly` will be supported and no height info will be sent. 
Others features/customization we will implement only when requested.

The **route** in the frontend:
`/content-only/[...slug]`

**Links**:
New tab or staying in the iframe: (pending)

Entkenntnis:
Some thoughts about page types: content-only is available to entities like article, taxonomy or page. For other pages, like the landing page, we will drop support completely, right? In this case the frontend part is done and this can be handled in the cloudflare worker issue.

elbotho:
> For other pages, like the landing page, we will drop support completely, right

We did not talk about this on monday, but yes that's my suggestion atm. 
Goes well with the choice to handle links in new tabs :)

elbotho:
Can now be tested in staging:
https://de.serlo-staging.dev/mathe/36070/rechtwinkliges-dreieck?contentOnly

Entkenntnis:
fixed an issue where csr links are not opened in new tabs: da775aad90f1ca015911196440bae12444bea100

Otherwise, I haven't noticed any technical issues

kulla:
Live in production.

PR #1498, opened 2022-03-15
content api quick test

elbotho:
proof of concept for https://github.com/serlo/frontend/issues/1493

### [Example](https://frontend-git-content-api-serlo.vercel.app/embed/1555)

### DIFF
--- a/src/components/user-tools/user-tools.tsx
+++ b/src/components/user-tools/user-tools.tsx
@@ -88,7 +88,7 @@ export function UserTools({
     return (
       <nav
         className={clsx(
-          'mr-4 -mt-4 mb-8',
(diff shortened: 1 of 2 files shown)


elbotho:
There is some duplicated code (but acceptable imho) and this would only support `taxonomy` and `single-entity`.

Todo:
We should really find out what urls are actually being used atm. and then define a feature set we want to support in the future.
Other than that this might just work. Nice idea @Entkenntnis!

Entkenntnis (approved):
straight forward implementation and a good starting point atm. Merging into staging/production shouldn't affect anything

elbotho:
I guess I'll add a noindex for all those pages just to make sure.

PR #1508, opened 2022-03-15
add qr code to share modal

elbotho:
seems works quite well.


Entkenntnis (changes requested):
Thanks! Looks straight forward. Can you check the mobile layout? It's breaking a bit for me, the modal is leaving the browser.

elbotho:
thanks, should work now.

PR #1515, opened 2022-03-16
hotfix: small courses

elbotho:
Bug report: https://community.serlo.org/channel/software-features-and-bugs?msg=ukZSykL6eD3nv3yLZ

I think currently courses with less than 4 pages fail (also in production).

Live-Example: https://de.serlo.org/community/241357/den-pc-verstehen

This fix should solve this issue.
I'm not sure why it happened but I guess it's related to my changes from yesterday.

I think we should deploy this to production directly since staging is quite busy already ^^

### DIFF
--- a/src/components/content/entity.tsx
+++ b/src/components/content/entity.tsx
@@ -35,13 +35,17 @@ export function Entity({ data }: EntityProps) {
   // state@/components/comments/comment-area
   const [open, setOpen] = useState(false)
 
-  // course
-  const [courseNavOpen, setCourseNavOpen] = useState(false)
+  // courseNav: start opened when only some entries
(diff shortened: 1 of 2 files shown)


Entkenntnis:
"Maximum update depth exceeded", you are setting the parents state while rendering, so this line

```
if (data.pages.length < 4) onOverviewButtonClick()
```

is causing this, good to know. Will deploy it now to production

Entkenntnis (approved):
:+1:

elbotho:
yeah, only this line is not new at all… not sure why it's causing a problem now.
`if (data.pages.length < 4) onOverviewButtonClick()`

thanks for deploying.

Entkenntnis:
looks like this bug existed for 4 months and nobody noticed :eyes:

PR #1518, opened 2022-03-16
use next eslint config

### LINKED ISSUE #1499: next.js eslint plugin
"The Next.js plugin was not detected in your ESLint configuration. See https://nextjs.org/docs/basic-features/eslint#migrating-existing-config"

Some rules seems useful, other are not relevant for us (e.g. Link or Image component). Check it someday.

Entkenntnis:
I have disabled img component rule, but the others are useful.

### DIFF
--- a/src/components/pages/consent-page.tsx
+++ b/src/components/pages/consent-page.tsx
@@ -1,5 +1,6 @@
 import { useState } from 'react'
 
+import { Link } from '../content/link'
 import { PageTitle } from '../content/page-title'
 import { useInstanceData } from '@/contexts/instance-context'
 import { replacePlaceholders } from '@/helper/replace-placeholders'
@@ -21,9 +22,7 @@ export function ConsentPage() {
       <p className="serlo-p">
         {replacePlaceholders(strings.consent.intro, {
           privacypolicy: (
-            <a className="serlo-link" href="/privacy">
-              {strings.entities.privacyPolicy}
-            </a>
+            <Link href="/privacy">{strings.entities.privacyPolicy}</Link>
           ),
         })}
       </p>
--- a/src/components/pages/consent-page.tsx
+++ b/src/components/pages/consent-page.tsx
@@ -22,7 +22,9 @@ export function ConsentPage() {
       <p className="serlo-p">
         {replacePlaceholders(strings.consent.intro, {
           privacypolicy: (
-            <Link href="/privacy">{strings.entities.privacyPolicy}</Link>
+            <Link href="/privacy" forceNoCSR>
+              {strings.entities.privacyPolicy}
+            </Link>
           ),
         })}
       </p>
--- a/src/components/pages/error-page.tsx
+++ b/src/components/pages/error-page.tsx
@@ -1,7 +1,7 @@
 import Head from 'next/head'
-import Link from 'next/link'
 import { useState, useEffect } from 'react'
 
+import { Link } from '../content/link'
 import { PageTitle } from '../content/page-title'
 import { HSpace } from '@/components/content/h-space'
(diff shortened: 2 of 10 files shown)


src/components/pages/consent-page.tsx
```diff
@@ -21,9 +22,7 @@ export function ConsentPage() {
       <p className="serlo-p">
         {replacePlaceholders(strings.consent.intro, {
           privacypolicy: (
-            <a className="serlo-link" href="/privacy">
-              {strings.entities.privacyPolicy}
-            </a>
+            <Link href="/privacy">{strings.entities.privacyPolicy}</Link>
```
elbotho: this was intentional since `/privacy` is and will not be handled by the frontend.
I felt using <a> was a simple and clear way to signal that. A `noCSR` on the link would probably work well.

src/components/pages/error-page.tsx
```diff
@@ -89,12 +90,11 @@ export function ErrorPage({ code, message }: ErrorPageProps) {
   function renderHomeLink() {
     //no csr here
     return (
-      <a
-        href="/"
-        className="serlo-button serlo-make-interactive-primary mr-4 mt-4"
-      >
-        {strings.errors.backToHome}
-      </a>
+      <Link href="/">
```
elbotho: Also intentional. The idea is to have a clean and save refresh after erroring.
So i'd prefer `<a>` or `noCSR`

src/components/pages/consent-page.tsx
Entkenntnis: We are currently having all this settings in one place within the `/content/link.tsx` file and `/privacy` is part of legacy already, so I thought it's fine to reuse this component here.

src/components/pages/error-page.tsx
Entkenntnis: ah, this looks like another use-case, in which case a prop like `forceNoCSR` is indead useful.
-> and a prop like `unstyled` to avoid using the `UnstyledLink` component

src/components/user/profile-experimental.tsx
```diff
@@ -73,12 +75,12 @@ export function ProfileExperimental() {
         <div>
           <h3 className="serlo-h3 mb-3">
             👻 Frontend: Altes Design{' '}
-            <a
+            <UnstyledLink
```
Entkenntnis: add a new prop unstyled to Link?

src/components/pages/consent-page.tsx
elbotho: since it's not a legacy link I'd like to make this more explicit. 
Otherwise I'm pretty sure I will remove the legacy logic one day without thinking about this :)

src/components/pages/consent-page.tsx
```diff
@@ -21,9 +22,9 @@ export function ConsentPage() {
       <p className="serlo-p">
         {replacePlaceholders(strings.consent.intro, {
           privacypolicy: (
-            <a className="serlo-link" href="/privacy">
+            <Link href="/privacy" forceNoCSR>
```
Entkenntnis, line 25: ok, let's add the prop here to make this more explicit

src/components/pages/error-page.tsx
elbotho: Notes and thoughts: 
- when I remember correctly @kulla refactored unstyled into an extra component in an effort to simplify the link component
- I think it's a matter of code style / preferences. You seem to prefer an all in one solution, Kulla wanted slim components with a single purpose and I liked the simplicity of just using a native tag for a native function. 🤷‍♂️

elbotho (approved):

PR #1525, opened 2022-03-18
hotfix editor csrf

elbotho:


src/components/pages/add-revision.tsx
```diff
@@ -16,6 +17,23 @@ export function AddRevision({
...
+  useEffect(() => {
+    if (window.location.hostname === 'localhost') {
+      setCookieReady(true)
+    } else {
+      fetch('/auth/password/change')
+        .then((res) => res.text())
+        .then(() => {
+          setCookieReady(true)
+        })
+        .catch(() => {})
```
inyono, line 31: Possible optimization: check if the cookie already exists.

inyono (commented):
LGTM

src/components/pages/add-revision.tsx
elbotho: sure it can not be outdated or smth?

elbotho: will add optimization later via staging.

PR #1531, opened 2022-03-21
create taxonomy term

### LINKED ISSUE #1511: create taxonomy term
final place: `/taxonomy/term/create/[parent]/[id]`

is not reachable from the frontend, so we need cf support to test this

CF issue: https://github.com/serlo/serlo.org-cloudflare-worker/issues/274

Entkenntnis:
todo: find out what info is needed for this


Entkenntnis:
testing this in staging

PR #1532, opened 2022-03-21
use new endpoints for adding revisions

### LINKED ISSUE #1523: use new mutations for edtr-io add-revision

elbotho:
### Todos

- [x] wait for new api release to get actual types
- [x] move onsave function logic to mutation
- [x] use `AddGenericRevisionInput` to build a base type and extend inputs only where needed
- [x] find a way to close save modal or redirect
- [x] fix next onbeforeunload problem
- [x] add as an exerimental feature (opt-in)
- [x] dirty state does not change after saving
- [x] check `needsReview` again
- [x] quick test for all types

### Bugs
- if `needsReview === true` API responds positive, but does not show a revision ([PR](https://github.com/serlo/api.serlo.org/pull/537/files))
 

### Backlog
- investigate error handling (how can we better prevent data loss)
- once we rely on this: remove unused code, refactor
- check types (e.g. we should probably have a type for supported `__typename` in the frontend, currently it's always string)

(closes #1523)


elbotho:
Testing the different types revealed another big TODO.
The legacy endpoint could handle updating an entity and it's "children" (e.g. `Course` and it's `CoursePages`) at the same time. The new endpoints only save the current entity. 
We need to replicate this logic either in the frontend or in the API.


Probably working: `Article, CoursePage, Video, Applet, Solution`

Not yet: `TaxonomyTerm, Exercise, ExerciseGroup, GroupedExercise, CoursePage`

elbotho:
🎉
Okay, with recursive calls all types now pass a superficial test.

elbotho:
This is currently waiting for more changes in the API.

elbotho:
@Entkenntnis I'd like to test this in staging. Since this is opt-in and deactivated in production it should be save.
Vetos or requests?

Entkenntnis (approved):
thank you! code looks good. will probably start to test it this week.

PR #1544, opened 2022-03-28
Box plugin

elbotho:


src/edtr-io/plugins/table.tsx
```diff
@@ -21,7 +14,7 @@ function MarkdownRenderer(props: { markdown: string }) {
   return <>{renderArticle(node.children)}</>
 }
 
-export const tablePlugin: EditorPlugin<TablePluginState, TablePluginConfig> = {
+export const tablePlugin = {
```
inyono: This change might be the reason why you need the `as` in `plugins.tsx`.

elbotho: thanks, I think this was solved after using the new edtr version

PR #1550, opened 2022-04-01
Table plugin rework

### LINKED ISSUE #1519: finish table plugin
we are waiting for changes in the edtr-io core for this.

elbotho:
### Demo
Can be tested without login (also no saving) here:
https://frontend-git-table-plugin-serlo.vercel.app/entity/repository/add-revision/236956
(don't mind the old tables :)

### Changes
- shared renderer for frontend, active and inactive editor states (maybe a good start for our refactoring)
- now renders in frontend
- single click now focuses slate
- tab navigation works (mostly)
- simplified data structure (there are only rows and cells, headers are defined through table type setting)
- ui for editors should be pretty much finished
- …

### TODOs
- needs a lot more testing (images, table types…)
- solution for "responsive" tables


Entkenntnis:
A note about sharing code between edtr-io and frontend: using the renderer directly is unfortunately bloating the bundle size to 255kb (+100kb), because it has to include a lot of edtr-io code. Bundlewise, its easier to write the renderer in the frontend and then use the frontend components in the edtr-io plugin (and avoiding duplication this way) hope this helps

elbotho:
@Entkenntnis I tried to find a solution that does not include any edtr code ([see file](https://github.com/serlo/frontend/pull/1550/files#diff-ac7db6c78af6f5f7ef242f2dbd3bf7934cc0e9cba1be3328d95147079d4ff4ee)). 
You are right, atm. the bundle size increases but I'm confident this could be avoided with this approach.

Edit: It's basically does what you describe. The difference is:
- it lives with the plugins code (makes sense imo)
- it accepts JSX elements to make it usable for both editor and frontend
but it completely uses frontend components, styling etc.

elbotho:
yay :) importing this adds 100kb
``` ts
export enum TableType {
  OnlyColumnHeader = 'OnlyColumnHeader',
  OnlyRowHeader = 'OnlyRowHeader',
  ColumnAndRowHeader = 'RowAndColumnHeader',
}
```

edit: we are back to normal sizes again.

Entkenntnis:
hey, thanks for all the effort! overlooked that `@/edtr-io` is not the same as `@edtr-io` :) and the renderer is indeed completely standalone, which is very nice!

elbotho:
Dear @kulla. I ended up changing a lot in the table plugin. Maybe you want to have a look. The plugin files especially:
- [Reduced renderer](https://github.com/serlo/frontend/blob/8c21d0fc412778e5dac8c30495a44b15e05ab6e2/src/edtr-io/plugins/serlo-table/renderer.tsx)
- [Plugin Index](https://github.com/serlo/frontend/blob/8c21d0fc412778e5dac8c30495a44b15e05ab6e2/src/edtr-io/plugins/serlo-table/index.tsx)

The rest ist mostly types and the frontend rendering part.

PR #1553, opened 2022-04-02
style icons with tailwind and use thin layer

### LINKED ISSUE #1552: review icon implementation
Just adding font-awesome is using up 16kb gzipped bundle size, which is huge (and not even includes the icon data of ~10kb) for features we are barely using ...

https://github.com/FortAwesome/react-fontawesome/issues/232 is a start

Entkenntnis:
This is a small poc for replacing fontawesome-svg-core with a thin layer and style everything with css and tailwind.

Decreases bundle size by ~17kb js and ~1kb css and gives us a lot of control back.

Result: new bundle size is 144kb gzip for slug

- [x] spin -> tailwind spin
- [x] flip -> tailwind transform scale -1
- [x] automatic sizing of icons, (only useful if icon is dynamic and size should vary) -> height:1em is doing the trick
- [x] title attribute removed, because it was only used in one place -> wrap in span instead


elbotho:
Thanks! Very interesting. I think we only really use `size` and `spin` (in one or two places ). 
Wishes:
- a more descriptive name for the component that includes the word `Icon`. 
- if we go ahead with this we should maybe do it after deploy and merging of the two open PRs to avoid conflicts.
- title attributes

edit: there is a fancy use of layers in `license-notice`. but we'll find another solution here.

Entkenntnis:
I'll wait with this after the current deployment. (we can ignore the increased bundle size if we fix it in the next deployment)

elbotho:
while we are at it :)
https://fontawesome.com/docs/web/dig-deeper/accessibility

elbotho:
> I'll wait with this after the current deployment. (we can ignore the increased bundle size if we fix it in the next deployment)

I just checked. Only two files would be affected. So you can go ahead if you want.

src/components/landing/subjects/subject-landing-featured.tsx
```diff
@@ -93,8 +93,6 @@ export function SubjectLandingFeatured({
 
   function renderTypeIcon(type: string) {
     const icon = getIconByTypename(type)
-    return (
-      <FontAwesomeIcon icon={icon} title={getTranslatedType(strings, type)} />
-    )
+    return <FaIcon icon={icon} />
```
elbotho, line 96: removed title

elbotho (changes requested):
one title missing, rest looks good 👍 

I'd want the title attribute at some point because we can use it to set the [screen reader visibility accordingly](https://fontawesome.com/docs/web/dig-deeper/accessibility) but that can be a different PR.

src/components/landing/subjects/subject-landing-featured.tsx
Entkenntnis: I have removed the title here because the type string is visible directly next to it and a title would just be a duplicate

Entkenntnis:
not a big fan of the title-attribute itself: https://www.tpgi.com/using-the-html-title-attribute-updated/

should use `aria-label`, etc, ... as appropriate, but it's most of time not the job of the icon to handle this...

src/components/landing/subjects/subject-landing-featured.tsx
elbotho: Check. Thanks for the explanation.

elbotho:
I understand your resentment :)

Most of our icons are decorative anyway (by design).
At some point I'd like to go through the ones that are not (e.g. the Notification Bell) and make sure they have labels.

elbotho (approved):

PR #1555, opened 2022-04-03
Comments: Edtr text-plugin

elbotho:
revived version of #1212

`+1kb` in [...slug] first load compared to staging


elbotho:
@Entkenntnis if you think this is ok for testing in staging I'll add the auth-check again and merge.

Entkenntnis:
I remembered that we had some issues with the table plugin (because the converter is loading the markdown parser etc..) that where difficult to separate - but with the new table puglin this shouldn't be a big issue soon. I'd like to wait after the table plugin is done, before we continue here...

PR #1572, opened 2022-04-05
use internal import for icons

Entkenntnis:
slightly better code splitting and tree shaking (-2kb gzipped) - close to the optimum splitting strategy.

Probably not worth enforcing all the time - but if reviewed time by time this helps scaling icon usage.


elbotho:
Nice :)

> Probably not worth enforcing all the time - but if reviewed time by time this helps scaling icon usage.

Does it optimize also when single Icons are not imported like this?

Entkenntnis:
Yes, it still optimizes the existing icons

PR #1583, opened 2022-04-07
Add revision changes

### LINKED ISSUE #1574: don't add new revision if there was no change
the new implementation adds a new revision for every exercise/solution in a exercise-group - even if there was no change.

Do we have enough information to detect this and skip unchanges entites? This will keep the revision history cleaner.

There is also the option to move this check into the API?

elbotho:
did some refactoring and added the check if children of the edited entity actually changed before adding new revisions for them.


Entkenntnis (approved):
thank you! code looks good, will test it in staging env

PR #1585, opened 2022-04-08
setup graphql codegen

elbotho:
some wips:

- ~create-title is also used in revision and subscriptions~ done
- ~dedupe dependencies~ done
- ~breadcrumbs is used to fetch editor data~ marked as todo
- ~revision query not finished yet~ done
- ~manage subscriptions with custom title finder~ replacement done
- test if everything is fine


Entkenntnis:
@elbotho thank your for the preparing work and sorry for the late review!

I would really like to use codegen to autogenerate query types and graphql codegen looks like a great candidate :)

I'm happy to help with integrating this, I think the biggest challenges will be the refactoring of our code to use the generated types, this could take some time (but I'm glad if we can delete all manually written query codes)

elbotho:
> I'm happy to help with integrating this, 

Nice! 🎉

> I think the biggest challenges will be the refactoring of our code to use the generated types, this could take some time 

That's true. Also we will have to fix some typescript errors in the process for sure.
But we should be able to do this iteratively query by query.

Entkenntnis:
I'll try to start with the main query. If this is successful, then we can merge this and continue to work iteratively.

Entkenntnis:
Main query type checks :100:, but it's still messy

some thoughts:

- dev experience is really nice! I can change the raw query, run `yarn codegen`, and the types are magically correct
- some typescript magic like `Omit` or `Extract` is needed to get correct types
- It reveals some subtle bugs and inconsistencies we are having in our `create-XXX` functions -> a good time to review all this.

todo:

- [x] review this PR slowly and check if everything is fine
- [x] fix test

package.json
```diff
@@ -9,7 +9,7 @@
   "license": "Apache-2.0",
   "author": "Serlo Education e.V.",
   "scripts": {
-    "_eslint": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
+    "_eslint": "eslint \"src/edtr-io/**/*.{js,jsx,ts,tsx}\"",
```
elbotho: on purpose?

src/components/pages/manage-subscriptions.tsx
```diff
@@ -90,3 +92,30 @@ export function ManageSubscriptions({
...
+    object.__typename == 'Applet' ||
+    object.__typename == 'CoursePage' ||
+    object.__typename == 'Course' ||
+    object.__typename == 'Event'
+  ) {
+    return object.currentRevision?.title
+  }
+
+  return null
+}
```
elbotho: this whole block is more or less duplicated now, right?

elbotho:
I renamed `MainPageQuery` to `MainUuidQuery` and tried to simplify some types. 

Other than the commented: LGTM.
Thank you for proceeding with this.

package.json
Entkenntnis: reverted

src/components/pages/manage-subscriptions.tsx
Entkenntnis: I have recombined the queries (good an idea how to do this :)

PR #1601, opened 2022-04-21
add confirm-promt to trash

### LINKED ISSUE #1596: Show warning prompt before deleting an entity
After clicking "Other tools -> move to recycling bin": 
2022-04-20-125715_414x344_scrot
there shall be a prompt warning the author that the entity shall be deleted. Only after accepting the prompt the mutation shall be executed

elbotho:
closes #1596


Entkenntnis (approved):
Thanks! It's doing it's job (maybe not that pretty, but functional and easy)

PR #1605, opened 2022-04-21
enable mhchem

### LINKED ISSUE #1604: enable mhchem extension
katex as some extended chem support: https://mhchem.github.io/MathJax-mhchem/?msclkid=cc156accc18e11ecbdc645558b42d383

It was enabled in legacy, but not anymore in edtr-io.

Entkenntnis:
this looks like everything we need to enable mhchem in frontend + edtr-io


Entkenntnis:
some testing on staging

PR #1608, opened 2022-04-22
Plugins: pageLayout & pageTeam

### LINKED ISSUE #1643: Static pages: Add column layout plugin

elbotho:
one of the blockers for converting static pages to edtr-io is the ability to place content side by side

https://frontend-git-page-layout-plugin-serlo.vercel.app/entity/repository/add-revision/18922


src/edtr-io/plugins/page-layout/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  number,
+  object,
+  EditorPluginProps,
+} from '@edtr-io/plugin'
+
+import { editorContent } from '../types/common/common'
+import { PageLayoutEditor } from './editor'
+
+export const pageLayoutState = object({
+  widthPercent: number(), // first column defined second
```
kulla: I would use an enum with 3 values here. Then I would add a function converting this enum into a percentage number.

kulla (changes requested):

src/edtr-io/plugins/page-layout/index.ts
elbotho: Why would you do that over using the value directly?

PR #1620, opened 2022-04-27
use new setEntity mutations

elbotho:
- can be tested locally.
- parts of the code are not nice. But right now we can't save on staging so I want this to be merged soon.


Entkenntnis (approved):
thanks! let's test this further in staging

PR #1626, opened 2022-05-02
page special content: plugin for special blocks on pages (team page)

### LINKED ISSUE #1644: Static pages: Add team page plugin

elbotho:
can be tested e.g. here
https://frontend-git-page-special-content-plugin-team-serlo.vercel.app/entity/repository/add-revision/21468


Entkenntnis:
@elbotho Can you check the preview link? It's not the team page.

elbotho:
@Entkenntnis yeah, the idea was to avoid confusion. Did not work it seems :D 
The plugin can be added on all static pages. 
It would replace the manually constructed team grid on the team page.

elbotho:
@Entkenntnis sorry, you probably need more context. 
We want to convert the static pages that still use the old editor soon so I created some plugins exclusively for pages (that are not as polished, because they are only used by a few team members).

Basically a column plugin and this team plugin are the only real blockers for now.

Entkenntnis:
ah, I found it - it's a new plugin that you can add

PR #1639, opened 2022-05-07
discussions: use generated types

elbotho:
note: to make this work for `GetCommentsQuery` I had to change the codegen config again.


Entkenntnis:
thank you! I will take a look at the config and the types

src/fetcher/use-comment-data.ts
```diff
@@ -0,0 +1,81 @@
...
+          }
+        }
+      }
+    }
+  }
+`
+
+export type GetCommentsNode = Extract<
+  GetCommentsQuery['uuid'],
+  { threads: any }
```
Entkenntnis, line 45: @elbotho using `Extract` in this way enables checking for existence of a certain property

src/fetcher/use-comment-data.ts
```diff
@@ -0,0 +1,81 @@
...
+  })
+
+  const { data, error } = resp
+
+  const uuid = data?.uuid
+
+  // eslint-disable-next-line no-console
+  if (error) console.log(error)
+
+  if (uuid && hasOwnPropertyTs(uuid, 'threads')) {
```
Entkenntnis, line 62: don't know why other type guards fail, but this version finally did the job

codegen.yml
```diff
@@ -20,8 +22,6 @@ generates:
       - 'introspection'
 config:
   gqlImport: graphql-request#gql
-  onlyOperationTypes: true
-  preResolveTypes: true
```
Entkenntnis, line 24: switching back to default config for these to options

Entkenntnis (approved):
Thank you for working on this! I have digged into some issues and found some solutions that work with default codegen config options.

I'd like to stick with `preResolveTypes: true` (the default) because types are a bit more readable

PR #1651, opened 2022-05-10
add image captions

### LINKED ISSUE #1642: Add image captions

elbotho:


Entkenntnis:
@elbotho added a fix and a small refactoring, so the code is working now.

Entkenntnis:
the gap in https://frontend-iz9lg88v1-serlo.vercel.app/108552 is probably a bit small imo?

Entkenntnis (approved):
code looks good. merging into staging would make this available for all users?

elbotho:
@Entkenntnis thanks for the fix.

> the gap in https://frontend-iz9lg88v1-serlo.vercel.app/108552 is probably a bit small imo?

Those are hacked captions using the old two column layout. See:
https://de.serlo.org/nachhaltigkeit/108552/nachhaltigkeit-von-verschiedenen-textilfasern 

> code looks good. merging into staging would make this available for all users?

yes, and because it's a edtr core feature without a config to deactivate it we can't really add this as an experiment I think.
So we need to test it in staging. Maybe we wait till after the next deploy?

Entkenntnis:
Wait after next deploy is a good idea, then we can do some tests and feedback on staging

PR #1654, opened 2022-05-11
fix save modal typing 

### LINKED ISSUE #1653: typing in save modal is slow
probably the edtr-io is rerendered on every keystroke? can we fix this from the frontend?

Entkenntnis:
using a intermediate state (instead of plugin state) solves the issue

working now


elbotho (approved):
Thanks! 👍
The refactoring of the save logic will be great if we don't need the legacy compability any more :)

PR #1661, opened 2022-05-12
use codegen to fetch edtr-io data

### LINKED ISSUE #1652: use codegen for fetch editor data query
next up for refactoring

Entkenntnis:
looks doable

a lot of little fixes due to outdated types, some testing wouldn't be bad (how much do we trust the types :) )


elbotho (approved):
Thank you, looks great!

PR #1663, opened 2022-05-12
1662 use codegen for unrevised revisions

### LINKED ISSUE #1662: use codegen for unrevised revisions
after refactoring this part, we can remove quite a lot of manual types

-> some subtypes are still used in certain places ...

Entkenntnis:
based on #1661 

:tada: removed all handwritten main query types


elbotho (approved):
Very nice, thanks!

PR #1664, opened 2022-05-12
make new `/discussions` the default

elbotho:
This also passed testing and can be deployed.


Entkenntnis (approved):
thank you, looks good

PR #1670, opened 2022-05-16
Navigation data frontend

### LINKED ISSUE #1640: Migration: Move navigation into frontend
The main side navigation (see picture) shall be stored in the config of the navigation: 
2022-05-09-082935_251x252_scrot

elbotho:
closes #1640

Might need some testing, but most of the menus seem to be in the right place.

There are some standardisations we could do in the future, but the aim of this PR is only to replicate the legacy behaviour.

### Demo:
https://frontend-git-navigation-data-frontend-serlo.vercel.app/serlo


Entkenntnis (approved):
Nice! So the navigation data is handled together with all other language-dependant data.

Haven't tested it much, I'm giving this to the test-group.

PR #1674, opened 2022-05-16
use endpoints to save/create pages and taxonomy terms

elbotho:
- `page: addRevision`
- `taxonomyTerm: create / setNameAndDescription`
- this code is really loking forward to being refactored when we don't need the legacy compability any more :)

### DIFF
--- a/src/components/pages/add-revision.tsx
+++ b/src/components/pages/add-revision.tsx
@@ -13,8 +13,10 @@ import { useAddPageRevision } from '@/helper/mutations/use-add-page-revision-mut
 import {
   AddPageRevisionMutationData,
   SetEntityMutationData,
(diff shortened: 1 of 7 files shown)


Entkenntnis (approved):
Thank you! The code looks good.

But as you have noted on slack, the only taxonomy type that is working is `4` - the other types (and other ids on langauge versions) fail with an 500 api error. (probably this: https://github.com/serlo/serlo.org-database-layer/issues/258)

Entkenntnis:
as this code is hidden behind a feature flag, merging into staging should be save (and wait for API fix)

elbotho:
@Entkenntnis https://github.com/serlo/serlo.org-database-layer/issues/258 is closed and deployed so we can test again.

PR #1683, opened 2022-05-19
Hotfix: Fix course saving

Entkenntnis:


Entkenntnis:
@elbotho If you want you can deploy this. For me, it would be also fine to deploy this on monday

PR #1690, opened 2022-05-20
update api, add sorting tax content 

### LINKED ISSUE #1595: Migration: Sorting children (entities + taxonomy terms) in taxonomy terms
## where

After click "Other tools > sort content":

2022-04-15-230614_469x404_scrot

Link `https://en.serlo-staging.dev/taxonomy/term/sort/entities/:taxonomyTermId`

## New behavior

In the overview next to entities also children taxonomy terms shall be able to be sorted. The best solution is that they are already grouped by type and the sorting is done in each group separately. 
2022-04-19-201951_688x625_scrot

## Old Design:

2022-04-15-230742_864x521_scrot

## mutation

See https://github.com/serlo/api.serlo.org/issues/582 with `taxonomyTermId = parentId of entities` and `childrenIds = new list of children` (can be the list of each group concatenated) 

## Blocker:

- [x] https://github.com/serlo/serlo.org-database-layer/issues/214
- [x] https://github.com/serlo/api.serlo.org/issues/582

elbotho:
- fixes #1595
(works quite nicely now🔥)
- api update, adapt to breaking changes


Entkenntnis (approved):
Thank you for the feature and the API update!

PR #1692, opened 2022-05-21
fix duplicate footer

### LINKED ISSUE #1682: Footer showing twice
Example:
- https://de.serlo.org/lerntipps/186900/lernen-mit-eselsbr%C3%BCcken
- https://frontend-git-navigation-data-frontend-serlo.vercel.app/community/neue-f%C3%A4cher-themen

I did not see it first, now it's on all pages.

@Entkenntnis I looked into it a bit already but have no clue what is happening. 
It's probably something in the last deploy?
Do you have an idea?

image

elbotho:
Okay… the problem seems to come from a nested Lazy component.
The complete comment-area and also the threads were lazy loaded and somehow this tripped up the render (probably SSR).

I think the problem is gone now, but we should keep an eye on [react-lazyload](https://github.com/twobin/react-lazyload). It's not well maintained any more and there are some open issues with SSR and also new react versions.

### DIFF
--- a/src/components/comments/comment-area.tsx
+++ b/src/components/comments/comment-area.tsx
@@ -4,7 +4,6 @@ import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons/faQuestionCi
 import { Thread as AuthThread } from '@serlo/authorization'
 import { Fragment, useState } from 'react'
 
-import { Lazy } from '../content/lazy'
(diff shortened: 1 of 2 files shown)


Entkenntnis (approved):
Thank you for the fix! If we encounter more issue, I def' agree to review the react-lazyload library and see if there are some alternatives

PR #1697, opened 2022-05-23
add helper to check for visible text

### LINKED ISSUE #1695: Hide solution strategy if content is only whitespace

Entkenntnis:


src/helper/check-content-is-empty.ts
```diff
@@ -0,0 +1,19 @@
+import { hasOwnPropertyTs } from './has-own-property-ts'
+import { FrontendContentNode } from '@/data-types'
+
+export function hasVisibleContent(content: FrontendContentNode[]): boolean {
```
elbotho: nice, love the helper. 
could you change the filename or the exported function name here so they match?

elbotho (changes requested):

src/helper/check-content-is-empty.ts
Entkenntnis: done

PR #1698, opened 2022-05-23
setEntity improvements

elbotho:


Entkenntnis (approved):
thanks, this will run every in sequence

PR #1719, opened 2022-06-04
use code style also for edtr-io

### LINKED ISSUE #1712: unify styling of inline code
edtr-io has a different styling than the frontend:

edtr-io:

grafik

frontend:

grafik

we can probably choose one and add it to the edtr-io styling.

Entkenntnis:


elbotho:
Thanks 🎉

PR #1720, opened 2022-06-07
Display containing folder in notification view for types Exercise, Solution, ... and other improvements. 

LarsTheGlidingSquirrel:
Notification display now shows more context when entities of type Exercise, ExerciseGroup, GroupedExercise and Solution are referenced. 
Example: [Peter](http://localhost:3000/user/169563/Peter) hat eine [Bearbeitung 247053](http://localhost:3000/247053) von [Teilaufgabe | Aufgaben zu Wachstums- und Zerfallsprozessen](http://localhost:3000/mathe/247044/247044) übernommen.
From: http://localhost:3000/event/history/247044

Some entities will show their ID

Future tasks: 
- Use AbstractUuid.title when it is available https://github.com/serlo/api.serlo.org/pull/653
- Change link in example above to: ... Teilaufgabe in [Aufgaben zu Wachstums- und Zerfallsprozessen](http://localhost:3000/mathe/26418/aufgaben-zu-wachstums-und-zerfallsprozessen) (based on feedback from Meli & Kathi)


Entkenntnis (commented):
Thank you Lars for your first PR in the frontend!

I have two small (internal) notes before merging this - in general we are using some new stuff here and there.

src/components/user/event.tsx
```diff
@@ -18,6 +18,7 @@ import {
   SetThreadStateNotificationEvent,
   SetUuidStateNotificationEvent,
   TaxonomyTerm,
+  TaxonomyTermType,
   AbstractUuid,
   Thread,
 } from '@serlo/api'
```
Entkenntnis: internal note: maybe using the new graphql codegen here is a good refactoring because the query is already quite complex and it could help with the types

src/components/user/event.tsx
```diff
@@ -279,38 +280,150 @@ export function Event({
...
+    return hasPath(['type'], object)
+  }
+
+  function hasTaxonomyTerms(
+    object: AbstractUuid & { __typename?: string }
+  ): object is AbstractUuid & { __typename?: string } & {
+    taxonomyTerms: { nodes: { name: string; type: TaxonomyTermType }[] }
+  } {
+    return hasPath(['taxonomyTerms', 'nodes'], object)
+  }
```
Entkenntnis: I think `hasOwnPropertyTs` is our newer approach to handling this

src/components/user/event.tsx
```diff
@@ -279,38 +280,150 @@ export function Event({
...
 
+  // Gets localized entity name (for example "Solution") and containing folder (for example "Exercises about topic X")
+  // Full example: "Solution | Exercises about topic X"
+  function getEntityStringIncludingFolder(
+    object: AbstractUuid & { __typename?: string }
+  ) {
+    const taxonomyTerms = getTaxonomyTerms(object)
+
+    if (taxonomyTerms.length == 0) {
+      return getEntityStringByTypename(object.__typename, strings)
```
elbotho: Since you use `getEntityStringByTypename` three times and in all cases you should probably do 
``` ts
const entityString =  getEntityStringByTypename(…)
```

elbotho (dismissed):
@LarsTheGlidingSquirrel 🎉 welcome the the frontend.
Thank you for the PR, noted one thing, rest looks very clean.

As @Entkenntnis noted, we improved some things in the meantime (we can now generate exact typescript types for our queries automagically) and also the `hasOwnPropertyTs` helper is quite handy. 

We (you or me, as you like) could work those in, but it might make more sense to just wait for https://github.com/serlo/api.serlo.org/pull/653 to be on staging since that will change the code again anyway. What do you think?

Entkenntnis:
@LarsTheGlidingSquirrel I have prepared the codegen-part in this PR: https://github.com/serlo/frontend/pull/1721

As botho suggested, you can move on from there (the main file is not refactored yet) or we can take this over :)

src/components/user/event.tsx
LarsTheGlidingSquirrel: Thanks! I kind of expected that there is a cleaner way to do these checks.

LarsTheGlidingSquirrel:
Thank you @elbotho and @Entkenntnis for the feedback and warm welcome! :)

I'll have a look into the graphQL codegen. Waiting for https://github.com/serlo/api.serlo.org/pull/653 seems to make sense. I'll be in the mountains the coming two weeks so there wont be any progress from my side but feel free to continue if you want. Otherwise I'll pick it up again once I'm back.

elbotho:
Sounds good!

elbotho:
this cleaned up a lot with entity titles 🎉

Entkenntnis (approved):
looks good and is working

PR #1725, opened 2022-06-08
api updates (license, alias)

elbotho:
- rename field to `isDefault` to avoid problems when destructuring. (`default` is protected)
- use new `shortTitle` field
- remove alias checks


elbotho:
@Entkenntnis I'll merge this now to make staging work again, but feel free to review and I'll create a new PR if needed.

PR #1732, opened 2022-06-09
header: variante a (unified header, search refactor)

elbotho:
- landing page style header everywhere with some compromises
- quickbar for search everywhere

**todos**
- [x] test more
- [x] get some feedback
- [x] restyle and use searchinput on search page again


Entkenntnis:
Hi @elbotho, 

the new header feels fresh :) and lighter -> makes content more prominent, which is a good thing, I think.

Here are some things I noticed:

1. the search box feels a bit too narrow, especially on desktop when there is more space - maybe allow quickbar to grow a bit more?

grafik

2. double search on weird styling on landing page:

grafik

3. love this, it's much more clearer:

grafik

4. search input is missing styling only directly after consent is giving: (If you reload, the styling is restored)

grafik

Entkenntnis:
Code looks good!

The next step is probably asking for some more feedback

elbotho:
Thanks @Entkenntnis for the thorough feedback!

1. I thought about allowing the overlay to grow bigger than the input and now added this. What do you think?
2. Landing: nice catch, not only one search bar (the lower one) is visible and styled like the others
3. ☺️
4. Puh, this resulted in quite some refactoring. There was a problem with hydration, but I chose to use googles default input now. It's not visible very often and simpifies the code even more

Entkenntnis:
Thanks!

1. This is quite fine for me.
2. :heavy_check_mark: 
3. :heavy_check_mark: 
4. Looks. I think this makes it clearer that this is an external google search.

Another mini issue: If you use the quickbar on a content page and navigate to another content with **enter**, the quickbar is still open -> maybe we can automatically close the quickbar after navigation to improve UX

Entkenntnis (approved):
Overall I think this is quite a nice refactoring

elbotho:
Thanks again!

> Another mini issue: If you use the quickbar on a content page and navigate to another content with enter, the quickbar is still open -> maybe we can automatically close the quickbar after navigation to improve UX

ah nice, saw this before but did not reproduce. it only happens with keyboard navigation. should be fixes with last commit.

elbotho:
@Entkenntnis one more thought: with this change we load the new font on every page load.
I don't think this is a big problem since it should not block loading and rendering, but we could also replace the logo with an svg.

Entkenntnis:
lighthouse performance score is 100. I mean, doesn't look like a problem :)

(we have some other issues, though)

elbotho:
follow up in #1759

PR #1737, opened 2022-06-11
new entity, admin taxonomy stuff, refactoring

### LINKED ISSUE #1735: Updates for subjects
After investigating https://github.com/serlo/api.serlo.org/issues/642 the following changes should be done:

- [ ] In root folders like https://serlo.org/3 instead of displaying `Add topic folder` the string in the menu shall be `Add subject`
- ~[ ] In the metadata title of subjects one can can change the logic which folder is a subject to: `taxonomy_term.type === "subject" || (taxonomy_term.type === "topic" && rootIds.includes(taxonomy_term.parent_id)`~

The root taxonomy ids can be hardcoded in the frontend since they do not change often (only when a new instance is added). They are:

```
mysql> select id from term_taxonomy where parent_id is null;
+--------+
| id     |
+--------+
|      3 |
|  23590 |
| 140404 |
| 140438 |
| 140439 |
| 141572 |
+--------+
```

elbotho:
- [x] refactor "new entity" logic
- [x] use new infrastructure to save tax elements
- [x] allow in all instances
- [x] updates api
- [x] rename `topicFolder` to `exerciseFolder`
- [x] filter now outdated `locale` types for subject landing pages
- [x] some testing
- ❇️ add instance to /discussions query

closes #1735


Entkenntnis (approved):
Thank you, a lot of refactorings here

PR #1739, opened 2022-06-11
make sure legacy is logged in before edit

### LINKED ISSUE #1729: improve session check in add-revision
this code is making a connection:

https://github.com/serlo/frontend/blob/ba635e9a0bc1a114b701be270543da43867f339d/src/components/pages/add-revision.tsx#L36

we should also check if the page is actually logged in (e.g. with regex)

elbotho:


elbotho:
merged to test

PR #1746, opened 2022-06-14
special error page for deleted comments

elbotho:


src/data/en/index.ts
```diff
@@ -247,6 +247,10 @@ export const instanceData = {
       refreshNow: 'Refresh now',
       backToPrevious: 'Back to previous page',
       backToHome: 'To our home page',
+      deletedComment: {
+        title: "Whoops, this is not here anymore",
+        text: "Sorry, this %type% is no longer online.%break% But it was deleted for a reason and was probably not worth you time anyway 💚"
```
inyono:
```suggestion
        text: "Sorry, this %type% is no longer online.%break% But it was deleted for a reason and was probably not worth your time anyway 💚"
```

Entkenntnis (approved):
I will merge this for now, but I'm not 100% if there could be some hydration mismatch for this because generated html could differ from client-side renderer html.

But it's quite an edge case, so happy to fix this later (if it becomes an issue)

PR #1750, opened 2022-06-19
update edtr

elbotho:
- small change in image captions (hide description field)
- use caption as alt if alt is empty
- alt fallback i18n


Entkenntnis (approved):
nice!

PR #1754, opened 2022-06-21
refactor newsletter popup, only on allowed pages

elbotho:
- cleaner approach
- let's see how `next/script` affects bundles
- there was a bug that the popup would open on pages where it should not open

===========
long term:
own implementation without exernal code would not be that hard.
Probably with https://www.npmjs.com/package/react-mailchimp-subscribe


Entkenntnis (approved):
bundle size looks good.

PR #1762, opened 2022-06-23
deploy: donation banner

elbotho:


elbotho:
@Entkenntnis could you glance over the code? It's temporary and hacky but should still be save and work :)

Entkenntnis (approved):
Works well as a temporary solution! As a performance critical part, we may have to optimize for bundle size, but the current solution is fine atm.

Feel free to merge into staging / production as you prefer.

PR #1763, opened 2022-06-24
Refactorings

elbotho:
- codegen types for everything (mutations excluded)
- `@serlo/api` dependency is removed (using codegen typed instead)
- explicit typing for `type`, `typename`… that used to be `string`
- add `check:all` script like in the other repos


elbotho:
@Entkenntnis I think I'm done here for now.
Merge or request changes if you like :)


PS: While this cleans up a lot and also fixes minor bugs: I'm sure it introduces some new ones as well. So having it on staging for testing might be a good idea.

Entkenntnis (changes requested):
one last blocker, rest looks very good:

prefer moving enums into separate files to improve tree-shaking and bundle size -> currently, importing enums is messing with bundling in some places, e.g. `ExerciseInlineType`

elbotho:
@Entkenntnis we are at +1kB  compared to current staging now.

``` diff
-    ├ ● /[...slug]                                            4.5 kB          146 kB
+    ├ ● /[...slug]                                            4.52 kB         147 kB
```

PR #1769, opened 2022-06-28
add iframeResizer.contentWindow.min.js

inyono:
To get `iframe-react-resizer` to work correctly (which is necessary for a good UX in the serlo injection plugin used for RLP), we unfortunately need to include this script.


elbotho:
> To get iframe-react-resizer to work correctly (which is necessary for a good UX in the serlo injection plugin used for RLP), we unfortunately need to include this script.

This only needed for editing, right? Then we should make sure we only load it when the editor is loaded.

inyono:
Nope. That's required when someone else (e.g. RLP Editor) is integrating our content via iFrame. So we need it on all pages that we want people to theoretically embed (so at least all UUIDs).

elbotho:
> Nope. That's required when someone else (e.g. RLP Editor) is integrating our content via iFrame. So we need it on all pages that we want people to theoretically embed (so at least all UUIDs).

Sorry for insisting but adding this amount of js is something we do not take lightly here ;)
So only requests to our content API need this script?  We have a special route for this where we could add it then.

inyono:
That would be awesome. As long as all content api requests have that script I'm fine with that. While we are at it: what's the current approach on how the frontend detects those?

elbotho:
> While we are at it: what's the current approach on how the frontend detects those?

1. CF-Worker detects search params [here](https://github.com/serlo/serlo.org-cloudflare-worker/blob/1e675a9a9e4e2d62f4ca3016795776a9ac5c8ef5/src/utils/url.ts#L58)
2. CF-Worker redirects to `'/content-only' + url.pathname` ([code](https://github.com/serlo/serlo.org-cloudflare-worker/blob/main/src/frontend-proxy.ts#L68))
3. Frontend has special route that handles those slighty different ([code](https://github.com/serlo/frontend/blob/staging/src/pages/content-only/%5B...slug%5D.tsx))

Entkenntnis (changes requested):
I'm fine with adding this script to contentOnly paths - and it would be nice if we only load the script after idleing, this has a lower impact on performance. A [Script](https://nextjs.org/docs/basic-features/script)-Tag should do the trick.

Is there any plan on https://github.com/edtr-io/edtr-io/issues/392?

Entkenntnis:
@inyono can you test it with e.g. https://frontend-git-iframeresizercontentwindowminjs-serlo.vercel.app/content-only/1795

inyono:
> @inyono can you test it with e.g. https://frontend-git-iframeresizercontentwindowminjs-serlo.vercel.app/content-only/1795

Works!

https://github.com/edtr-io/edtr-io/issues/392 is no longer relevant, because the underlying library behind `iframe-resizer-react` (namely `iframe-resizer`) is actively maintained (and I also didn't find a good replacement). It seems like `iframe-resizer-react` is just stable enough to not warrant frequent updates. I was just too dumb and didn't realise that they wanted me to include the script so that it works correctly.

Entkenntnis (approved):

PR #1775, opened 2022-07-04
experiments update

elbotho:
- `useAddRevisionMutation`: on by default
- `usePagePlugins`: on by default
- `legacyEditor`: remove (off by default)

this does not include a big refactor of the save logic, but is the first step :)


Entkenntnis (approved):
looks good!

PR #1779, opened 2022-07-05
feat: implement all Kratos flows

### LINKED ISSUE #1784: Implement user flows with Kratos
edit: this issue was changed to have a much bigger scope.

_note: we are doing gradual user management migration, that's why some flows have some extra specifications_

# Flows
## Login flow
Conditions:
- users were migrated to the user management system (kratos)
- users maintain a legacy_id

## Registration flow
1. Create user in user management system (kratos)
2. Call endpoint in api to add user in our table, which responds with the legacy table id
3. Update user in kratos with legacy_id

## Reset password flow (Settings flow)
Actually, it is not a full fledged settings, since it updates only the password. Having a settings that updates also description etc. would imply some synchronisation between kratos and the legacy user table, which is a complexification that we don't need for now.

## Verify flow 
 A verify email flow will not be necessary since we will force the user to verify email before being able to login. So, the verify flow implemented for now is just the verification of email after registration.

Also have login, consent and logout for OAuth2.

hugotiburtino:
# TODOs

- [x] https://github.com/serlo/frontend/issues/1886
- [x] https://github.com/serlo/frontend/issues/1887
- [x] #1903
- [x] #1902


src/pages/api/auth/logout-callback.ts
```diff
@@ -20,7 +20,7 @@ async function callback(req: NextApiRequest, res: NextApiResponse) {
   if (csrf !== req.cookies['auth-csrf']) return fail('CSRF validation failed')
 
   try {
-    const token = JSON.parse(req.cookies['auth-token']) as Token
```
hugotiburtino: Although I haven't changed anything in this file, it started to complain about 
```console
src/pages/api/auth/logout-callback.ts:23:30 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

23     const token = JSON.parse(req.cookies['auth-token']) as Token
                                ~~~~~~~~~~~~~~~~~~~~~~~~~
```
So I added the !, but I'm not sure what caused it and, if we could solve that without changing this file for now, it would be great.

src/pages/api/auth/logout.ts
```diff
@@ -18,7 +18,7 @@ async function logout(req: NextApiRequest, res: NextApiResponse) {
   let idToken: string
 
   try {
-    const token = JSON.parse(req.cookies['auth-token']) as { id_token: string }
```
hugotiburtino: See comment to [logout-callback.ts](https://github.com/serlo/frontend/pull/1779/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.local&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.production&file-filters%5B%5D=.sh&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&file-filters%5B%5D=.yaml&file-filters%5B%5D=.yml&show-deleted-files=true&show-viewed-files=true#r913552635)

src/pages/api/auth/refresh-token.ts
```diff
@@ -23,7 +23,7 @@ async function refreshToken(req: NextApiRequest, res: NextApiResponse) {
 
   try {
     const accessToken = oauth2ClientCredentials.createToken(
-      JSON.parse(req.cookies['auth-token']) as Token
```
hugotiburtino: See comment to [logout-callback.ts](https://github.com/serlo/frontend/pull/1779/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.local&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.production&file-filters%5B%5D=.sh&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&file-filters%5B%5D=.yaml&file-filters%5B%5D=.yml&show-deleted-files=true&show-viewed-files=true#r913552635)

src/pages/api/auth/logout-callback.ts
Entkenntnis: @hugotiburtino you have upgraded all dependencies of the frontend. This is a bigger maintanance task and should be a separate PR (except you need some upgraded for this to work)

The above error is very likely caused by upgrading next.js

package.json
```diff
@@ -28,6 +33,9 @@
     "tailwind:status": "node ./scripts/tailwind-status.js",
     "codegen": "graphql-codegen --config codegen.yml"
   },
+  "resolutions": {
+    "@types/react": "^17.0.0"
```
Entkenntnis: Is there a reason why you are using this? the frontend is expecting react 18

package.json
```diff
@@ -56,13 +64,17 @@
...
+    "@ory/kratos-client": "^0.10.1",
+    "@oryd/hydra-client": "^1.9.0-alpha.3",
     "@serlo/authorization": "^0.42.0",
     "@serlo/legacy-editor-to-editor": "Entkenntnis/tmp-serlo-legacy-editor-to-editor",
     "@serlo/markdown": "Entkenntnis/tmp-serlo-markdown",
     "@tippyjs/react": "^4.0.0",
     "algebra.js": "^0.2.0",
     "array-move": "^4.0.0",
     "autoprefixer": "^10.0.0",
+    "axios": "^0.27.2",
```
Entkenntnis: hint: don't use axios, use `fetch` directly in the frontend. next.js is polyfilling it automatically

docs: https://nextjs.org/blog/next-9-4#improved-built-in-fetch-support

src/pages/api/auth/logout-callback.ts
hugotiburtino: Thanks for your comment. That explain some things 😄.
I had to solve conficts in `yarn.lock`, so I deleted it and ran `yarn install`. 
But I try it other way next time.

package.json
hugotiburtino: That is old code, from the first time Jonas worked on it. I see now, there is no `resolutions` in the most recent `package.json`, so I'll remove this field. Thanks for spotting it

package.json
hugotiburtino: For now we need Axios just because we need `AxiosError` in some code. But I will put it as a TODO to refactor it so that we don't need Axios anymore.

src/components/pages/login.tsx
```diff
@@ -0,0 +1,105 @@
+import {
```
elbotho: Could we maybe put all the auth pages in a subfolder?

hugotiburtino: sure, as you prefer

src/pages/api/auth/logout-callback.ts
elbotho: I allowed myself to revert the lock changes so the [files tab](https://github.com/serlo/frontend/pull/1779/files) loads faster again :)

hugotiburtino: @elbotho Maybe we want to revert this line to 
`const token = JSON.parse(req.cookies['auth-token'])...` without the !, so that we somehow don't change the behavior related to hydra for now. I can do that later

src/components/pages/login.tsx
hugotiburtino: done

hugotiburtino: Sorry, now I noticed that I also moved the pages itself to an `auth/` folder. Did you mean also that?

hugotiburtino: It somehow created the bug that after a successful login you get redirected to /login (404).

hugotiburtino: > It somehow created the bug that after a successful login you get redirected to /login (404).

Reason: something like this https://github.com/serlo/frontend/blob/kratos/src/components/auth/flow.tsx#L144
Again this very abstract component...

But anyways, if you wanted the pages themselves (not only the components) in an auth folder I can leave them there and adjust the Flow component. Otherwise, I can revert it.

elbotho: thank you.
pages in folder is what I mean, yes.

elbotho: Thanks!

hugotiburtino:
moving this commentary to PR description in order to make it more visible

src/helper/urls/get-auth-urls.ts
```diff
@@ -0,0 +1,9 @@
+// helpers while we test kratos alongside the old version
+
+export function getPasswordChangeUrl() {
```
elbotho: @hugotiburtino does this need to be a function?
Could we use a const string as well?
(not important, just wondered)

hugotiburtino: Sure, I'll change it

hugotiburtino: @elbotho , do you prefer hard coding it (like in other [menu strings](https://github.com/serlo/frontend/blob/staging/src/data/de/index.ts#L415)?)?
edit: anyway it is not important, I'll do it hardcoded to maintain the convention.

PR #1786, opened 2022-07-13
fix forceCookieDomain in local environment

inyono:
Issue was that `COOKIE_DOMAIN` was `serlo-staging.dev` even if run on `localhost:3000`.


hugotiburtino (approved):

hugotiburtino:
Thanks a lot

PR #1787, opened 2022-07-13
Node v16, update yarn

inyono:
Vercel finally supports Node v16, so this updates the CI etc. to use that version. Also updates yarn because why not.


elbotho:
Thank you!

PR #1789, opened 2022-07-19
Add basic setup for integration tests with cucumber and selenium

hugotiburtino:
This is a basic implementation of Cucumber with Selenium. We could use other tool (like cypress or codeceptjs) instead of selenium.
The beaty of cucumber (files with extension `*.feature`) is that non-technical people can read it or maybe write a draft of it (at the end of the day though the devs know better how to define the steps). By the way, tests using cucumber do not have to be integration tests, it can be sometimes unit tests, it is developer's call.
See if that is something for the future.
Meanwhile I'm going to use this setup when I start refactoring and optmizing Kratos integration.


elbotho:
@Entkenntnis since you've done the most testing in the frontend I'll let you decide here.
My first instinct is to keep the e2e in a seperate Repo until we decide on a framework like you did with codecept.

hugotiburtino:
> My first instinct is to keep the e2e in a seperate Repo until we decide on a framework like you did with codecept.

Thanks for your feedback.

**Pros for another repo:**
- We could use another language (hello Python :wink:)
- People could work indepedently on it without breaking frontend's CI-CD pipeline (which we would later set up)
- It would make more sense in terms of separation of concerns, since such test involves all services, not only frontend

**Cons:**
- Unit and component testing will be impossible
- Maybe we do want to have constant GUI testing Action in frontend in each PR. It would make development slowlier but more secure...

Anyways, we could set up also a pipeline in `infrastructure-env-staging` in order to run this test after each push on `main`.

elbotho:
> Unit and component testing will be impossible

Sure, those already are or would be in the frontend.

> Maybe we do want to have constant GUI testing Action in frontend in each PR. It would make development slowlier but more secure...

As I said, I'm talking short term. Long term this might make sense.

Entkenntnis:
result from dev-meeting: tests are nice, especially for complex workflows.

the best place atm would be another repo. Feel free to use any technology you see fit.

PR #1794, opened 2022-07-25
Image upload test

### LINKED ISSUE #1777: migration: Upload of images
Use https://github.com/serlo/api.serlo.org/issues/690 in upload of images. Blocker:

- [x] https://github.com/serlo/api.serlo.org/issues/690

### LINKED ISSUE #1785: semantic image filenames
Currently we use image names like:
`https://assets.serlo.org/${uuidv4}.png`

But the CF-Worker makes sure this link would also work:
`https://assets.serlo.org/${uuidv4}/semantic-name.png`

(see: https://github.com/serlo/serlo.org-cloudflare-worker/blob/main/__tests__/assets.ts#L64)

This would be a small but easy benefit for users and SEO. I suggest we either use the current entity title or parts of the caption/alt text to create this filename.

elbotho:
ping @kulla: relevant code is here:
https://github.com/serlo/frontend/pull/1794/files#diff-a403694802f78c70b0e66a4e308e22860be3a6629101bab8ccb2874c482b3e8d


src/edtr-io/plugins/image.tsx
```diff
@@ -85,44 +93,102 @@ export function createUploadImageHandler(getCsrfToken: () => string) {
...
+        reader.onload = function (e: ProgressEvent) {
+          if (!e.target) return
+          const formData = new FormData()
+          formData.append('attachment[file]', file)
+          formData.append('type', 'file')
+
+          fetch(data.media.newUpload.uploadUrl, {
+            method: 'PUT',
+            headers: { 'Content-Type': 'image/jpeg' }, // TODO: fixed only for testing
+            body: formData,
```
kulla:
```suggestion
            body: file,
```

Directly add the file as content instead of using form data (Google doesn't support it)

src/edtr-io/plugins/image.tsx
```diff
@@ -85,44 +93,102 @@ export function createUploadImageHandler(getCsrfToken: () => string) {
...
+      const authenticationPayload = parseAuthCookie()
+
+      if (!authenticationPayload?.token) return
+
+      const gqlFetch = createAuthAwareGraphqlFetch({
+        current: authenticationPayload,
+      })
+      const args = JSON.stringify({
+        query: uploadUrlQuery,
+        variables: { mediaType: MediaType.ImagePng },
```
kulla: This must match the content type given later in https://github.com/serlo/frontend/blob/9d8c02a9ba5c2bb93726ac5625dc22b528040a4b/src/edtr-io/plugins/image.tsx#L123 -> This is the reason why currently we get an error.

kulla (changes requested):
I commented the lines why there is currently an error. I will add a commit fixing the errors so that you can upload `png` files. You need to add the logic that the mime type is set correctly in order to upload JPG / SVG / .... So you can work further on this PR...

src/edtr-io/plugins/image.tsx
elbotho: 🤦‍♂️

src/edtr-io/plugins/image.tsx
```diff
@@ -85,44 +110,92 @@ export function createUploadImageHandler(getCsrfToken: () => string) {
...
+      token: access_token as string,
+      refreshToken: () => Promise.resolve(),
+      clearToken: () => {
+        return
+      },
+    }
+  } catch {
+    return null
+  }
+}
```
elbotho: @inyono do you maybe have an idea how we could easily get rid of the duplication here?
The normal way is inside a hook but since we are quite far away from any component here I did copy it for the POC.

kulla:
~@elbotho https://github.com/serlo/frontend/pull/1794/commits/2646c47a88d7cf395888b8329c3d23dda4514706 contains the deduplication of `parseAuthCookie()`~ -> Sry, doesn't work like this...

elbotho:
@kulla I removed the duplication with an okay hack I think :)

src/edtr-io/plugins/image.tsx
```diff
@@ -85,44 +107,70 @@ export function createUploadImageHandler(getCsrfToken: () => string) {
...
+      )
+      if (!authenticationPayload?.token) return
+
+      const gqlFetch = createAuthAwareGraphqlFetch({
+        current: authenticationPayload,
+      })
+      const args = JSON.stringify({
+        query: uploadUrlQuery,
+        variables: {
+          mediaType: mimeTypesToMediaType[file.type as SupportedMimeType],
```
kulla, line 82: I have the feeling that we should check `file.type` instead of the file name extension -> that is more save...

kulla (approved):
LGTM so that this feature can go into testing. I have some ideas for refactoring `src/edtr-io/plugins/image.tsx` -> However this I want to do in a separate PR then

elbotho:
thanks!
Feel free to open a PR or create an issue with the refactorings you want.

PR #1796, opened 2022-07-27
Display id of objects in notifications that do not have a title.

LarsTheGlidingSquirrel:
Whenever an object does not have a title (for example 'Exercise') the ID is displayed next to it. To make it less visually distracting the ID is superscript. 

Example: 
image

Templates for notification text were changed slightly to reflect that change. But only in German and English. Other languages might still need to be changed slightly.


src/components/user/event.tsx
```diff
@@ -101,7 +101,8 @@ export function Event({
           thread: renderThread(event.thread),
           comment: (
             <Link href={`/${event.comment.id}`} forceNoCSR>
-              {`${strings.entities.comment} (${event.comment.id})`}
+              {strings.entities.comment}
```
elbotho: I'd give it a space in between the title and the id `{' '}`.

src/data/de/index.ts
```diff
@@ -314,21 +314,21 @@ export const instanceData = {
     events: {
       entityInParentPreposition: 'in',
       commentInParentPreposition: "zu",
-      setThreadStateArchived: "%actor% hat eine %thread% archiviert.",
-      setThreadStateUnarchived: "%actor% hat eine %thread% aus dem Archiv geholt.",
-      createComment: "%actor% hat einen %comment% in einer %thread% erstellt.",
-      createThread: "%actor% hat eine %thread% in %object% erstellt.",
+      setThreadStateArchived: "%actor% hat %thread% archiviert.",
```
elbotho: Only change the `en/index.ts` all other files are generated by [crowdin](https://crowdin.com/translate/serlo/76/en-de). Sorry to say but this will be overwritten and have to be done again in the tool.

elbotho: I added a warning in https://github.com/serlo/frontend/wiki/Internationalization.

src/components/user/event.tsx
elbotho: (same for the other `sub`s)

elbotho (changes requested):
Some comments, rest of the code looks good!

src/components/user/event.tsx
LarsTheGlidingSquirrel: I agree, that looks better!

src/data/de/index.ts
LarsTheGlidingSquirrel: Oh, seems like I missed that. I rolled back the changes to de/index.ts. 

I can also do the changes in Crowdin but I don't know if I have access.

src/components/user/event.tsx
```diff
@@ -101,7 +101,7 @@ export function Event({
           thread: renderThread(event.thread),
           comment: (
             <Link href={`/${event.comment.id}`} forceNoCSR>
-              {strings.entities.comment}
+              {`${strings.entities.comment} `}
```
elbotho: I believe [best practice](https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx) of adding a space between two jsx elements would be to
- use `{' '}`
- put both elements in one line seperated by space (if prettier allows, otherwise it adds option 1 automatically)
- in this case we could also use `&nbsp;` to make sure title and id are in the same line (non-breaking)

src/data/de/index.ts
elbotho: I think you can use your GitHub account to sign in?
If not ping me and I figure out the access :)

elbotho: (but this happens after we merge it to staging)

src/components/user/event.tsx
LarsTheGlidingSquirrel: ``&nbsp;`` seems very useful in this case!

src/components/user/event.tsx
```diff
@@ -256,14 +256,33 @@ export function Event({
...
   }: EventObject | EventParent) {
     return (
-      <Link href={alias}>{renderTitle(title, __typename as UuidType)}</Link>
+      <Link href={alias}>
+        <>
+          {renderTitle(title, __typename as UuidType, id)}
+          {shouldRenderParent(__typename as UuidType) ? (
+            <>{renderParent(title, __typename as UuidType)}</>
+          ) : (
+            <></>
```
elbotho: you can also just return `null` here. sorry, did not spot that earlier.

LarsTheGlidingSquirrel: Good to know! :)

elbotho (approved):
Thanks!

PR #1800, opened 2022-08-01
hide table, activate serloTable

elbotho:
use new `serloTable` plugin by default.

### backlog
- old one should still be supported until we have the mutations done. 
- also there is a bit of a problem with unconverted content. I don't think it's worth the effort to update the converter code to create the state for `serloTable` directly.

### DIFF
--- a/src/edtr-io/get-plugin-registry.tsx
+++ b/src/edtr-io/get-plugin-registry.tsx
@@ -17,7 +17,6 @@ import {
 import { faSquare } from '@fortawesome/free-regular-svg-icons/faSquare'
 import { faGripLinesVertical, faUsers } from '@fortawesome/free-solid-svg-icons'
 
-import { shouldUseFeature } from '@/components/user/profile-experimental'
 import { LoggedInData, UuidType } from '@/data-types'
 
 export function getPluginRegistry(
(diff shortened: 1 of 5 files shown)


src/edtr-io/get-plugin-registry.tsx
```diff
@@ -104,12 +103,6 @@ export function getPluginRegistry(
       description: editorStrings.edtrIo.spoilerDesc,
       icon: createIcon(faCaretSquareDown),
     },
-    {
-      name: 'table',
-      title: editorStrings.edtrIo.table,
-      description: editorStrings.edtrIo.tableDesc,
-      icon: createIcon(faTable),
-    },
```
inyono, line 112: If you want the old tables to still work for now, the plugin still has to be there. You could hide it, though, e.g. by removing/filtering it from the rows plugin registry.

inyono (changes requested):

src/edtr-io/get-plugin-registry.tsx
elbotho: This file creates the registry. 
`table` is still part of the [`plugins.tsx`](https://github.com/serlo/frontend/blob/hide-table-plugin/src/edtr-io/plugins.tsx#L368) and [should still work](https://frontend-git-hide-table-plugin-serlo.vercel.app/entity/repository/add-revision/206397).

inyono (approved):
LGTM

src/edtr-io/get-plugin-registry.tsx
inyono: I see, makes sense!

PR #1801, opened 2022-08-08
add bundle analyis ci

elbotho:


inyono (commented):
Looks good as far as I can tell. Is it "intended" that workflow doesn't run through (maybe because it needs a "first" run on production)?

elbotho:
> Looks good as far as I can tell. Is it "intended" that workflow doesn't run through (maybe because it needs a "first" run on production)?

yes, exactly. But I did not dare pushing a non functional workflow to prod so far :)

PR #1810, opened 2022-08-19
refactor(images): Deduplicate parseAuthCookie()

kulla:
@elbotho Can you check whether this fixes your problem with the duplicate code of `parseAuthCookie()`?


elbotho:
Sorry this does not seem to work:
- breaks upload
- also: https://reactjs.org/docs/hooks-rules.html

PR #1819, opened 2022-08-30
feat: Authenticate using Kratos

hugotiburtino:


src/auth/auth-provider.tsx
```diff
@@ -61,78 +60,40 @@ export function AuthProvider({
...
+  }, [])
 
-  function clearToken() {
-    if (!loggedIn) return
-    Cookies.remove('auth-token')
-    setLoggedIn(false)
-  }
+  const [loggedIn, setLoggedIn] = useState(() => {
+    return session !== null
+  })
```
inyono: I'd like to change that up so the information whether the user is logged in is synchronous. Currently, it's always false until kratos reports info about the session. Which isn't necessary, since we already know that the user is most likely logged in if the cookie exists (if the session is invalid, we might need to clean that cookie manually if kratos doesn't do that for us, but yeah). Similarly: we could think about storing the session information locally, too, so that we don't need to wait for the response. And only check if the session is still valid or something.

src/mutations/helper/use-mutation-fetch.ts
```diff
@@ -78,9 +72,10 @@ export function useMutationFetch() {
       const client = new GraphQLClient(endpoint, {
         headers: auth.current
           ? {
-              Authorization: `Bearer ${auth.current.token}`,
+              Authorization: `Bearer `,
```
inyono: This can be removed now, right?

src/pages/___graphql.tsx
```diff
@@ -51,7 +39,7 @@ function GraphQL() {
           'Content-Type': 'application/json',
           ...(auth.current
             ? {
-                Authorization: `Bearer ${auth.current.token}`,
+                Authorization: `Bearer `,
```
inyono: Same here

inyono (changes requested):

src/auth/auth-provider.tsx
```diff
@@ -61,78 +60,40 @@ export function AuthProvider({
...
-    }
+function useAuthentication(): [RefObject<AuthenticationPayload>, boolean] {
+  const [session, setSession] = useState<Session | null>(null)
+  const authenticationPayload = useRef<AuthenticationPayload>(null)
 
-    return pendingRefreshTokenPromise.current
-  }
+  useEffect(() => {
+    void (async () => {
+      const { data } = await kratos.toSession().catch()
```
inyono: The `.catch()` looks weird and might really catch all potential errors without reporting them.

src/auth/auth-provider.tsx
inyono: Okay, one slight problem: we can't check the cookie directly since it's httpOnly. But what we can do is the following:

- Persist the response of `toSession` to a cookie so that it's existent on first render.
- Use that cookie as initialValue to detect the user as logged in "optimistically".
- Still do the `toSession` to check if the session is still valid. If it's not (401 status code), then remove the cookie and report the user as not logged in anymore.

inyono: And bonus points if we also set that session cookie directly after log in so that we never have a wrong `null` in the beginning (which would lead to flashing content).

src/mutations/helper/use-mutation-fetch.ts
hugotiburtino: ~Yes, we just have to synchronize with cloudflare (and api) first.~
yes, I've taken a look at cloudflare and we can indeed remove, just have to adjust the api now (and later delete some dead code in cloudflare worder)

src/pages/___graphql.tsx
hugotiburtino: see https://github.com/serlo/frontend/pull/1819#discussion_r959374594

src/auth/auth-provider.tsx
hugotiburtino: Take a look please at https://github.com/serlo/frontend/pull/1819/commits/cbee43211ef4fb5812e603cb133670969edeb8f9 and see if you mean a specific error handling.

hugotiburtino: Sorry, actually, it is not good as implemented in https://github.com/serlo/frontend/commit/cbee43211ef4fb5812e603cb133670969edeb8f9 because it keeps logging into the console `GEThttp://localhost:3000/api/.ory/sessions/whoami [HTTP/1.1 401 Unauthorized 700ms]`.
We have to have the catch after toSession(), but I got your idea.

src/auth/auth-provider.tsx
hugotiburtino: Please take a look at https://github.com/serlo/frontend/pull/1819/commits/bb16c8a431f77883d039f4e7ac8b8582aec46142 and see if that is what you mean

inyono: Added two comments to https://github.com/serlo/frontend/commit/bb16c8a431f77883d039f4e7ac8b8582aec46142 which aren't shown in the PR directly for some reason.

hugotiburtino: Alright, I answered your comments there.

If you think it is overall good, I will just change it instead of persisting the whole response of the toSession, I will store a simple value like '1' or 'true' (and of course refactor it). The whole response is too much info exposed (like user email) that would be available to malicious sites.

inyono: Malicious sites don't have access to the cookie, though? But yeah, you can remove many parts, but you probably need the legacy id, right?

hugotiburtino: Sorry,  you are right, I was afraid of the possibility of doing it inside an iframe, until I read it is indeed not possible (it would be a quite obvious security flaw 😆 ).

But anyway, I get now I understood you wrong at the first time and you mean with initialValue also the data about the user (not only that she is logged in).

hugotiburtino: done (I hope) in [22fc88c](https://github.com/serlo/frontend/pull/1819/commits/22fc88c88275cd152f85ce5e24a6b9d434671149)

elbotho (approved):
nice simplification 👍
let's see how it works in the branch :)

PR #1820, opened 2022-08-31
Add "Not-Equal-To" sign to equations plugin

kulla:
Fixes https://serlo.slack.com/archives/C0BMSC431/p1661763605954029

### DIFF
--- a/src/components/content/equations.tsx
+++ b/src/components/content/equations.tsx
@@ -1,6 +1,7 @@
 import clsx from 'clsx'
 import { ReactNode, Fragment } from 'react'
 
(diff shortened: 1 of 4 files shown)


inyono (dismissed):

elbotho:
@kulla this is fine for the edtr, but I think this would not render in the frontend atm.

See: https://github.com/serlo/frontend/blob/staging/src/components/content/equations.tsx

elbotho (changes requested):
see comment.

elbotho:
@kulla ping me if I should take over

kulla:
@elbotho Ready to be merged (and you can take over)

PR #1822, opened 2022-09-06
refactor: Separate Hydra and Kratos authentications

hugotiburtino:


inyono (commented):
Didn't find anything in the code that tripped me up. If it seems to work correctly, this is good for testing from my side.

PR #1827, opened 2022-09-08
Refactor top menu and user tools

elbotho:
Complete rewrite of the header menu component using a primitive from RadixUI ([Navigation Menu](https://www.radix-ui.com/docs/primitives/components/navigation-menu)).

- split up in several components
- some simplifications
- combines both mobile and normal menu into the same components 🎉
- way better a11y and keyboard navigation

~This is the preparation for the upcoming menu changes.~
This includes the more complex dropdown for the community section in the menu.
Design might change.


src/components/navigation/header/header.tsx
```diff
@@ -1,23 +1,26 @@
...
+  const { strings } = useInstanceData()
   const router = useRouter()
 
   const hideQuickbar = router.route === '/search' || router.route === '/'
 
+  useEffect(() => {
+    document.body.addEventListener('keydown', (event) => {
+      if (event.key === 'Escape') setMobileMenuOpen(false)
+    })
+  })
```
inyono:
```suggestion
  useEffect(() => {
    document.body.addEventListener('keydown', (event) => {
      if (event.key === 'Escape') setMobileMenuOpen(false)
    })
  }, [])
```
Otherwise, you are adding a new event listener on each rerendering.

src/components/navigation/header/header.tsx
```diff
@@ -1,23 +1,26 @@
...
+
   // close mobile menu on client side navigation, we need the global Router instance
   Router.events.on('routeChangeStart', () => {
```
inyono: Unrelated to the PR, but I'm pretty sure this should be in a `useEffect`, too.

src/components/navigation/header/menu/item.tsx
```diff
@@ -0,0 +1,89 @@
...
+  'text-brand block border-b border-brand-lighter p-4',
+  'md:serlo-menu-entry-special',
+  'md:block md:serlo-button-blue-transparent md:text-[0.9rem] md:leading-tight md:transition md:text-brand-light',
+  'md:my-0 md:mt-[2px] md:py-0.5 md:px-[7px]',
+  'md:text-center',
+  'hover:no-underline hover:bg-brand-300'
+)
+
+export const preventHover = (event: any) => {
+  const e = event as Event
```
inyono:
```suggestion
export const preventHover: PointerEventHandler = (event) => {
```

src/components/navigation/header/header.tsx
elbotho: Uh, nice catch.

src/helper/colors.ts
```diff
@@ -1,6 +1,7 @@
 export const colors = {
   brand: '#007ec1',
   brandGreen: '#95bc1a',
+  brandGreen300: '#B9D957',
```
kulla, line 4: What does `brandGreen300` mean? Is there a better name?!

tailwind.config.js
```diff
@@ -1,4 +1,3 @@
-const { tint, lighten } = require('polished')
```
kulla, line 1: Have you removed `polished` as a dependency? I guess it is not used somewhere else now, right?

src/components/navigation/header/menu/icon.tsx
```diff
@@ -0,0 +1,51 @@
...
+          ? 'md:w-auto md:h-auto md:inline-block md:mr-0 md:bg-transparent'
+          : 'md:hidden'
+      )}
+    >
+      {element ? (
+        <>
+          {element} <span className="sr-only">{alt}</span>
+        </>
+      ) : icon ? (
+        <FaIcon icon={menuIconMapping[icon]} style={{ fontSize: '23px' }} />
```
kulla: is `alt` here missing?!

src/components/navigation/header/menu/icon.tsx
```diff
@@ -0,0 +1,51 @@
...
+  community: faUserFriends,
+  donate: faHandHoldingHeart,
+  user: faUser,
+  login: faUser,
+  notifications: faBell,
+}
+
+export type IconIdentifier = keyof typeof menuIconMapping
+
+export interface IconProps {
```
kulla, line 25: Am I interpreting the type right, that either `icon` or `element` need to be set? How about:

```typescript
type IconProps = { icon: IconIdentifier, alt: string } | { element: JSX.Element, alt: string }
```

Shall `element` be named `children` so that the following is possible:

```typescript
<Icon alt="...">
   <Element></Element>
</Icon>
```

src/components/navigation/header/menu/item.tsx
```diff
@@ -0,0 +1,89 @@
...
+  specialContent?: JSX.Element
+}
+
+export function Item({ link, specialContent }: ItemProps) {
+  const hasChildren = link.children !== undefined
+
+  const textAndIcon = (
+    <>
+      <Icon icon={link.icon} element={specialContent} alt={link.title} />
+      <span className={clsx(specialContent && 'md:sr-only')}>{link.title}</span>
```
kulla:
```suggestion
      <span className={clsx(specialContent ?? 'md:sr-only')}>{link.title}</span>
```

kulla: See https://mariusschulz.com/blog/nullish-coalescing-the-operator-in-typescript

src/components/navigation/header/menu/menu.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { AuthItems } from './auth-items'
+import { Item } from './item'
+import { useInstanceData } from '@/contexts/instance-context'
+
+export function Menu() {
+  const { headerData } = useInstanceData()
+  const [mounted, setMounted] = useState(false)
+
+  useEffect(() => setMounted(true), [])
```
kulla, line 12: Why do you have this construction?

src/components/navigation/header/menu/sub-participate-mega.tsx
```diff
@@ -0,0 +1,60 @@
...
+    </div>
+  )
+
+  function renderNewArea() {
+    return (
+      <>
+        <CommunityBird className="hidden md:block w-[8.1rem] px-side" />
+        <NextLink href="/community" passHref>
+          <NavigationMenu.Link className="ml-4 my-3 serlo-button-green text-base rounded-4xl w-max py-[0.1rem]">
+            <span>Hier gehts los!</span>
```
kulla, line 54: i18n needed?

kulla (approved):
LGTM. I haven't seen big issues. See comments for some improvments

src/helper/colors.ts
elbotho: It's a tailwind convention I adopted more of in this PR.

elbotho: See https://tailwindcss.com/docs/customizing-colors

src/components/navigation/header/menu/icon.tsx
elbotho: we have [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) on the icons.

src/components/navigation/header/menu/sub-participate-mega.tsx
elbotho: Nope, `de` only.

src/components/navigation/header/menu/menu.tsx
elbotho: We render in two passes.
First there is SSR that assumes no authed user.
We hide the "user" part of the menu until we run this component on the client and have the user data available.

src/components/navigation/header/menu/item.tsx
elbotho: What the current version wants to archive: Add `md:sr-only` if `specialContent` is defined. 
The safest way would be `specialContent ? 'md:sr-only' : ''` which we actually use in JSX now to avoid accidentally rendering the left hand side. But I think inside `clsx` this  shorthands is safe enough.

src/components/navigation/header/menu/icon.tsx
elbotho: I refactored the icon comp a little bit. Should be cleaner now.

src/components/navigation/header/menu/icon.tsx
elbotho: (see above).

kulla (approved):

PR #1831, opened 2022-09-12
Lenabi mocked

elbotho:


src/api/endpoint.ts
```diff
@@ -1,6 +1,12 @@
...
-export const endpoint =
+export const endpointHost =
   process.env.NEXT_PUBLIC_ENV === 'local'
-    ? 'http://localhost:3001/graphql'
-    : `https://api.${serloDomain}/graphql`
+    ? 'http://localhost:3001'
+    : `https://api.${serloDomain}`
+
+export const endpointGraphql = `${endpointHost}/graphql`
+export const endpointEnmeshed = `${endpointHost}/enmeshed`
```
kulla, line 9: we have removed this -> so this variable shouldn't be in the code any more...

src/auth/auth-provider.tsx
```diff
@@ -1,5 +1,5 @@
 import { AuthorizationPayload } from '@serlo/authorization'
-import Cookies from 'js-cookie'
+// import Cookies from 'js-cookie'
```
kulla, line 2: not needed source code should always be deleted -> we have git for this...

kulla (commented):
Thx! (No approval so that nobody accidentially merges this PR)

PR #1847, opened 2022-09-26
graphql localhost proxy

elbotho:


src/pages/api/frontend/localhost-graphql-fetch.ts
```diff
@@ -0,0 +1,31 @@
...
+  const { query, variables } = JSON.parse(req.body as string) as ParsedArgs
+
+  if (!req.headers.cookie)
+    res.status(403).json({ message: 'No auth cookie provided!' })
+
+  function executeQuery() {
+    const client = new GraphQLClient(endpoint, {
+      credentials: 'include',
+      headers: { Cookie: req.headers.cookie! },
+    })
```
hugotiburtino, line 20: Very good. I think, we will need to take a look at it again when it is in kratos stuff in api is deployed in staging. I'm uneasy with the fact, that we send only two headers, but indeed, testing locally, that worked while `headers: req.headers` haven't.

hugotiburtino (approved):

src/pages/api/frontend/localhost-graphql-fetch.ts
elbotho: what headers might be missing compared to the client request?

PR #1849, opened 2022-09-28
donations banner round 2️⃣

elbotho:


kulla (changes requested):
Some (not really big) changes in the commits -> I tried to get rid of the hack but did not succeed (will try later). However I noticed that the banner is rendered under each grouped exercise. Is this intended?!

elbotho:
> the banner is rendered under each grouped exercise.

Do you have a link? Have not seen this.

kulla:
> Do you have a link? Have not seen this.

http://localhost:3000/mathe/26408/gemischte-aufgaben-zum-ableiten-von-funktionen: 
2022-09-29-144743_510x739_scrot

kulla:
The banner is then also visible when the solution is viewed -> I don't know whether this is an issue...

src/components/content/donations-banner-experiment/donations-banner-inline.tsx
```diff
@@ -0,0 +1,194 @@
...
+              if (
+                entry &&
+                entry.isIntersecting &&
+                globalThis.hack__lastId !== globalThis.hack__id
+              ) {
+                globalThis.hack__lastId = globalThis.hack__id
+                submitEvent(`spenden-seen-${banner.id}`)
+              }
+            },
+            { threshold: 0.8 }
```
kulla, line 101: BTW: Here you can play with the threshold (before an event was send when the first pixel of the banner was visible). Know you get an event, every time the threshold is crossed (more than 80% of the banner becomes visible / less than 80% becomes visible)

elbotho:
> http://localhost:3000/mathe/26408/gemischte-aufgaben-zum-ableiten-von-funktionen:

Thanks!

PR #1850, opened 2022-09-30
feature: experimental paste plugin for raw edtr.io JSON state

elbotho:


src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const store = useScopedStore()
+
+  function replaceWithStateString() {
+    console.log('starting replace')
+    const newDocuments: DocumentState[] = []
+
+    const content = JSON.parse(dummyContent) as {
+      plugin: 'rows'
+      state: DocumentState[]
+    }
```
kulla: Here you can use io-ts

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,61 @@
...
+    // get parent rows plugin:
+    //const parent = getParent(props.id)(store.getState())
+
+    store.dispatch(
+      replace({
+        id: props.id,
+        plugin: 'rows',
+        state: newDocuments,
+      })
+    )
```
kulla: Here you create a nested row -> we need an implementation where we replace this plugin with multiple other plugins...

elbotho: Yeah, that's one of the questions I asked jonas and then you. Any ideas?

elbotho: https://serlo.slack.com/archives/D0H3UB1E2/p1666600535956429

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const store = useScopedStore()
+
+  function replaceWithStateString() {
+    console.log('starting replace')
+
+    // TODO: Use io-ts here
+    const content = JSON.parse(dummyContent) as {
+      plugin: 'rows'
+      state: Document[]
+    }
```
kulla: Which types do we want to accept as input?! Examples:

* `type State = { plugin: string, state: unknown }`
* `{ plugin: "rows", state: State[] }`
* `State[]`

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const parentPlugin = getParent(props.id)(store.getState())
```
kulla: TODO: Check that parent is rows plugin...

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+      store.dispatch(
+        insertChildBefore({
+          parent: parentPlugin.id,
+          sibling: props.id,
+          document,
+        })
+      )
+    }
+
+    store.dispatch(removeChild({ parent: parentPlugin.id, child: props.id }))
```
kulla: Seems to work. I am fine with deploying it, when it is tested in staging very well. Can we only allow the Paste-Plugin for certain users? (e.g. Anne, Claudia, ...)

kulla (changes requested):

src/edtr-io/plugins/paste-hack/editor.tsx
kulla: Here is my idea, which now seems to work: https://github.com/serlo/frontend/pull/1850/commits/a6543689ea29c800008658e27daa8a943a075478

src/edtr-io/plugins/paste-hack/editor.tsx
kulla: @elbotho Can you test what happens, when a nested plugin is inserted (e.g. there is a box in the inserted data)

elbotho: > @elbotho Can you test what happens, when a nested plugin is inserted (e.g. there is a box in the inserted data)

Yeah, good idea.

elbotho: > Can we only allow the Paste-Plugin for certain users? (e.g. Anne, Claudia, ...)

I would put the plugin into an experimental feature anyway (this way only power-authors will find it) and we could also only show this experiment for selected roles or users.

src/edtr-io/plugins/paste-hack/editor.tsx
elbotho: I tried to make the type more explicit in the io-ts decoder. Basically `{ plugin: "rows", state: State[] }` with a check if the rows child plugins are supported.

src/edtr-io/plugins/paste-hack/editor.tsx
elbotho: It's now behind a feature flag and also forbidden in production for now.

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+        t.literal('video'),
+        t.literal('table'),
+        t.literal('serloTable'),
+        t.literal('highlight'),
+        t.literal('injection'),
+        t.literal('layout'),
+        t.literal('multimedia'),
+        t.literal('spoiler'),
+        t.literal('important'),
+        t.literal('blockquote'),
```
kulla: We do not support Blockquote any more, right?

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+  plugin: t.literal('rows'),
+  state: t.array(
+    t.strict({
+      plugin: t.union([
+        t.literal('article'),
+        t.literal('articleIntroduction'),
+        t.literal('geogebra'),
+        t.literal('anchor'),
+        t.literal('video'),
+        t.literal('table'),
```
kulla: Same here

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const store = useScopedStore()
+  const textareaRef = useRef<HTMLTextAreaElement>(null)
+
+  function throwError(error?: unknown) {
+    showToastNotice('⚠️ Sorry, something is wrong with the data.', 'warning')
+    // eslint-disable-next-line no-console
+    console.error(error)
+    throw new Error(
+      'JSON input data is not a valid edtr-state or contains unsupported plugins'
+    )
```
kulla, line 54:
```suggestion
    throw new Error(
      'JSON input data is not a valid edtr-state or contains unsupported plugins', { cause: error }
    )
```

Can you check whether it works?

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+export const PasteHackEditor: React.FunctionComponent<PasteHackPluginProps> = (
+  props
+) => {
+  const store = useScopedStore()
+  const textareaRef = useRef<HTMLTextAreaElement>(null)
+
+  function throwError(error?: unknown) {
+    showToastNotice('⚠️ Sorry, something is wrong with the data.', 'warning')
+    // eslint-disable-next-line no-console
+    console.error(error)
```
kulla, line 51:
```suggestion
```

Maybe not needed when you can add the cause to the error...

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+    throw new Error(
+      'JSON input data is not a valid edtr-state or contains unsupported plugins'
+    )
+  }
+
+  function replaceWithStateString() {
+    if (!textareaRef.current || !textareaRef.current.value) return
+    const decoded = StateDecoder.decode(JSON.parse(textareaRef.current.value))
```
kulla: try-catch needed when content is not JSON parsable...

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const content = decoded.right
+
+    try {
+      const parentPlugin = getParent(props.id)(store.getState())
+
+      if (
+        parentPlugin == null ||
+        serializeDocument(parentPlugin.id)(store.getState())?.plugin !== 'rows'
+      ) {
+        const msg = 'only use on rows plugin!'
```
kulla:
```suggestion
        const msg = 'Paste plugin should only be used inside a rows plugin!'
```

src/edtr-io/plugins/paste-hack/editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+        showToastNotice(msg)
+        // eslint-disable-next-line no-console
+        console.error(msg)
+        return
+      }
+
+      const documents = content.plugin === 'rows' ? content.state : [content]
+
+      for (const document of documents) {
```
kulla:
```suggestion
      for (const document of content.state) {
```

kulla (changes requested):
@elbotho I think it can go to testing in staging afterwards...

src/edtr-io/plugins/paste-hack/editor.tsx
elbotho: in general we still need to support it, but in this case it's probably a good idea to remove 👍

src/edtr-io/plugins/paste-hack/editor.tsx
elbotho: Can't make it work.
Try e.g. `throw new Error('the error', {cause: "test string"})` in browser console.

PR #1873, opened 2022-10-26
spoiler: small refactor

elbotho:
the current implementation displays the spoiler content when using ssr (for SEO) but caused a display bug.

this pr:
- refactor: simplify
- design: rounded corners
- hide on ssr (seo downsides are probably small) 
- add simple animation

Note: 
One person reported a problem where the arrow and title don't stay in the same line when opening the spoiler but I could not reproduce and I suspect a cache issue.
Since the current implementation triggers hydration errors I want to merge this.

Examples:
https://frontend-git-spoiler-tiny-refactor-serlo.vercel.app/90412
https://frontend-git-spoiler-tiny-refactor-serlo.vercel.app/1565


kulla (approved):

Entkenntnis:
I will look into this

Entkenntnis:
nice animation - and very simple to add.

would be interested what the typescript problem was (upgrading is fine)

Entkenntnis (commented):
👍

elbotho:
> would be interested what the typescript problem was (upgrading is fine)

when upgrading to the current typescript version before I ran into problems with eslint.
Upgrading those as well now did the trick:
```
"@typescript-eslint/eslint-plugin": "^5.48.0",
"@typescript-eslint/parser": "^5.48.0",
```

Not sure if I forgot that or if they just took some time to upgrade typscript-eslint.

(so the typescript upgrade is not part of the spoiler change)

PR #1877, opened 2022-11-04
kratos: resolve merge conflicts against staging

hugotiburtino:
Unfortunately it is difficult to track the changes. 
I've tested manually regarding user management and it works.


elbotho (approved):
Let's try :)

PR #1878, opened 2022-11-04
kratos: Implement UI improvements as requested by testers

hugotiburtino:
- [x] Registration: tell the user that by registrering they consent to site rules
- [x] add friendly message if username "does not match pattern \"^[\\\\w\\\\-]+$\""  
- [x] In verification flow do not show email field if email already verified (or just redirect)
- [x] remove /auth/login-check page
- [x] add possibility of i18n at the [flow node level](https://github.com/serlo/frontend/blob/kratos/src/components/auth/node.tsx#L103) 
- [x] put link to verification page in case of  "Account not active yet. Did you already  verify your email address? "
- [x] add some instructions at verfirification, password change, login and registration


src/components/auth/flow.tsx
```diff
@@ -137,7 +137,17 @@ export function Flow<T extends SubmitPayload>({
                 : text
 
               // TODO: check context
-              const message = replacePlaceholders(rawMessage, { reason: text })
+              const message = replacePlaceholders(rawMessage, {
+                reason: text,
+                verificationLink: (
+                  <a
```
elbotho: Do you need a page reload here? Otherwise using the `Link` component would make sense as suggested by eslint.

src/components/auth/flow.tsx
```diff
@@ -207,7 +217,7 @@ export function handleFlowError<S>(
         return
       case 'session_already_available':
         // User is already signed in, let's redirect them home!
-        await router.push('/auth/login-check')
+        await router.push('/')
```
elbotho: Bonus points:
- add a toast message like: "Welcome back – you are already logged in"
- maybe try redirecting to last page first

src/components/auth/node.tsx
```diff
@@ -26,6 +27,34 @@ export function Node(props: NodeProps) {
 
   if (isUiNodeInputAttributes(attributes)) {
     const shortName = attributes.name.replace('traits.', '')
+    const fieldName = hasOwnPropertyTs(strings.auth.fields, shortName)
+      ? (strings.auth.fields[shortName] as string)
+      : shortName
+
+    const messages = node.messages.map(({ text, id }, k) => {
+      const hasTranslatedMessage = hasOwnPropertyTs(strings.auth.messages, id)
+      const rawMessage = hasTranslatedMessage
```
elbotho: maybe we sohuld turn this into a helper / component since we use it in `node` and `flow`.

src/components/auth/node.tsx
```diff
@@ -26,6 +27,34 @@ export function Node(props: NodeProps) {
...
+        reason: text.replace(
+          'does not match pattern "^[\\\\w\\\\-]+$"',
+          strings.auth.usernameRules
+        ),
+        field: fieldName,
+      })
+
+      return (
+        <span
+          key={`${id}-${k}`}
```
elbotho: I think `k` would only make sense if a message would come up twice on the same comp.
Probably safe to remove it.

src/data/en/index.ts
```diff
@@ -354,18 +358,23 @@ export const instanceData = {
...
-        1080002: 'You successfully verified your email address.',
+        1080002: emailVerifiedSuccessfully,
         4000001: '%reason%',
-        4000002: 'Property %field% is missing.',
+        4000002: '%field% is missing.',
         4000003: 'Length must be \u003e= %expected_length%, but got %actual_length%.',
         4000004: '"%value%" is not valid "%format%"',
         4000005: 'The password can not be used because %reason%.',
-        4000006: 'The provided credentials are invalid, check for spelling mistakes in your password or username/email.',
+        4000006: 'The username, email adress or password was wrong. Please check for spelling mistakes.',
```
elbotho: Personally I'd prefere "not correct", "incorrect" instead of "wrong".

src/data/en/index.ts
```diff
@@ -354,18 +358,23 @@ export const instanceData = {
...
-        4000010: 'Account not active yet. Did you already verify your email address?',
+        4000010: 'Have you already verified your email address?. %verificationLink%',
         4060004: 'The recovery token is invalid or has already been used. Please retry the flow.',
-      }
+      },
+      usernameRules: 'Your username may only contain letters, digits, underscores (_) and hyphens (-).',
+      registrationAgreement: 'By clicking %signup%, you agree to our %privacypolicy% and %terms%. You may receive email notifications from us and can opt out at any time.',
+      terms: 'Terms',
+      signUp: 'Sign up',
+      verificationLink: 'Go here to get the verification email again.'
```
elbotho: Maybe "Click to send verification email again"?

src/components/auth/node.tsx
```diff
@@ -26,6 +27,34 @@ export function Node(props: NodeProps) {
...
+      : shortName
+
+    const messages = node.messages.map(({ text, id }, k) => {
+      const hasTranslatedMessage = hasOwnPropertyTs(strings.auth.messages, id)
+      const rawMessage = hasTranslatedMessage
+        ? strings.auth.messages[id as keyof typeof strings.auth.messages]
+        : text
+
+      const message = replacePlaceholders(rawMessage, {
+        // couldn't we find more elegant way to handle %reason% placeholder?
```
elbotho: yeah we should, but I think it's an issue with kratos, right?

src/data/en/index.ts
```diff
@@ -354,18 +358,23 @@ export const instanceData = {
         1060001: 'You successfully recovered your account. Please change your password in the next minutes.',
         1060002: 'An email containing a recovery link has been sent to the email address you provided.',
         1080001: 'An email containing a verification link has been sent to the email address you provided.',
-        1080002: 'You successfully verified your email address.',
+        1080002: emailVerifiedSuccessfully,
         4000001: '%reason%',
-        4000002: 'Property %field% is missing.',
+        4000002: '%field% is missing.',
         4000003: 'Length must be \u003e= %expected_length%, but got %actual_length%.',
         4000004: '"%value%" is not valid "%format%"',
```
elbotho: side note: we never replace the placeholders here and nobody saw those strings while testing so we can maybe remove them.

src/data/en/index.ts
```diff
@@ -2,6 +2,9 @@
 import { Instance } from '@/fetcher/graphql-types/operations'
 import { getLogoutUrl, getPasswordChangeUrl } from '@/helper/urls/get-auth-urls'
 import { headerData, footerData, landingSubjectsData,secondaryMenus } from './menu-data'
+
+const emailVerifiedSuccessfully = 'You have successfully verified your email address'
```
elbotho: It makes sense, but since we did not need this in the other 1000 lines I'd rather avoid it.
Maybe just `string.auth.messages.1080002` with a comment above?

elbotho (changes requested):
Thanks for cleaning up this PR 👍
Way better reviewing experience now.
Code looks good, left some comments.

I can also take over from here if you like.

hugotiburtino:
Thanks for your review. 

> I can also take over from here if you like.

And thanks for your offer. Let me try a little bit more in order to keep learning React. If I get stock, I ping you.

src/data/en/index.ts
hugotiburtino: We can do so. 
Just to make it clear: after clicking there you go to another page, where you have to insert the email.

elbotho: Hmm, I see. Then maybe smth like "Click here to request the verification email again"?

src/data/en/index.ts
```diff
@@ -359,13 +357,12 @@ export const instanceData = {
         1060001: 'You successfully recovered your account. Please change your password in the next minutes.',
         1060002: 'An email containing a recovery link has been sent to the email address you provided.',
         1080001: 'An email containing a verification link has been sent to the email address you provided.',
-        1080002: emailVerifiedSuccessfully,
+        1080002: 'You have successfully verified your email address.',
```
elbotho, line 360: what do you say to my suggestion to not duplicate this but instead request this string with `…messages.[1080002]` same way you did [here](https://github.com/serlo/frontend/pull/1878/commits/f3a432016a83ca640c6276a09400ec8e675a8e39#diff-a5a83ae18eb89871dba21e44baa7a2df8e0374fed3429ae52ac536cc56bb03afR62)?

elbotho (approved):
Rest looks good 👍
Thank you.

src/data/en/index.ts
hugotiburtino: I actually thought that it was to little to deserve much code, so I didn't want to spend much time here.
But now I found a solution ([example here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAlpKBDMwCmARJyYF4YDeAUDDNAE4IDmEAXISaTGCFHOnQ00wPQ-Pg0MEADMYUABZoIQgO4hyAa0bc+MNAFskcADYA1NJRFw0AEwDKAV2AcIIyzp0BPepLgQAdBWqeklyR4a0hBIVNIA2gCMAAwAHNEJAEwAuiq8-Jra+oZwxmZWNsH2ji7wiCjoWMheUJRgNB5+AUEQIWEQUXEJ0SlEaaRqECAwskISSABuQlDDpsNwsJLSckhO-TBhsJm6BkYmFta2xc4AFACUBDDkaFCW5GBl0BWY2Eg1dQ1NEoHBoREx8SSyRgAF91moAEZoYB+GTwMSjGBzcQSdwjVbwWDjCDiYZQmAwxxmGBIHH2VBscDgnjbbJ7fKHIoOU4XK43O4PBBPVAvareeq+fzfFptf5dIGgtIggA0aS+9GI3FIIr+nEVStIAO6iXoAHIAJogSwwcZTMiM1rHJwwKb00wwJxG8jqLS6EmmUzXVoeXWyjVg7gByVgoigSAgHRoDw6EBUE66qGm6gknEAB3IIFThigTl9j2QPKqb35DRYbA4Hlpu1y+wKR2ZTjORD4YaGkejsfjhMj9tJMHJwEpYDzXILlVe7x8HjL7GkldddJrDMKlob5zOQA)) and I'm going to implement it

elbotho: This is what I meant: https://github.com/serlo/frontend/pull/1880/commits/b832ae9f5716c415b3beca36ece30c15b99b132a
Merge it if you like.

src/components/pages/auth/login.tsx
```diff
@@ -107,7 +107,21 @@ export function Login({ oauth }: { oauth?: boolean }) {
 
   async function onLogin(values: SubmitSelfServiceLoginFlowBody) {
     if (!flow?.id) return
-    const originalPreviousPath = sessionStorage.getItem('previousPathname')
+    const redirection = filterUnwantedRedirection(
+      sessionStorage.getItem('previousPathname')
+    )
+
+    function filterUnwantedRedirection(previousPath: string | null) {
```
elbotho: Cool, is this useful for the other auth redirects as well?
(If yes, maybe export it as a helper)

elbotho (approved):
👍

src/components/pages/auth/login.tsx
hugotiburtino: Good idea. It makes sense in `auth/logout`, but also in general. I'll make a helper out of it

elbotho (approved):

PR #1880, opened 2022-11-06
kratos feedback suggestion

elbotho:
- Suggestion for https://github.com/serlo/frontend/pull/1878
- Some refactoring, cleaner, less code, but not tested so far :)

### DIFF
--- a/src/components/pages/auth/verification.tsx
+++ b/src/components/pages/auth/verification.tsx
@@ -18,6 +18,8 @@ export function Verification() {
   const router = useRouter()
   const { flow: flowId, return_to: returnTo } = router.query
 
+  const emailVerifiedSuccessfully = strings.auth.messages[1080002]
+
   useEffect(() => {
     if (!router.isReady || flow) {
       return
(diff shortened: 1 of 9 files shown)


src/components/pages/auth/verification.tsx
```diff
@@ -18,56 +17,45 @@ export function Verification() {
   const router = useRouter()
   const { flow: flowId, return_to: returnTo } = router.query
 
+  const emailVerifiedSuccessfully = strings.auth.messages[1080002]
```
hugotiburtino, line 20: ah, that was what you meant! 😆
Sorry for overcomplicating it

hugotiburtino:
The refactorings seem to haven't broken anything (I've tested all flows).
~Please mark it as ready for review when you are ready.~
Now I saw https://github.com/serlo/frontend/pull/1878#discussion_r1014899036. Merging it

src/components/pages/auth/verification.tsx
elbotho: No worries 😀 Pretty interesting solution you found there with the `get`. 
Never saw that before.

PR #1894, opened 2022-11-10
Refactor editor integration

elbotho:


src/components/pages/add-revision.tsx
```diff
@@ -95,6 +97,41 @@ export function AddRevision({
...
+            } as SetEntityMutationData,
+            willNeedReview,
+            initialState,
+            taxonomyParentId
+          )
+
+    return new Promise((resolve: (value: void) => void, reject) => {
+      if (success) resolve()
+      else reject()
+    })
```
kulla, line 132:
```suggestion
    return success ? Promise.resolve() : Promise.reject()
```

src/components/pages/add-revision.tsx
```diff
@@ -95,6 +97,41 @@ export function AddRevision({
...
+          )
+        : await setEntityMutation(
+            {
+              ...data,
+              __typename: type,
+            } as SetEntityMutationData,
+            willNeedReview,
+            initialState,
+            taxonomyParentId
+          )
```
kulla, line 127: just a note: here I see the potential to combine all endpoints into one... would be a bigger refactoring step though...

src/mutations/use-set-entity-mutation/use-set-entity-mutation.ts
```diff
@@ -44,287 +44,265 @@ export function useSetEntityMutation() {
...
-          loggedInData,
-          isRecursiveCall: true,
-          savedParentId,
-          initialState,
-          router,
-        })
-        if (!success) throw 'revision of one child could not be saved'
+        try {
+          const result = await syncLoop()
+          return result
```
kulla, line 236:
```suggestion
          return await syncLoop()
```

kulla (approved):

PR #1897, opened 2022-11-17
kratos: hide show pw toggle

elbotho:


hugotiburtino:
Elegant.

UI question: usually you see the eye icon inside of the input field, but I suppose it is not that easy to do with the dynamic rendering of the fields, isn't it? (By the way, you don't need to change it, let's see what the stakeholders say)

General refactoring question: are you still comfortable with the dynamic rendering of the fields? If not, please open an issue for like hardcoding a little bit the /auth pages (it would be by the way a 'good first issue')

hugotiburtino (approved):
Let's ship it to staging, and the questions in https://github.com/serlo/frontend/pull/1897#issuecomment-1318365871 are more like inputs, don't need to be answered right now.

elbotho:
Thanks!

> UI question: usually you see the eye icon inside of the input field, …

I thought about it but opted to go for a more verbose button with text to make the function easier to understand.
Example from mail.google.com:
image

PR #1898, opened 2022-11-18
kratos: improve auth pages and mail templates

### LINKED ISSUE #1886: kratos: improve mail templates
@hugotiburtino:
>  make recovery and verification email templates more similar to the legacy ones (and the ones that we do not have in legacy make them simply more like our style - that is a matter of feeling)

elbotho:
- closes #1886
- a bit of refactoring and design improvements on the auth flow pages


hugotiburtino (approved):
Very good, thanks, it is better to read and see now.

PR #1900, opened 2022-11-21
kratos: refactor auth provider

### LINKED ISSUE #1887: kratos: auth/* pages don't update
> update /auth/* pages if in other tab session changed (example: one tab at /auth/verification, other at / and in this last one you log out)

elbotho:
- use state instead of ref to store authed state
- use window `online` (network change) and document `visibilitychange` (tab change/window change etc) events to refresh frontend auth state
- closes #1887

notes for reviewers:
- never mind the branch name
- [most relevant file](https://github.com/serlo/frontend/pull/1900/files#diff-f303b9fa2f4b9034b25b051c5022232a8bfddb1b8e152f4501be05f93f4672fe)

TODO:
- check for additional rerenders because of this change

--- 

> in 1900 there was the [Paris Exposition](https://en.wikipedia.org/wiki/Exposition_Universelle_(1900))

### DIFF
--- a/src/components/pages/auth/logout.tsx
+++ b/src/components/pages/auth/logout.tsx
@@ -2,10 +2,12 @@ import { useRouter } from 'next/router'
 import { useEffect } from 'react'
 
 import { filterUnwantedRedirection } from './utils'
+import { AuthSessionCookie } from '@/auth/auth-session-cookie'
 import { fetchAndPersistAuthSession } from '@/auth/fetch-auth-session'
 import { kratos } from '@/auth/kratos'
 import { AxiosError } from '@/auth/types'
 import { useAuth } from '@/auth/use-auth'
+import { useAuthentication } from '@/auth/use-authentication'
 import { LoadingSpinner } from '@/components/loading/loading-spinner'
 import { useInstanceData } from '@/contexts/instance-context'
 import { showToastNotice } from '@/helper/show-toast-notice'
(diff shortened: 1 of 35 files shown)


src/components/pages/auth/logout.tsx
```diff
@@ -2,35 +2,44 @@ import { useRouter } from 'next/router'
...
 
   useEffect(() => {
-    fetchAndPersistAuthSession().catch(() => {
-      return router.push('/')
-    })
-
     const redirection = filterUnwantedRedirection({
       desiredPath: sessionStorage.getItem('previousPathname'),
-      unwantedPaths: ['/auth/settings'],
+      unwantedPaths: ['/auth/settings', 'auth/login'],
```
hugotiburtino, line 25: Indeed, being in auth/logout, it is not possible to have auth/login as previousPathname, since everytime the loggedin user goes to auth/login, they get redirected. 
But in principle I can't see why someone that has just logged out may not go to the login page? Did you meet any problem here?

hugotiburtino (approved):

src/components/pages/auth/logout.tsx
elbotho: I had some infinite redirect at some point, but could also have been unrelated.

The idea of the redirects in my mind is, that users that start on a particular page but need auth (or no auth) for some reason end up on that page again after the auth action.
Automatically going from logout back to login makes no sense to me.

PR #1904, opened 2022-11-23
Kratos fix instances login

hugotiburtino:


hugotiburtino:
wrong base branch and right one already exists. Closing

PR #1905, opened 2022-11-23
kratos: fix login among different instances

hugotiburtino:
- [ ] Check if the solution will work also at localhost


hugotiburtino:
solved already by #1906

Issue #1903, opened 2022-11-22
kratos: test language versions


elbotho:
make sure other language versions work fine as well

hugotiburtino:
Problems found:

- [x] serlo/frontend#1907
- [x] serlo/frontend#1908
- [x] serlo/cloudflare-worker#340

elbotho:
Current behaviours:
- in legacy it's completely instance specific. I can be logged in as one user on de.serlo.org and as another on en.serlo.org
- in the current production it's a bit unstable and cause of some bugs where frontend and legacy are not in sync

hugotiburtino:
> in legacy it's completely instance specific. I can be logged in as one user on de.serlo.org and as another on en.serlo.org

I think the community would prefer to just be logged in as only one user . We could implement that already and it seems actually easier to do so.

Issue #1902, opened 2022-11-22
kratos: translate all strings


elbotho:
As soon as we are on staging make sure at least all german strings are translated

- [x] invalid verification. Subject. "Someone tried to verify this email address | Someone tried to verify this email address" and typo "sonst kannst" -> Sonst kannst
- [x] german string
- [x] spanish strings

(rest is not critical)

PR #1915, opened 2022-11-24
kratos: all flows

elbotho:
Original PR: https://github.com/serlo/frontend/pull/1779

# TODOs
- [x] #1903
- [ ] some testing with german strings
- [ ] rocket chat: general oauth login flow
- [ ] rocket chat: it works in localhost, but not in staging. Problem identified (always remeber: in such cases, run to CF Worker!), Solution on the way
- [ ] rocket chat: register user that has no rocket chat acc yet

Done 🎉
- [x] https://github.com/serlo/frontend/issues/1886
- [x] https://github.com/serlo/frontend/issues/1887
- [x] #1902
- [x] check bundle size
- [x] todo comments
- [x] test image upload
- [x] test image upload with just registered user


.env.local
```diff
@@ -10,3 +10,6 @@ HYDRA_HOST_STAGING=https://hydra.${NEXT_PUBLIC_SERLO_DOMAIN_STAGING}
 HYDRA_CLIENT_ID_LOCAL=frontend.serlo.org
 HYDRA_CLIENT_SECRET_LOCAL=frontend.serlo.org
 HYDRA_HOST_LOCAL=http://localhost:4444
+KRATOS_HOST_STAGING=https://kratos.${NEXT_PUBLIC_SERLO_DOMAIN_STAGING}
+KRATOS_HOST_LOCAL=http://localhost:4433
+ORY_SDK_URL="(NEXT_PUBLIC_ENV=local)?KRATOS_HOST_LOCAL:KRATOS_HOST_STAGING"
```
hugotiburtino: ORY_SDK_URL is maybe unnecessary, check it

elbotho: Seems to be unused. will remove it now, let's see if it breaks something :)

hugotiburtino: Just for info: Its use is hidden in https://github.com/serlo/frontend/blob/ded405a75d9d924a0fc1880becbb46493bddfbd9/src/pages/api/.ory/%5B...paths%5D.ts
See https://github.com/ory/integrations#nextjs
But if you go to the [source code](https://github.com/ory/integrations/blob/main/src/next-edge/index.ts#L44-L58) you see that we already overwrite it

src/pages/api/.ory/[...paths].ts
```diff
@@ -0,0 +1,22 @@
...
+
+const KRATOS_HOSTS = {
+  production: process.env.KRATOS_HOST_PRODUCTION,
+  staging: process.env.KRATOS_HOST_STAGING,
+  local: process.env.KRATOS_HOST_LOCAL,
+}
+
+const KRATOS_HOST = KRATOS_HOSTS[process.env.NEXT_PUBLIC_ENV || 'local']
+
+// TODO: this should probably be handled in CF Worker instead since it changes independent of Frontend.
```
Comment (elbotho) [inline src/pages/api/.ory/[...paths].ts:17]: I think this is fine for the foreseeable future. @hugotiburtino or do you have a solution in the making?

hugotiburtino:
[Umzug Usermanagement](https://trello.com/c/XhyRVD6U/12-umzug-usermanagement)

PR #1927, opened 2022-11-28
kratos: less data in auth cookie

elbotho:
if we sort this out we can merge `kratos` into `staging` imho


src/pages/api/oauth/accept-consent.ts
```diff
@@ -14,7 +14,7 @@ export default async function acceptConsent(
 ) {
   const { consent_challenge } = req.query
 
-  const session = AuthSessionCookie.parse(req.cookies)
+  const session = (await kratos.toSession()).data
```
hugotiburtino: It caused connection refused 127.0.0.1:80

src/pages/api/oauth/accept-login.ts
```diff
@@ -14,7 +14,7 @@ export default async function acceptLogin(
 ) {
   const { login_challenge } = req.query
 
-  const session = AuthSessionCookie.parse(req.cookies)
+  const session = (await kratos.toSession()).data
```
hugotiburtino: It caused connection refused 127.0.0.1:80, see other comment

src/pages/api/oauth/accept-consent.ts
```diff
@@ -2,7 +2,7 @@ import { gql } from 'graphql-request'
 import { NextApiRequest, NextApiResponse } from 'next'
 
 import { createGraphqlFetch } from '@/api/graphql-fetch'
-import { AuthSessionCookie } from '@/auth/auth-session-cookie'
+import { kratos } from '@/auth/kratos'
 
 export const config = {
```
hugotiburtino, line 8: Aside your code, this experimental edge (that I may have added after testing), is causing at least in localhost that query in req.query (line 15) is undefined.

src/pages/api/oauth/accept-login.ts
```diff
@@ -2,7 +2,7 @@ import { gql } from 'graphql-request'
 import { NextApiRequest, NextApiResponse } from 'next'
 
 import { createGraphqlFetch } from '@/api/graphql-fetch'
-import { AuthSessionCookie } from '@/auth/auth-session-cookie'
+import { kratos } from '@/auth/kratos'
 
 export const config = {
```
hugotiburtino, line 8: Let's remove it, see comment above.

hugotiburtino (changes requested):

hugotiburtino:
@elbotho 
if you want to try it by yourself, you can try with local-dev-env again (make sure to git pull on main and to follow the instruction in the readme there).

~If you do it, you may notice that something is strange in the authentication, when user tries to login in rocket chat without previous login in serlo `api/oauth/accept-login?login_challenge=function toString() {%20%20%20 [native code]}`, but this problem is also in kratos branch. I'll investigate it~ solved in kratos branch

src/pages/api/oauth/accept-consent.ts
elbotho: I did not touch the `experimental-edge` setting or `req.query`.
Not sure what is different then. Sure the same test in `kratos` branch has a different outcome?

elbotho:
for me the session fetching now works locally

src/pages/api/oauth/accept-consent.ts
hugotiburtino: > Sure the same test in kratos branch has a different outcome?

Sorry, I meant that also in `kratos` branch we have this issue (BTW only with localhost frontend and local dev env) and we could remove it.

elbotho: ahhh, now it makes sense :)
The syntax is a bit different for the edge routes (see [ea50fb1](https://github.com/serlo/frontend/pull/1927/commits/ea50fb1c439880e2f3c55525483fac3da2cd2fce))

hugotiburtino: >  (BTW only with localhost frontend and local dev env)

I have to correct myself here: localhost frontend, also against staging backend services.

hugotiburtino: @elbotho I'll remove them anyway, I'm getting unsure about it after so many warnings that the edge API may change .

It would also be hard to debug, if they sunddelly change  it...

elbotho: Alright, if you do it in this branch you'd have to revert my updates.

src/pages/api/oauth/accept-consent.ts
```diff
@@ -1,38 +1,49 @@
...
-    mutation ($input: OauthAcceptInput!) {
-      oauth {
-        acceptConsent(input: $input) {
-          redirectUri
+  try {
+    const { searchParams } = new URL(req.url)
+    const challenge = searchParams.get('consent_challenge')
+
+    const cookie = req.cookies.get('ory_kratos_session') ?? ''
+    const sessionResponse = await fetch(`${KRATOS_HOST}/sessions/whoami`, {
```
hugotiburtino, line 23: I'm getting here the following error:
```
TypeError [ERR_INVALID_URL]: Invalid URL
```
~it makes sense also.~(sorry, it doesn't)

Besides that and more important: `kratos.toSession` is just a less verbose wrapper around a request that hits this same endpoint. Let me try again with it and I'll tell you here

elbotho: interesting, in my tests I never got this error.

> Besides that and more important: kratos.toSession is just a less verbose wrapper around a request that hits this same endpoint.

true, I just tried it this way, but `.toSession` should work fine on the server if we forward the cookie.

src/pages/api/oauth/accept-consent.ts
elbotho: > It would also be hard to debug, if they sunddelly change it...

I follow nextjs changelogs quite closely and they usually communicate those things quite well and the types would also change. So either way I'm not that worried.

elbotho:
superseded by #1938 for now

PR #1928, opened 2022-11-29
kratos: stripped-down cookie

elbotho:


src/auth/auth-session-cookie.ts
```diff
@@ -37,3 +34,37 @@ export const AuthSessionCookie = {
     return JSON.parse(session) as Session
   },
 }
+
+interface Traits {
```
hugotiburtino, line 38: wonderful, I was missing it.

hugotiburtino (approved):

PR #1929, opened 2022-11-29
hotfix: discussions bugs

elbotho:
- threads can now be collapsed again
- scrolling is deactivated on `/discussions` page

for https://serlo.slack.com/archives/C0BMSC431/p1666617395168269


src/components/comments/comment-area.tsx
```diff
@@ -56,6 +57,10 @@ export function CommentArea({
       window.location.hash.startsWith('#comment-')) ||
     highlightedCommentId !== undefined
 
+  const [showThreadChildren, setShowThreadChildren] = useState<string[]>(
```
kulla, line 60: Why is a state necessary here?

kulla (approved):

src/components/comments/comment-area.tsx
elbotho: because some users really like the option to toggle the visibility of the comments in one thread

kulla: I ask because `setShowThreadChildren` is nowhere used. Just a comment though -> no blocker :smile:

elbotho: Alright, It's used in this file, so we are good :)

PR #1938, opened 2022-12-02
kratos: cookie refactor light

elbotho:
@hugotiburtino this one should not change behaviour

(should build when current api is online again)


hugotiburtino (approved):

hugotiburtino:
thanks.
Resuming the discussion about experimental edge, I would suggest to keep it aside for now (as you made in this PR, thanks for that also) and, when we are sure that rocket chat in production is working fine, you could try it again if you like it

PR #1940, opened 2022-12-05
kratos: simplify oauth

elbotho:


src/auth/oauth-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const oauthHandler = async (type: HandlerType, challenge?: string) => {
+  const session = AuthSessionCookie.parse()
+
+  if (!challenge) throwError('no challenge provided!')
+
+  const variables = {
+    input: {
+      session,
```
hugotiburtino, line 51: Just FYI: We'll remove it later...

src/auth/oauth-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+      challenge,
+    },
+  }
+  const args = JSON.stringify({ query: queries[type], variables })
+
+  try {
+    const dummyAuth = { username: '', id: 0 }
```
hugotiburtino, line 58: Shouldn't we put a TODO here (or something else that Actions don't complain about, but to remember us to improve it later)?

hugotiburtino (approved):
Thanks, it is cleaner and

hugotiburtino:
Thanks a lot, it is cleaner so and works perfectly.

PR #1948, opened 2022-12-07
kratos: Simple password checker

hugotiburtino:


elbotho (changes requested):
Hej @hugotiburtino,
thanks for the PR. I have some requests, should I comment them all or take over this feature?

hugotiburtino:
> should I comment them all or take over this feature?

You can take it over, thanks ;)

elbotho:
@hugotiburtino do you agree with my changes?

- files should be named like their export
- more generic rules to make it work better for german passwords as well
- results in shortened list

src/helper/check-for-simple-password.ts
```diff
@@ -0,0 +1,35 @@
+export function checkForSimplePassword(password?: string) {
+  if (!password) return false
+  if (/^[0-9]+$/.test(password)) return true
```
hugotiburtino:
```suggestion
```
A very long number is secure enough

src/data/en/index.ts
```diff
@@ -345,6 +345,7 @@ export const instanceData = {
       register: {
         registerTitle: 'Register your Serlo Account',
         passwordRequirements: 'At least 8 characters, longer is better.',
+        passwordTooEasy: 'This password is quite easy to guess. Make it longer or mix in letters, numbers and special characters.',
```
hugotiburtino: Actually, mixing character types is not recommended.

> Don't require mixtures of characters types

[https://www.ory.sh/docs/concepts/password-policy#best-practices](https://www.google.com/url?q=https://www.ory.sh/docs/concepts/password-policy%23best-practices&sa=D&source=docs&ust=1670505703220758&usg=AOvVaw2CO5j2rCo-p-XWBXvYERy2)


Here are the reasons:
image

src/helper/check-for-simple-password.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (password.length < 12) {
+    const withoutCommonNumbers = password
+      .replace('123', '')
+      .replace('1234', '')
+      .replace('1', '')
+    if (/^[a-z]+$/.test(withoutCommonNumbers)) return true
```
hugotiburtino, line 9: Actually, it can be hard enough:
image

hugotiburtino (commented):
@elbotho 
About the refactorings sure thanks 👍
Regarding rules: to be honest I don't think we can make it so dynamic without using a library.

PR #1952, opened 2022-12-10
fix: Redirect user to OAuth client after first wrong login attempt

### LINKED ISSUE #1951: kratos: oauth bug after first wrong password attempt
Steps to reproduce (in staging):
1. not logged in at serlo
2. go to rocket chat and try login with serlo
3. first attempt with wrong password
4. second with right password
5. stuck at serlo (not get redirected)

But if at 3 I use the right password, I get correctly redirected.

bug found: after clicking submit button the query paramenter is replaced by the one with the flow id

hugotiburtino:
~It has to be used in conjunction with remember=false.~ Actually, it doesn't, I was misled by the fact described [here](https://github.com/serlo/frontend/pull/1952#issuecomment-1345386501)

It should also fix the infinite login in serlo login page after coming from rocket chat.

Pushing flow id to url is a way of user not lose data. We may want to test if it is what we actually want to happen. edit: tested and it is indeed better, see discussion below.

Thinking: may we want to do the same for other flows... 🤔 edit: answer is yes

### DIFF
--- a/src/components/pages/auth/login.tsx
+++ b/src/components/pages/auth/login.tsx
@@ -153,7 +153,7 @@ export function Login({ oauth }: { oauth?: boolean }) {
         .submitSelfServiceLoginFlow(flow.id, values)
         .then(({ data }) => {
           void fetchAndPersistAuthSession(refreshAuth, data.session)
-          if (oauth) void oauthHandler('login', String(loginChallenge))
+          if (oauth) return oauthHandler('login', String(loginChallenge))
           const username =
             getAuthPayloadFromSession(data.session)?.username ?? 'Jane Doe'
           showToastNotice(
--- a/src/components/pages/auth/login.tsx
+++ b/src/components/pages/auth/login.tsx
@@ -147,9 +147,9 @@ export function Login({ oauth }: { oauth?: boolean }) {
 
   async function onLogin(values: SubmitSelfServiceLoginFlowBody) {
     if (!flow?.id) return
-    await router.push(`${router.pathname}?flow=${flow.id}`, undefined, {
-      shallow: true,
-    })
+
+    router.query.flow = flow.id
+    await router.push(router)
 
     const redirection = filterUnwantedRedirection({
       desiredPath: sessionStorage.getItem('previousPathname'),
--- a/src/components/pages/auth/login.tsx
+++ b/src/components/pages/auth/login.tsx
@@ -47,13 +47,8 @@ export function Login({ oauth }: { oauth?: boolean }) {
     }
 
     if (oauth) {
-      if (auth) {
-        void oauthHandler('login', String(loginChallenge))
-        return
-      } else {
-        // eslint-disable-next-line no-console
-        console.log('attempted oauth without auth')
-      }
+      void oauthHandler('login', String(loginChallenge))
+      return
     }
 
     if (flowId) {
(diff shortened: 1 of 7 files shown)


src/components/pages/auth/login.tsx
```diff
@@ -147,9 +147,9 @@ export function Login({ oauth }: { oauth?: boolean }) {
 
   async function onLogin(values: SubmitSelfServiceLoginFlowBody) {
     if (!flow?.id) return
-    await router.push(`${router.pathname}?flow=${flow.id}`, undefined, {
-      shallow: true,
-    })
+
+    router.query.flow = flow.id
+    await router.push(router)
```
elbotho: @hugotiburtino not sure why pushing router to router works and what it should do. Could you explain or send  a docs link?

hugotiburtino: I was also surprised that it works. Based on https://stackoverflow.com/questions/61466891/nextjs-appending-a-query-param-to-a-dynamic-route.

Since push expects a Url and router has the corresponding fields it ends up working.

elbotho: But what is the difference to the current code (besides the routing not being shallow)?

hugotiburtino: It preserves the other query params that the url may have. In the specific case of the OAuth path it is important to maintain the login_challenge query parameter in order to solve the bug of the linked issue.

hugotiburtino: We could alternatively not change the url at all and if the user refreshes the page they lose their previous inputs (and previous error messages,  if any). At the end of the day it is maybe what they want

elbotho: thanks for the explanation.

>We could alternately not change the url at all and if the user refreshes the page they lose their previous inputs (and previous error messages, if any). At the end of the day it is maybe what they want

yeah, I think for login we should try that for now.
Maybe it's different for e.g. registering.

hugotiburtino: Sorry, I have to correct myself: the recovery of data based on flow id is not implemented, Indeed we get the data from backend, but we don't fill in the fields with them So, the way it is now the user gets a very bad experience after refresh, 

See here in registration: error message with empty field
image

I think it is safe to delete these lines in login and in registration (I'll test with other flows and if I see that there isn't problem also you will get my commits on that)

hugotiburtino:
Regarding [086c9e6](https://github.com/serlo/frontend/pull/1952/commits/086c9e6b5ef2d7759c824ff1534b15c5bba182bb), I had the following bug: after clicking the Serlo login in local rocket chat, my browser redirected me sometimes to landing page, sometimes back to rocket chat. 
I supose there was some kind of race between the `oauthHandler` and `router.push(flow.return_to ?? redirection)`

src/components/pages/auth/login.tsx
hugotiburtino: > I think it is safe to delete these lines in login and in registration (I'll test with other flows and if I see that there isn't problem also you will get my commits on that)

Of course, unless you think we should make effort to implement it now...

Otherwise, such lines are not only useless but also reason for bad UX. And there is no downside: if user reloads the page, they get another flow id and can go on.

elbotho (approved):
👍

hugotiburtino:
Thanks for approving it. I'm not allowed to merge it. Could you do it?

PR #1960, opened 2022-12-12
kratos: add maintenance code to staging

elbotho:
this could be merged as soon as the community testing is finished today

### DIFF
--- a/src/components/no-login-tonight.tsx
+++ b/src/components/no-login-tonight.tsx
@@ -1,0 +1,23 @@
+//TODO: remove again after deploy
+
+import { PageTitle } from './content/page-title'
+
+export function NoLoginTonight() {
+  return (
+    <div>
+      <PageTitle title="Tut uns keid, keine Logins heute Nacht" />
+      <p className="serlo-p">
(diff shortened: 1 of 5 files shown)


src/components/no-login-tonight.tsx
```diff
@@ -0,0 +1,23 @@
+//TODO: remove again after deploy
+
+import { PageTitle } from './content/page-title'
+
+export function NoLoginTonight() {
+  return (
+    <div>
+      <PageTitle title="Tut uns keid, keine Logins heute Nacht" />
+      <p className="serlo-p">
+        Heute Nacht (12.12. ab 20:30) arbeiten wir am Login-System, deswegen ist
```
hugotiburtino:
```suggestion
        Heute Nacht (12.12. ab 20:30, deutsche Zeit) arbeiten wir am Login-System, deswegen ist
```

hugotiburtino (approved):
Thanks for improving the maintainance mode.
Sure, it would be better to have i18n, at least in English, but anyway it is good so.

PR #1964, opened 2022-12-12
production backup before kratos

elbotho:


hugotiburtino:
good idea, hopefully we won't need it

PR #1976, opened 2022-12-14
Temporarily disable Bot Or Not

hugotiburtino:


elbotho (approved):

hugotiburtino:
@elbotho Good that you didn't disabled it still. Actually, it is quite a lot manual work to delete bot otherwise. Let's let it enabled

PR #1996, opened 2023-01-04
swr 2.0

elbotho:
did not solve the loop issue :)
I did not include the new features and refactors we could use with 2.0 but migrated some small things and it does not increase the bundle, so …


Entkenntnis (approved):
looks innocent :)

PR #2001, opened 2023-01-05
fix for node 18 v2

Entkenntnis:
@elbotho after looking through all alternatives this was the best I could find, the module should be excluded from client now.


elbotho (approved):
Thanks you for looking into this!
Not sure why bundle size action did not run here but you checked it anyway, right?

Entkenntnis:
Yes, I tested it locally. The Deployment PR will run it then, I think.

PR #2005, opened 2023-01-06
add fallbackNodePolyfills with process.env checks

elbotho:
Not sure why, but those checks solve the error on localhost.


Entkenntnis (changes requested):
we are not 100% using .evn files correctly, see #2007

Entkenntnis:
@elbotho I have more time on monday and will merge it then into staging

Entkenntnis (approved):

PR #2008, opened 2023-01-06
fix warning after successful save

### LINKED ISSUE #2000: successful saving still triggers leave warning
I did a lot of refactoring on the edtr io implementation in the frontend.
Renamed variables to make them ambiguous, removed some legacy code, moved code into hooks, etc…

Most things seem to work fine but I'm running into an annoying problem.

the edtr check to see if there are unsaved changes (`useScopedSelector(hasPendingChanges())`) still has the old value `true` ("has changes") when I check if I should warn the user before changing the route.

@kulla do you have an idea if I can check this as a function (not a hook) instead or if I need to keep track of that myself in the frontend?

elbotho:
Needs testing:
- [x] Entity save
- [ ] Page save
- [ ] Profile description save
- [ ] Taxonomy save

Unfortunately local dev is pretty broken right now (https://github.com/serlo/cloudflare-worker/issues/361) and testing is postponed (or we do it on staging).

### DIFF
--- a/src/edtr-io/components/save-local-button.tsx
+++ b/src/edtr-io/components/save-local-button.tsx
@@ -2,7 +2,7 @@ import { useScopedStore } from '@edtr-io/core'
 import { serializeRootDocument } from '@edtr-io/store'
 import { useEffect, useState } from 'react'
 
-import { storeState } from '../serlo-editor'
+import { storeStateToLocalStorage } from '../serlo-editor'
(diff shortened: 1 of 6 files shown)


elbotho:
@Entkenntnis if you don't object I'd go ahead and test this on staging.

Entkenntnis (approved):
thanks

PR #2024, opened 2023-01-10
thanks: friendly response

elbotho:


src/components/content/thank-you.tsx
```diff
@@ -16,22 +15,24 @@ export function ThankYou() {
...
+    <p className="serlo-p text-right mt-8 mb-16">
+      {submitted ? (
+        <b>Danke für dein Feedback! 💙</b>
+      ) : (
+        <>
+          Hat dir der Inhalt geholfen?{' '}
+          <button
+            className="serlo-button-green ml-4"
+            onClick={() => {
+              submitEvent(`thank_you_${id}`)
```
elbotho: tbh. I'm not a big fan of this approach since it wrecks the simple analytics ui with the amount of events.
You can use https://frontend-git-sa-event-evaluation-serlo.vercel.app/___sa_event_evalulation (might need a tiny change) to display the pages where the events occurred when using the same event name (e.g. `thank_you_article`)

Entkenntnis: 👍 this approach was necessary in the legacy event system because it was difficult to find the path. As I have tested it yesterdey, The new events API provides more context including the path, so I can remove the $id part.

Entkenntnis (approved):
nice!

PR #2025, opened 2023-01-10
editor: tab change bug

elbotho:


Entkenntnis (changes requested):
#2027 would hopefully fix the root cause

elbotho:
> https://github.com/serlo/frontend/pull/2027 would hopefully fix the root cause

🎉 it did. If you don't mind I would merge https://github.com/serlo/frontend/pull/2025/commits/a077b9312e96890ad170ea287187cfab6a81fc54 to staging just to make to a bit more secure.

Entkenntnis:
yes, you can merge this extra check, won't hurt

elbotho:
> yes, you can merge this extra check, won't hurt

You already did, right?

PR #2027, opened 2023-01-10
fix stale closure in auth-provider

### LINKED ISSUE #2026: editing content directly after login leads to lost data when changing tabs
Steps to reproduce:
- log out
- refresh page (f5)
- login in
- edit page (navigate with csr only)
- do an edit
- change tab
- change back
- BUG: your edit is lost

Root of the cause is most like a [stale closure](https://dmitripavlutin.com/react-hooks-stale-closures/) in here:

https://github.com/serlo/frontend/blob/325799755713328cf0625e3a847066ea3a905c66/src/auth/auth-provider.tsx#L28-L50

The value of  `authenticationPayload` is captures as null and never changes in the function `refreshWhenVisible`.

Entkenntnis:


elbotho (approved):
you legend, thank you

PR #2029, opened 2023-01-10
fix dom validation

### LINKED ISSUE #2028: nested p in p
habens in article related exercise folders

Entkenntnis:
this line is wrapped in p:

https://github.com/serlo/frontend/blob/325799755713328cf0625e3a847066ea3a905c66/src/edtr-io/plugins/article/article-exercises.tsx#L24

but later it's wrapped in p again. Use div instead


elbotho (approved):
thanks! 🎉

PR #2038, opened 2023-01-13
hide thank you button in revision view

### LINKED ISSUE #2035: hide thank you button in review view
@elbotho so we have to check the path for `"revision"`?

Entkenntnis:


elbotho (approved):
👍

PR #2046, opened 2023-01-16
remove usage of @apply and @screen

### LINKED ISSUE #2012: refactor css
Unfortunately, there is no way of using styled-jsx with postcss without relying on babel -> this is a opt-out for the new swc compiler. The only way in the meanwhile is to remove unsupported features from styles:

### tailwind `@apply`

- set classname directly or
- use arbitrary selector to set style on multiple elements

### nesting

- use css modules?
- manually unnest
- ...?

I would like to finish this in the next 1 - 2 month, if we work on this now and then it should be doable. I can take care of the tailwind stuff, @elbotho can you take a look at the nesting situation? we would also like to use turbopack, and they have postcss support (https://nextjs.org/blog/next-13-1#turbopack-improvements)

Entkenntnis:
first step in the css refactoring

remove all usages of @apply and @screen


src/edtr-io/plugins/types/helpers/revision-history-loader.tsx
```diff
@@ -85,12 +85,18 @@ export function RevisionHistoryLoader<T>(
       </ModalWithCloseButton>
       <style jsx global>{`
         .ReactModalPortal .ReactModal__Content {
-          @apply w-[900px] max-w-[90vw] max-h-[80vh];
-          @apply translate-y-0 top-8 overflow-y-scroll;
-          @apply pt-0;
+          /* this css colides with save-modal.tsx */
```
Entkenntnis, line 88: I haven't refactored this, but instead kept it as is

src/helper/breakpoints.ts
```diff
@@ -0,0 +1,7 @@
+export const breakpoints = {
```
elbotho, line 2: those are straight from tailwind, right? 
Maybe we should add that in a short comment.

src/edtr-io/components/save-modal.tsx
```diff
@@ -226,17 +231,10 @@ export function SaveModal({
           }}
         />{' '}
         <span
-          className="license-wrapper"
+          className="license-wrapper [&_a]:!text-brand hover:[&_a]:underline"
```
elbotho, line 234: this syntax hurts a bit to be honest :D 
But if it works…

elbotho (approved):
Thanks for starting this! 👍
If you checked the visuals and mobile views for regressions I approve :)

src/edtr-io/components/save-modal.tsx
Entkenntnis: it works :)

src/helper/breakpoints.ts
Entkenntnis: ✔️

Issue #2041, opened 2020-07-22
edtr-io: persist last working state in localStorage


inyono:
Seems like there currently is an issue where one can't input changes (sometimes, possibly after long edit sessions... Could be any uncaught error, though). We should persist the last working state always in localStorage (not only when saving fails) so that one can restore that.

elbotho:
(this is a legacy issue, but is still relevant imho)

Entkenntnis:
we can listen to `onChange` in `serlo-editor.tsx` and store data in local storage

@elbotho Do you know if the existing localStorage logic is part of the frontend or core? It would be best if we can reuse the same code.

elbotho:
It's part of the frontend code and your suggestion would be easy to try.
We might have to tinker with performance on big states. Most solutions (for server-side storage though) I saw save changes not the whole state.

Entkenntnis:
deployed

PR #2059, opened 2023-01-20
try periodical local autosave

elbotho:
for #2041

Basic idea:
- Für jede Revision die ein User im Editor öffnet speichern wir den State im localStorage
- Das speichern passiert automatisch (5 secs after last edit) (~fixed 10s interval for now since onChange requires more work~)
- Wenn vorhanden wird per infobox informiert, dass es eine gespeicherte Version gibt (mit datum + uhrzeit) die geladen werden kann (some improvement over alert).
- Wenn die Version geladen wurde kann User an gleicher Stelle auch seine Änderungen verwerfen

TODO:
- [x] cleanup UI
- [x] wording and i18n


elbotho:
@Entkenntnis not finished but could you check the overal idea and flow?
I'll work on the rest when I'm a bit more rested.

`onChange` with a simple debounce would be a way nicer solution, but onChange is available in `SerloEditor` and the store is not so I implemented it this way for now.

Entkenntnis:
Thank you for this PR - the overall flow is good.

Time-based is fine, if you want to use `onChange` you can use the `getDocument()` function, this should give you the serialized state (https://edtr.io/docs/getting-started#persisting-documents)

storing everything in one localstorage entry could be a bottleneck, because you have to parse everything again... I was also thinking about session storage as an option if it's enough for the use case

elbotho:
> storing everything in one localstorage entry could be a bottleneck, because you have to parse everything again... I was also thinking about session storage as an option if it's enough for the use case

What difference between local- and sessionStorage are you refering to here?
Big downside to sessionStorage When the users browser is closed unintentionally or the system halts (or smth like that) all changes are still lost.

elbotho:
> Time-based is fine, if you want to use onChange you can use the getDocument() function, this should give you the serialized state (https://edtr.io/docs/getting-started#persisting-documents)

Thanks for the link 👍

elbotho:
@Entkenntnis ok ready from my side. Any more suggestions?
Otherwise I'd start testing this in staging.

Entkenntnis (approved):
thank you! you can start testing on staging

PR #2060, opened 2023-01-22
yarn: Add config "supportedArchitectures"

kulla:
See https://yarnpkg.com/configuration/yarnrc#supportedArchitectures (This should install the next version for `darwin` and `linux`)


elbotho (approved):
nice find, thank you.

PR #2066, opened 2023-01-24
Fix Typo in Conditional Wrap Component

pixelmord:
Still needs fix for all uses of the component


kulla (approved):

elbotho:
Thanks @pixelmord nice catch.

PR #2069, opened 2023-01-26
use swc compiler by removing babel plugin dependency

### LINKED ISSUE #2012: refactor css

Entkenntnis:
I have run all nested css manually through `tailwindcss/nesting`-plugin.

It's a bit desperate ... but I haven't found any better solution for the mean time


Entkenntnis:
bundle size decrease is surprising 👀 

~todo: fix jest~ done

kulla (approved):

elbotho (approved):

PR #2077, opened 2023-01-27
fix: Center first explanation in equation plugin

kulla:
Before: 
2023-01-27-140030_127x212_scrot
 After: 
2023-01-27-140252_122x219_scrot


Entkenntnis (approved):
thank you

PR #2081, opened 2023-01-31
add ability to sort exercises (=exercise folder)

### LINKED ISSUE #2080: Sorting children of some taxonomies is not working
# What is wrong?

In some taxonomies the sorting feature is not working.

# Steps to reproduce
can log in in staging as a user with this rights, go to [219731](https://de.serlo.org/mathe/219731/jahrgangsstufentest) and click to sort elements: no element is shown.
It affects many taxonomies in https://de.serlo-staging.dev/83249/mathematik-pr%C3%BCfungen

# How should it behave?
After clicking sort button you should be able to sort it like in https://de.serlo-staging.dev/taxonomy/term/sort/entities/91252

_Reported by Karin_

Entkenntnis:


src/pages/taxonomy/term/sort/entities/[id].tsx
```diff
@@ -43,6 +43,7 @@ export const allCategories = [
   TopicCategoryType.events,
   TopicCategoryCustomType.subterms,
   TopicCategoryCustomType.exercisesContent,
+  TopicCategoryType.exercises,
```
Comment (hugotiburtino) [inline src/pages/taxonomy/term/sort/entities/[id].tsx]: And what about the other categories in `TopicCategoryType`, don't they come up? (actually just `folders` now)

hugotiburtino (commented):
And maybe it is easier to see what is missing if `allCategories` were ordered alphabetically...

src/pages/taxonomy/term/sort/entities/[id].tsx
Comment (Entkenntnis) [reply src/pages/taxonomy/term/sort/entities/[id].tsx]: grafik

"Das passt zu mir" is a `folder`, but because it's nested, we don't want to sort them (they shouldn't appear on the higher level anymay. To sort them, you need to go to "Gesellschaftliches Engagement", so it is fine - I can add this as a comment

Comment (hugotiburtino) [reply src/pages/taxonomy/term/sort/entities/[id].tsx]: Thanks for your reply, I understand now.
Better than a comment it would be to find a more precise/descriptive name for the variable (allSupportedCategories, firstLevelCategories, I don't know...), but I think it is good enough so.

hugotiburtino (approved):

PR #2083, opened 2023-01-31
yarn: Add "show-unused-exports"

kulla:
Adds the command `yarn show-unused-exports` whichs shows unused exports (good to detect dead code in our database and thus good for refactoring):

2023-01-31-202830_930x773_scrot

@Entkenntnis @elbotho You might want to exclude more paths (especially autogenerated files) in the script command at https://github.com/serlo/frontend/blob/ts-unused-exports/package.json#L26


Entkenntnis (approved):
nice, not bad to review this from time to time

PR #2085, opened 2023-02-01
only allow numbers in injection input

### LINKED ISSUE #53: Trim input id of plugins
Not sure if it should be solved only in frontend (see https://github.com/serlo/backlog/issues/313)

Frontend could trim the id of plugin in order to avoid that users enter something like `9619 `, what is quite hard to an end user to perceive the trailing space and they may anyway not think that it can cause an error.

elbotho:
Closes https://github.com/serlo/serlo-editor-issues-and-documentation/issues/53

### DIFF
--- a/src/edtr-io/plugins/injection.tsx
+++ b/src/edtr-io/plugins/injection.tsx
@@ -51,7 +51,7 @@ function InjectionEditor(props: EditorPluginProps<typeof injectionState>) {
 
   const loggedInData = useLoggedInData()
   if (!loggedInData) return null
-  const editorStrings = loggedInData.strings.editor
+  const injectionsStrings = loggedInData.strings.editor.injection
 
   if (!props.editable) {
     return <InjectionRenderer src={props.state.value} />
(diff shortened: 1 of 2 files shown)


src/edtr-io/plugins/injection.tsx
```diff
@@ -80,11 +84,11 @@ function InjectionEditor(props: EditorPluginProps<typeof injectionState>) {
         <EditorInlineSettings>
           <EditorInput
             label={editorStrings.injection.serloId}
-            placeholder="123456"
+            placeholder="Serlo Id (e.g. 1565)"
```
hugotiburtino: It is in English, do we want it in all instances?

Entkenntnis (approved):
@elbotho Thank you for the fix, works fine. You can merge into staging.

I have added a translation string, hope it's added correctly

elbotho:
> I have added a translation string, hope it's added correctly

Thanks for i18n and review.
Almost perfect ;)

hugotiburtino (approved):

PR #2086, opened 2023-02-02
create breadcrumbs only with taxonomy data

### LINKED ISSUE #2075: refactor breadcrumbs: only depend on taxonomy term parents
previous issue: https://github.com/serlo/frontend/issues/1783

api issue: https://github.com/serlo/api.serlo.org/issues/703

One of the last blockers for legacy migration. After reviewing the technical details, here is a plan:

The frontend will generate the breadcrumbs depending on the taxonomy parents. It will not use `navigation` anymore. This will need a slightly hacky query (see api issue) and our existing choice algorithm for multiple assignments.

This will allow the api the remove all depencies to legacy and we can disable legacy in staging soon.

In a follow up, we could move this logic into the API.

Entkenntnis:
the main query is complete,

~open tasks: we have other queries that also need breadcrumbs - rewrite them as well~ done

- [x] en
- [x] es
- [x] fr
- [x] hi
- [x] ta


src/fetcher/create-breadcrumbs.tsx
```diff
@@ -1,28 +1,63 @@
...
+        MainUuidType,
+        { __typename: 'Article' }
+      >['taxonomyTerms']['nodes'][0]
+    | undefined,
+  instance: Instance
+): BreadcrumbsData | undefined {
+  if (taxonomyPath === undefined) return undefined
+  // because we don't have infinite recursion, this is the best we can do now
+  // move this function into the API
+  let current: RecursiveTree = taxonomyPath as RecursiveTree
```
hugotiburtino:
```suggestion
  let current = taxonomyPath as RecursiveTree
```

hugotiburtino (commented):
Wonderful, it is really good that you nailed it.

The only thing I noticed while manually testing was some differences between breadcrumbs before and after the changes. 

For example:
http://localhost:3000/220418 is "Mathematik > Deutschland > ... > Mittelschule > Klasse 6"
https://de.serlo.org/220418 is "Mathematik > Mittelschule (Hauptschule) > Klasse 6"

I suppose the second one is a little easier to navigate. Let's ask to stakeholders and see if it is a no-go.

Entkenntnis:
Thank you Hugo for testing this. That's actually another hidden functionality I overlooked, will check this on monday, but I think this can be fixed

PR #2103, opened 2023-02-13
remove broken lightbox svg fix

### LINKED ISSUE #2051: lightbox distorts images
image
e.g. https://de.serlo-staging.dev/mathe/1555/zylinder under "Volumen"

elbotho:
might not max out the display size, but it seems to work better without the hack (no more distorted svgs)


Entkenntnis (approved):
thank you, distorting is gone

PR #2110, opened 2023-02-14
image refactorings

### LINKED ISSUE #2109: native lazy image loading
Lazy loading attributes for images are now quite well supported and don't break things on older browsers.
We should consider replacing `react-lazyload` at least for the images.

https://caniuse.com/loading-lazy-attr
https://web.dev/browser-level-image-lazy-loading/

elbotho:
lazy loading seems to work with printing

test:
- in browser printing
- pdf export

(closes #2109)

### DIFF
--- a/src/components/content/image.tsx
+++ b/src/components/content/image.tsx
@@ -1,5 +1,4 @@
 import { useRouter } from 'next/router'
-import { ReactNode } from 'react'
 
 import { Lazy } from './lazy'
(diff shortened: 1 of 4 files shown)


Entkenntnis (approved):
thank you, looks good. printing can be testing on staging.

elbotho:
@Entkenntnis print export is pretty unstable in general, but the lazy loading images display fine in my tests

PR #2111, opened 2023-02-14
improve license metadata

elbotho:


Entkenntnis (approved):
thank you, let's try this out and hope that google is picking up the license

PR #2115, opened 2023-02-17
try next/image for team page

elbotho:
this being a limited amount of images it should be save to try here.

Performance is especially bad there because I forgot to lazy load the images and some of them are quite big.

Current results:
image

TODO: keep an eye on https://vercel.com/dashboard/serlo/usage


elbotho:
🔥
image

Entkenntnis (approved):
thank you, added dynamic import to fix bundle size

PR #2116, opened 2023-02-19
registration: add interest dropdown

### LINKED ISSUE #2099: ask for user background in registration
with theses options:
    Elternteil
    Lehrer*in
    Schüler*in
    Student*in
    Anderes/Sonstiges

## Blocker
- waiting for backend changes

hugotiburtino:
In legacy the field "I'm here as..." was called `interests`. 
That's the reason for the field name (in singular though).

Note that it has to be deployed in conjunction with https://github.com/serlo/cloudflare-worker/pull/380.
Otherwise, it will lead to errors.

The node.tsx component is getting every time worse with so many flags, we may want to use this feat to refactor it or to refactor the auth components to not be dependent of it.

### DIFF
--- a/src/components/auth/node.tsx
+++ b/src/components/auth/node.tsx
@@ -131,30 +131,35 @@ export function Node({
   function renderInput(attributes: UiNodeInputAttributes) {
     const basicFields = {
       className:
-        'text-xl serlo-input-font-reset serlo-button-light hover:bg-brand-150 focus:bg-brand-150 outline-none -ml-1 mt-1 text-brand hover:text-brand px-4 py-2 w-full',
+        'text-xl serlo-input-font-reset serlo-button-light hover:bg-brand-150 focus:bg-brand-150 focus:outline-none -ml-1 mt-1 text-brand hover:text-brand px-4 py-2 w-full border-2 border-transparent focus:border-brand border-solid',
       name: attributes.name,
       onChange: (e: { target: { value: string } }) => {
         void onChange(e.target.value)
       },
     }
+
     if (attributes.name === 'traits.interest') {
       return (
-        <select
-          {...{
-            ...basicFields,
-            className: basicFields.className + ' [&:invalid]:text-brand-400',
-          }}
-          required
-        >
-          <option value="" disabled selected className="hidden">
-            - {strings.auth.interests.pleaseChoose} -
-          </option>
-          <option value="parent">{strings.auth.interests.parent}</option>
-          <option value="teacher">{strings.auth.interests.teacher}</option>
-          <option value="pupil">{strings.auth.interests.pupil}</option>
-          <option value="student">{strings.auth.interests.student}</option>
-          <option value="other">{strings.auth.interests.other}</option>
-        </select>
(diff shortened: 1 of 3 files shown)


hugotiburtino:
CF worker is deployed in staging

src/components/auth/node.tsx
```diff
@@ -140,6 +130,40 @@ export function Node({
...
+      name: attributes.name,
+      onChange: (e: { target: { value: string } }) => {
+        void onChange(e.target.value)
+      },
+    }
+
+    if (attributes.name === 'traits.interest') {
+      return (
+        <select {...basicFields}>
+          <option value="nothing">{strings.auth.interests.pleaseChoose}</option>
```
elbotho: @hugotiburtino can we use an empty string if nothing is selected?

hugotiburtino: Hey @elbotho,
I've put a comment in 91bd831 (which was for some reason not pushed before, although my local git was saying it was, who know? Maybe gh cli bug...)
If you are uncomfortable with it, I could switch in [CF Worker](https://github.com/serlo/cloudflare-worker/pull/380/files) (add 'nothing'  there, in order to let frontend set it to ''). 

The only important things are: 
- some legacy users in the DB just don't have interest field (I supposed it was only later made required). We have to support it.
- In order to continue requiring the users to fill it in, value null/nothing/empty here and in CF should be different.

elbotho: Thanks for the explanation.
So I could replace `nothing` with `–` here? I'd just prefer something obviously different.

elbotho:
There are some styling issues, other than that I approve :)
I can also take over and try to solve the styling.

src/components/auth/node.tsx
hugotiburtino: > So I could replace nothing with – here? 

sure

hugotiburtino:
> I can also take over and try to solve the styling.

please 🙏 , I'm horrible with styling 😩

hugotiburtino:
Note that in the preview it has a quite good style, even though I haven't done anything for it...

(edit: ah, because I was accessing it from windows, in linux it continues ugly)

Entkenntnis (approved):
thanks for the work on this, I like it that the changes are very local and I'm looking forward to deploy this soon

elbotho:
> please 🙏 , I'm horrible with styling 😩

Styling form element so they look consistent across browsers feels like all of frontend-dev in the 90s ;)

src/components/auth/node.tsx
elbotho: @hugotiburtino In 3ac4774a76e17839b5fa6b6c400744fb9add415a I tried to find a default way of solving this.
Since it set to required it will actually not send a value for `interest` at all which seems to work fine with the backend.

Styling explained [here](https://www.w3docs.com/snippets/css/how-to-create-a-placeholder-for-an-html5-select-box-by-using-only-html-and-css.html) (second approach)

elbotho:
styling and refactoring done. @hugotiburtino merge if you approve the changes.

elbotho (approved):

hugotiburtino:
Thanks, also for solving the empty value issue

PR #2121, opened 2023-02-20
use modularized import for @fortawesome icons

### LINKED ISSUE #2013: import resolution for fontawesome
https://nextjs.org/blog/next-13-1#import-resolution-for-smaller-bundles is really nice and we can avoid optimizing fontawesome by hand ❤️

Entkenntnis:
I expect bundle size to stay the same or decrease :heavy_check_mark: 

DX improvement, because you don't need to worry about named imports of icons anymore


elbotho (approved):
Thank you! 🎉

PR #2123, opened 2023-02-20
fix init browser for kratos

### LINKED ISSUE #2023: add error message when login/kratos responds with 403 / csrf error
probably not happening often, but when I try to login in an incognito window and navigate to auth/login directly (not via client side routing) my first attempt to login fails with an 403 response.

Entkenntnis:
use effect is called twice, we add a safeguard to prevent this. This should solve #2023 

- [ ] Do some testing on staging after merge


Entkenntnis:
merging this for testing on staging

elbotho (commented):
If it works, I'm happy with this :) 
Let's hope we don't need this for all flows…

PR #2138, opened 2023-02-27
make comment tools more obvious

### LINKED ISSUE #2135: Make comment actions more intuitiv
Actions for comments (like editing, archiving and so on) are under the timestamp. Could we try to make it more intuitiv, like under three dots?

_From the beta testing document_

elbotho:
closes #2135


hugotiburtino (approved):

hugotiburtino:
Very good, 
In /discussions, when it has a bege background it is hard to see green line of the comment input field when not focused.
image
But how to highlight the last comment is something we want to change anyway.
We could ship for testing

Entkenntnis (approved):
thank you, works and looks good

PR #2145, opened 2023-03-01
2092 remaining usages of navigation

### LINKED ISSUE #2092: remaining usages of navigation
What I have found:

- create-title uses `navigation` to find subject
- `GetTaxonomyTypeQuery`
- `GetUuidPathsQuery`
- `GetSubscriptionsQuery`

We also need to refactor these.

Entkenntnis:
changes should be complete -> no more use of `navigation` in frontend

testing required, postponed until staging stabilizes

creating breadcrumbs should only happen server side

fix tests


Entkenntnis:
test protocol:

- go to exercise folder, click exercise: make sure that title contains subject
- create entity outside of testarea -> isTestArea is false
- create entity inside of testarea -> isTestArea is true
  - existing code looks broken, though (and nobody noticed?), out of scope here
- update linked taxonomy on entity -> paths are shown
- remove path -> works
- readd path -> works
- fix unique key warning on the way
- fix tests

hugotiburtino (approved):
Thanks, it looks even better now 😊

PR #2146, opened 2023-03-01
this should enable login on production build on localhost

### LINKED ISSUE #2097: production build on localhost should use local config
currently, it uses staging kratos and login fails (e.g. interesting for running e2e tests on localhost)

retry e2e tests after fixing this

Entkenntnis:
login/logout now working on localhost in production builds


hugotiburtino (approved):
Thanks. I'll have to check what it does if we are using local backend (serlo/local-dev-env), and maybe make a suggestion later.

BTW: the comment at l. 18 can be removed, we won't do it

Entkenntnis:
@hugotiburtino merging into staging for testing, if you have a suggestion I can of course adapt it

PR #2147, opened 2023-03-01
every entity should remount

### LINKED ISSUE #2100: exercises in course pages are not remounted
interesting case where excercises on page 2 & 3 are very similiar and not remounted (input is still visible):

https://de.serlo.org/community/266414/parameter-a

-> add e2e test for this

Entkenntnis:
I think it's not useful to optimize further as every entity should be it's "own" component


elbotho (approved):
Might even solve problems? Let's see!

PR #2149, opened 2023-03-01
avoid usage of random id

### LINKED ISSUE #2143: random ids of exercises causes hydration errors
I can finally reliable reproduce on localhost (http://localhost:3000/67055)

Entkenntnis:
the idBase should be unique enough for every use case I can currently think of


elbotho (approved):
nice if we don't need this weird hack 👍

PR #2160, opened 2023-03-03
h5p plugin

### LINKED ISSUE #2159: implement h5p as option for interactive exercise
This is how the intro-view is:

grafik

The first draft of the exercise should look exactly like geogebra and be enabled on staging. More features (content type checking, inline help, ... will be addressed in further issues)

Entkenntnis:
mvp, wip

https://frontend-git-2159-implement-h5p-as-option-for-inte-b31239-serlo.vercel.app/entity/create/Exercise/23869

### DIFF
--- a/src/components/content/privacy-wrapper.tsx
+++ b/src/components/content/privacy-wrapper.tsx
@@ -1,3 +1,4 @@
+import { faFaceGrinStars } from '@fortawesome/free-regular-svg-icons'
 import { faHeart, faSpinner } from '@fortawesome/free-solid-svg-icons'
 import clsx from 'clsx'
 import { useState, KeyboardEvent, useEffect } from 'react'
@@ -16,7 +17,7 @@ interface PrivacyWrapperProps {
   children: JSX.Element
   className?: string
   placeholder?: JSX.Element
-  type: 'video' | 'applet' | 'twingle'
+  type: 'video' | 'applet' | 'twingle' | 'h5p'
   provider: ExternalProvider
   embedUrl?: string
   twingleCallback?: () => void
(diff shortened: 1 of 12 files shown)


src/edtr-io/plugins/h5p.tsx
```diff
@@ -22,32 +38,146 @@ export const H5pPlugin: EditorPlugin<H5pPluginState> = {
...
+                    setError(
+                      'Unerlaubte Plugintypen - nutze bitte nur die vier genannten Inhaltstypen'
+                    )
+                    setMode('edit')
+                  }
+                  //console.log(json)
+                  props.state.set(url)
+                  setMode('preview')
+                } catch (e) {
+                  alert(e)
```
elbotho, line 150: could this also be handled with setError?

Entkenntnis: I will investigate this

src/components/content/privacy-wrapper.tsx
```diff
@@ -84,6 +87,7 @@ export function PrivacyWrapper({
       provider: provider,
     })
     if (isTwingle && showIframe) return null
+    if (type == 'h5p' && showIframe) return null
```
elbotho, line 90: @Entkenntnis  can we talk about `==` vs. `===` in our codebase sometime?

Entkenntnis: my current understanding: with typescript, we are actually not allowed to compare two different types, so there is already a (better) safety net in place.

I'm a bit lazy here :) so if you want to enforce `===`, you can add https://eslint.org/docs/latest/rules/eqeqeq so we are consistent

elbotho: Good point about typescript. Since the compiler should catch those it's mostly a styling thing, but most ppl seem to lean towards still using `===` (ts docs does so for example).
I'll open a PR and try the rule.

elbotho: https://github.com/serlo/frontend/pull/2180

PR #2166, opened 2023-03-04
kratos: Add SSO Button in login for NBP

hugotiburtino:
- It depends on the configuration of kratos (for now only testable using serlo/local-dev-env)
- After coming from OIDC flow, the config of kratos has to hit a route where `fetchAndPersistAuthSession` is called, otherwise the user won't know that they are logged in. For that it would be nice if frontend had an endpoint specific for OIDC
- Styling has to be improved

### DIFF
--- a/src/components/pages/auth/login.tsx
+++ b/src/components/pages/auth/login.tsx
@@ -181,9 +181,6 @@ export function Login({ oauth }: { oauth?: boolean }) {
           void router.push(flow.return_to ?? redirection)
           return
         })
-        .catch((e: Error) => {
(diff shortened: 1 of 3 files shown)


Entkenntnis:
@hugotiburtino I would like to reduce the amount of wip in the frontend. Changes look small enough that we could merge them now. Other option would be to close this PR for now (and keep the branch)

SSO is not priorities right now and I would like to only start working on it when h5p is done.

hugotiburtino:
You can merge it and also delete branch

PR #2180, opened 2023-03-07
try eslint rule eqeqeq

elbotho:


Entkenntnis (approved):
thanks, looks good

elbotho:
are you okay with this rule? 
or should we change it to `warn` for now?

Entkenntnis:
error is fine, I'm not aware of any places where we rely on `==`-behaviour (shouldn't even be possible, because of typescript)

PR #2182, opened 2023-03-07
count parents before breadcrumbs logic

### LINKED ISSUE #2181: choose main taxonomy by shortest path
use real taxonomy parent count and ignore navigation short-cuts.

This should restore old behaviour

Entkenntnis:
at least fixes https://frontend-git-2181-choose-main-taxonomy-by-shortest-path-serlo.vercel.app//1555

### DIFF
--- a/src/fetcher/create-breadcrumbs.tsx
+++ b/src/fetcher/create-breadcrumbs.tsx
@@ -120,10 +120,12 @@ export function createBreadcrumbs(uuid: MainUuidType, instance: Instance) {
     instance: Instance
   ) {
     if (taxonomyPaths === undefined) return undefined
+
+    // we select first shortest taxonomy path as main taxonomy and show it in breadcrumbs
+    // this happens directly on taxonomy parents and is not taking short-cuircuits into account
     let mainTax
     let count = -1
 
-    // select first shortest taxonomy path
     for (const path of taxonomyPaths) {
       if (!path) continue
       if (count === -1 || count > countParents(path)) {
--- a/src/fetcher/create-breadcrumbs.tsx
+++ b/src/fetcher/create-breadcrumbs.tsx
@@ -15,6 +15,17 @@ type TaxonomyTermNodes = Extract<
   { __typename: 'Article' }
 >['taxonomyTerms']['nodes']
 
+function countParents(taxonomyPath: TaxonomyTermNodes[0]) {
+  function count(current: RecursiveTree): number {
+    if (current.parent) {
+      return count(current.parent) + 1
+    } else {
+      return 0
+    }
+  }
+  return count(taxonomyPath as RecursiveTree)
+}
+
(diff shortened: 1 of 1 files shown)


src/fetcher/create-breadcrumbs.tsx
```diff
@@ -15,6 +15,17 @@ type TaxonomyTermNodes = Extract<
...
+  function count(current: RecursiveTree): number {
+    if (current.parent) {
+      return count(current.parent) + 1
+    } else {
+      return 0
+    }
+  }
+  return count(taxonomyPath as RecursiveTree)
+}
+
```
elbotho, line 28: there are some refactorings that come to mind (I think this could be a good use case for `reduce`.
But at least `return current.parent ? count(current.parent) + 1 : 0` in L20 would look nicer ;)

src/fetcher/create-breadcrumbs.tsx
```diff
@@ -109,20 +120,19 @@ export function createBreadcrumbs(uuid: MainUuidType, instance: Instance) {
...
   ) {
     if (taxonomyPaths === undefined) return undefined
-    const breadcrumbCandidates = taxonomyPaths.map((candidate) =>
-      taxonomyParentsToRootToBreadcrumbsData(candidate, instance)
-    )
-    let breadcrumbs
+    let mainTax
+    let count = -1
 
     // select first shortest taxonomy path
```
elbotho: since we forgot/misunderstood before: maybe we should explain a little bit here?

elbotho (changes requested):
see notes above.
Thanks for the fix, let's just try and see if it works in most cases :)

src/fetcher/create-breadcrumbs.tsx
Entkenntnis: Good idea

src/fetcher/create-breadcrumbs.tsx
Entkenntnis: This is everyday work, not a code golf competition ;) this is code I would teach in high school

Besides of the textbook-like look: is there anything actually unclear what this function is doing that would require a refactoring?

elbotho:
Sorry, definitely no refactoring that is required and definitely not the worst code style in the repo :) 
I'm okay with merging this and the one liner is a matter of taste I guess. 

Why I commented in the first place:
I feel like it's not easy to understand what is going on in this file (main problem is probably that we do stuff here that should happen in the API).

Another thing is that we use `let` only a number of times in the repo at all.
Again, that's a matter of taste, but I've been nudged by Kulla and Jonas to adopt a more functional style for my code here.

But in any case: Feel free to merge.

Entkenntnis:
Thanks for the explanations, this file and the navigation feature is indeed in need of a refactoring, but as long as it's holding together and is not affecting other parts of the system the refactoring is not urgent.

PR #2185, opened 2023-03-07
page partners plugin

elbotho:
a lot of work to add such a simple plugin, but the current solution (dev console to change some styles and then make a screenshot for `/partners`) was also pretty …inconvenient


Entkenntnis (approved):
cleanly done - looks good

a creative way of using plugins

PR #2192, opened 2023-03-08
lumi proxy

### LINKED ISSUE #2191: lumi proxy

Entkenntnis:
Use frontend api route as proxy, so we can safely load embedded content from lumi without exposing personal information from client. This way we can avoid using a wrapper altogether and always load h5p content directly.

works nicely so far, example https://frontend-git-2191-lumi-proxy-serlo.vercel.app/269868 (or create new one on staging)

(thinking aloud: by setting good cache headers we can also improve performance significantly without putting load on lumi, using the proxy, we can also get acceses to finishedData and get some metrics out of it - win win here)


Entkenntnis:
@elbotho merging into staging for now to allow testing, if you have any feedback, add it here and I will look through it

PR #2223, opened 2023-03-21
 Add consent checkbox and validation before submit

CodingDive:
Added consent checkbox and a simple validation before submitting

[preview](https://frontend-git-fix-signup-explicit-consent-serlo.vercel.app/auth/registration)

- [x] add other translations (added German micro copy with Gregor but the rest will be handled by crowdin as described [here](https://github.com/serlo/frontend/wiki/Internationalization) )
- [x] clean up code
- [x] design review & iteration
- [x] product review
- [x] sync & update Hugo about the addition of the newsletter


src/components/pages/auth/registration.tsx
```diff
@@ -19,6 +19,13 @@ import { useInstanceData } from '@/contexts/instance-context'
 import { replacePlaceholders } from '@/helper/replace-placeholders'
 
 export function Registration() {
+  const [isConsentCheckboxChecked, setIsConsentCheckboxChecked] =
+    useState<boolean>(false)
```
Entkenntnis: you can omit <boolean>, because type can be inferred

src/components/pages/auth/registration.tsx
```diff
@@ -19,6 +19,13 @@ import { useInstanceData } from '@/contexts/instance-context'
...
+  const [
+    hasValidationErrorMissingConsent,
+    setHasValidationErrorMissingConsent,
+  ] = useState<boolean>(false)
```
Entkenntnis: same here

Entkenntnis:
@CodingDive thanks for your first PR in the frontend, otherwise, changes look good

Entkenntnis:
@CodingDive You can replace the image with the live url: https://frontend-git-fix-signup-explicit-consent-serlo.vercel.app/auth/registration

(this way, you avoid out-dated screen-shots)

src/components/pages/auth/registration.tsx
```diff
@@ -131,7 +138,12 @@ export function Registration() {
           <Flow
             flow={flow}
             flowType={FlowType.registration}
-            onSubmit={onSubmit}
+            onSubmit={
+              isConsentCheckboxChecked
+                ? onSubmit
+                : () =>
+                    Promise.reject(setHasValidationErrorMissingConsent(true))
```
Entkenntnis: find way to fix types

CodingDive: I tried to fix it but couldn't find a good solution. The more I think about it, the more I also like rejecting a Promise here as it would allow us to catch the rejected promise within the `flow.tsx` component in the future and it'd already be clear that the submitting was non successful. Meanwhile if we were to pass `() => void`, the inner Flow component would not get that feedback.

Let me know what you think and if you think it's worth it for me to investigate a hacky way to make the types nicer.

Entkenntnis: Promise.reject is fine, maybe add a small comment why it's here (e.g. to notify the flow that an error occured) and we're good to go.

src/components/pages/auth/registration.tsx
```diff
@@ -158,27 +173,62 @@ export function Registration() {
...
+    return (
+      <div className="mt-12 serlo-p mx-0 text-base">
+        <label className="flex">
+          <input
+            className="mr-2 accent-brand scale-125"
+            style={{
+              // min-width should fix ios problem with rendering the checkbox
+              // super tiny, see
+              // https://stackoverflow.com/questions/64227252/my-checkbox-inputs-stays-ridiculously-small-on-ios-13-7
+              minWidth: '14px',
```
Entkenntnis: for consistency we use tailwind styles, e.g. `min-w-[14px]`

Entkenntnis (approved):
Thank you @CodingDive 

You can merge staging into this branch to resolve the conflict (we use merge commits), after that you can merge it into staging and test it there.

You have to add the german translation into crowdin, otherwise it's not saved and the integration will have problem.

PR #2227, opened 2023-03-22
refactor: upgrade kratos sdk version

### LINKED ISSUE #10: kratos: upgrade sdks
The image of Ory Kratos was updated to the newest state (March 7, 6d83dc9). Let's test it out and upgrade the sdks, as a preparation for implementing SSO.

- [ ] First, test all auth flow (just to be sure it's working in your machine)
- [ ] Open branch in frontend, upgrade ory sdk packages there
- [ ] Open branch in api, upgrade ory sdk packages there
- [ ] Test thoroughly all auth flows
- [ ] If everything ok, open PRs in frontend and api

Latest version of SDK
https://www.npmjs.com/package/@ory/client 1.1.21
Breaking changes
https://www.ory.sh/docs/guides/upgrade/sdk-v1

*All auth flows* (test please only happy paths)
- registration 
- verification (go to localhost:4436 to see verification email)
- password reset
- login
- logout

*Important Note*:  No need to test SSO for now. Comment out lines 25-34. We have to improve the frontend code first, since clicking on Anmelden (normal login) leads us to SSO Login.

AndreasHuber:
See https://github.com/serlo/local-dev-env/issues/10
See also https://github.com/serlo/api.serlo.org/pull/826.

The upgrade of kratos implies breaking changes, that's the reason of this PR.


src/components/pages/auth/login.tsx
```diff
@@ -63,7 +60,7 @@ export function Login({ oauth }: { oauth?: boolean }) {
     // Currenty not in use
     if (flowId) {
       kratos
-        .getSelfServiceLoginFlow(String(flowId))
+        .getLoginFlow({ id: flowId as string })
```
hugotiburtino: We should try to avoid `as` (type assertion), since we are trying to trick the compiler with it by not making our code type safe (just in some cases it is wanted). See v.g. [this article](https://www.bytelimes.com/why-you-should-avoid-type-assertions-in-typescript/).
The previous solution was not that better (converting into string, with String object), since it could lead to bugs. The best solution would be to make a type check.
But since we are about to get rid of this whole block, it is not that important anyway.

hugotiburtino (approved):
Thanks, LGTM

Entkenntnis:
@hugotiburtino Works on localhost for me. You can merge it into staging.

Entkenntnis (approved):
thank you andreas

PR #2238, opened 2023-03-24
fix content-only link bug

### LINKED ISSUE #2233: content only: links are not opened in new tab
example: https://de.serlo.org/mathe/5/mathe?contentOnly

grafik

"Zahlen und Größen" opens in new tab, but "Grundrechenarten" is not.

The intended behaviour defined in `Link` is that all links are opend in new tabs.

Wasn't able to find an obvious problem (link component is used, check is existent)

@kulla

Entkenntnis:


Entkenntnis:
merging into staging, as the bug can't be reproduced locally

elbotho:
Nice, thank you!


> merging into staging, as the bug can't be reproduced locally

it should work by using the`/content-only/` route directly (not `?contentOnly=`)?

Entkenntnis:
the problem is that part of the logic is handled by the cloudflare worker (the redirect) and the bug only happens if the url is redirected

edit: yes, using /content-only is working, also before this fix.

Issue #2254, opened 2023-04-03
editor: fix bug on load


Entkenntnis:
This problem is occuring when loading the preview

![grafik](https://user-images.githubusercontent.com/13507950/229500464-250c55ac-c53d-4eba-8c37-0980e5a6eecd.png)

hejtful:
The issue is caused by the difference between the previously used and current Slate version. In the old version, empty `Element` nodes were tolerated, while in the new version "[All `Element` nodes must contain at least one `Text` descendant](https://docs.slatejs.org/concepts/11-normalizing#built-in-constraints)".

The correct approach to this issue seems to be to [deserialize](https://docs.slatejs.org/concepts/10-serializing#deserializing) the nodes coming from the database and fill the empty `Element` nodes with `Text` nodes.

PR #2257, opened 2023-04-03
fix(editor text plugin): deserialize empty elements

hejtful:
https://github.com/serlo/frontend/issues/2254

### DIFF
--- a/src/serlo-editor-repo/plugin-text/index.tsx
+++ b/src/serlo-editor-repo/plugin-text/index.tsx
@@ -19,6 +19,7 @@ import type {
   TextEditorState,
 } from './types'
 import { emptyDocumentFactory } from './utils/document'
+import { isEmptyObject } from './utils/object'
 
 /**
  * @param config - {@link TextEditorConfig | Plugin configuration}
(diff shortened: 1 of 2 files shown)


Entkenntnis (approved):
thank you, can confirm that this fixes the bug

src/serlo-editor-repo/plugin-text/index.tsx
```diff
@@ -35,6 +36,17 @@ const createTextPlugin = (
       return value
     },
     deserialize(value) {
+      // If the first child of the Element is an empty object,
+      // replace it with an empty Text node
+      // https://docs.slatejs.org/concepts/11-normalizing#built-in-constraints
+      const firstChild = (value[0] as CustomElement).children[0]
+      if (isEmptyObject(firstChild)) {
+        return {
+          value: [{ children: [{ text: '' }] } as CustomElement],
```
Entkenntnis, line 45: @hejtful What happens if `value` has more than one element? Will this case never happen? Because it looks like all remaining children are discarded here.

hejtful: @Entkenntnis Slate automatically adds this empty text node in children in the new version, so the new articles are safe.

The condition on line 43 will only be true in cases where an empty text editor was saved with the old version, therefore we don't expect any more children.

Or do you see something I missed?

PR #2259, opened 2023-04-03
Fix/legacy converter

### LINKED ISSUE #2256: editor: loading old revisions is not working
e.g. https://frontend-git-2252-move-edtr-io-source-code-into-th-a83fc7-serlo.vercel.app/entity/repository/add-revision/74888/123513

I have removed https://github.com/Entkenntnis/tmp-serlo-legacy-editor-to-editor because the text-plugin has removed support for slate migration serliazer.

Unfortunately, it looks like that we still have a lot of revisions in legacy format...

~will fix this by including relevant source code into the repo as well~

will use another approach

Entkenntnis:
fixes #2256 

Some typescript errors have to be fixed


Entkenntnis:
this approach is not quite viable, trying another approach

Issue #2255, opened 2023-04-03
editor: upgrade inline-input


Entkenntnis:
The inline-input was using an older version of slate, see https://github.com/serlo/frontend/pull/2253/files#diff-380499a67e983e544050cdd6cd9763c5b40657b12129a0ad3ceff902f3e25512

Slate is needed because we want to properly position the hovering overlay, therefore an normal input is not enough.

Entkenntnis:
some more context: this input is used in the sources-section of an article, here you can see the previous version:

![grafik](https://user-images.githubusercontent.com/13507950/229506076-632ec810-bd4a-4dd2-926a-52169d41d238.png)

PR #2265, opened 2023-04-04
fix(inline-input helper): add new version of Slate

hejtful:
https://github.com/serlo/frontend/issues/2255


Entkenntnis (approved):
thank you for the fix, can confirm that it works

Issue #2260, opened 2023-04-03
editor: redo throws error


elbotho:
steps: 
- open any [content](http://localhost:3000/entity/repository/add-revision/1555)
- select text
- make bold
(maybe select something else)
- undo (works)
- redo: throws on of those errors:
`Error: Cannot resolve a DOM point from Slate point: {"path":[0,1],"offset":10}`
`Cannot get the leaf node at path [0,5] because it refers to a non-leaf node: …`

PR #2268, opened 2023-04-04
fix(plugin-text): update referenced value when updating Slate state

hejtful:
https://github.com/serlo/frontend/issues/2260


Entkenntnis (approved):
thank you, nice fix, works now

Issue #2262, opened 2023-04-03
editor: creating inline-formula makes cursor jump into line above


elbotho:
(see title).
Instead the math formula should be selected and ready to edit.
<img width="144" alt="image" src="https://user-images.githubusercontent.com/1258870/229536626-c7fa94a4-741a-4c07-be0c-ff21e79ff081.png">

PR #2269, opened 2023-04-04
fix(plugin-text): manually position cursor after inserting math node

hejtful:
https://github.com/serlo/frontend/issues/2262


elbotho (approved):
Works great, thank you 🎉

Entkenntnis (approved):
nice work

PR #2270, opened 2023-04-04
format editor error msg

elbotho:
Not sure if this relies too much on frontend code, but it definitely looks better :)

### DIFF
--- a/src/serlo-editor-repo/core/sub-document/index.tsx
+++ b/src/serlo-editor-repo/core/sub-document/index.tsx
@@ -1,4 +1,3 @@
-import { faWarning } from '@fortawesome/free-solid-svg-icons'
 import * as React from 'react'
 
 import { PluginProps } from '../../internal__plugin-state'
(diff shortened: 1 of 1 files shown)


Entkenntnis:
Hate to say it, but yes, this relies too much on frontend code :see_no_evil: 

To keep it decoupled, you would need to:
- rewrite the component in the folder `serlo-editor-repo`
- avoid tailwind and write all styles with styled-components (as css object)

... which is an unreasonable amount of work to do for such a change :sweat_smile: 

OK, so the rule for now would be: Avoid importing any frontend code within the folder `serlo-editor-repo`. Otherwise it's difficult to reuse this folder without the whole frontend as a dependency.

I hope we will find a good solution for RLP, so we can maybe loosen this rule - this will happen after the release so I'll ask for a bit of patience :pray: 

(and yes, using frontend stuff in edtr-io would make a lot of things easier)

elbotho:
@Entkenntnis removed dependency, used tailwind :)

Entkenntnis (approved):
edtr-io meets tailwind, it's now official

Issue #2271, opened 2023-04-05
editor: unselect spaces around word before applying link and inline code


Entkenntnis:
it makes it quite hard to get out of the link/inline code in certain situations + we don't want to have the whitespace as part of the link title.

It should be quite similiar to the fix for bold/italic

Previous version was trimming the whitespace

hejtful:
Note: applied the same for math as well, to make it consistent with the previous version.

PR #2272, opened 2023-04-05
fix(plugin-text): trim selection before creating math, code or link elements

hejtful:
https://github.com/serlo/frontend/issues/2271

- `toggleMath` util also restructured to minimize nesting using early returns.


Entkenntnis (approved):
thank you, now it works nicely

PR #2273, opened 2023-04-05
fix(plugin-text): make stepping out of inlines more explicit

hejtful:
- Stepping out of a Slate inline with arrow keys will not step over the next character
- Pass the config object to the math formula and editor


Entkenntnis (approved):
thanks for fixing this

PR #2274, opened 2023-04-05
fix(plugin-text): hovering toolbar submenu behavior

hejtful:
- When selecting a heading, the hovering toolbar will reset to the root menu, closing the headings sub-menu
- When selecting a color, the same will happen as above, + the selection will collapse to the end

Both behaviors are taken from the current live version.


Entkenntnis:
@hejtful Thank you for the fix. As we are already fixing whitespace issues: I found a small regression with colors, the whitespace is not trimmed here as well, so the cursor is jumping to the start of the next word.

Probably we could trim whitespace before changing colors as well.

Entkenntnis (changes requested):
see previous comment

hejtful:
@Entkenntnis Good catch, updated

Entkenntnis (approved):
thanks

PR #2280, opened 2023-04-06
restore previous behavior of cursor at end of link

### LINKED ISSUE #2276: editor: find and implement the most useful behavior for stepping in and out of links
When the user selects a link and tries to step out of it using the arrow keys, we can:
1. Step out of the link **and** over the next character/empty space. This makes for a smoother flow through the text and hides the difference between a link and normal text from the user.
2. Step out of the link, **but** not over the next character/empty space. This allows for more control, as the user can choose to continue typing inside the link, or step just outside the link and then continue typing.
3. Use the first approach when there is an empty space outside the link, otherwise use the second approach.
4. Go for the first approach, but when the cursor is at the edge of the link, make the link unselected.
5. Something else?

Note: There is a relevant Chromium bug which makes the behavior of the second approach inconsistent between Firefox and Chrome. Here is how to fix it: https://github.com/ianstormtaylor/slate/blob/main/site/examples/inlines.tsx#L241

Entkenntnis:
@hejtful don't like to keep issues open for too long, so I attempted a fix

### DIFF
--- a/src/serlo-editor-repo/plugin-text/components/inline-chromium-bugfix.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/inline-chromium-bugfix.tsx
@@ -1,5 +1,0 @@
-export const InlineChromiumBugfix = () => (
-  <span contentEditable={false} style={{ fontSize: 0 }}>
-    ${String.fromCodePoint(160) /* Non-breaking space */}
-  </span>
-)
--- a/src/serlo-editor-repo/plugin-text/components/inline-chromium-bugfix.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/inline-chromium-bugfix.tsx
@@ -1,0 +1,5 @@
+export const InlineChromiumBugfix = () => (
+  <span contentEditable={false} style={{ fontSize: 0 }}>
+    ${String.fromCodePoint(160) /* Non-breaking space */}
+  </span>
+)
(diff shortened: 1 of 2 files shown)


hejtful:
Works like a charm for keyboard arrow keys navigation!

When clicking your mouse just right of the link, the link is still selected. In the old version, the link was not selected in this case. Would you like to address this case in this PR as well?

Entkenntnis:
To be sure we are talking about the same thing:

- If I click left of the yellow line, the cursor is but on the edge of the link and the link is selected.
- If I click right of the yellow line, the cursor is on the edge but the link is not selected.

grafik

I have tested it on staging and this PR and as far as I see the behavior is the same for both (you have to be quite precise, though) -> there is no regression for me.

UPDATE

It works in FF, but in Chrome, you can't unselect the link by clicking behind it, not in production and not here - and keyboard navigation is broken as well

-> So this PR is improving the situation and not causing regressions, the problem you described is already present. -> fix later

(sorry for the messy text, it's becoming a stream of consiousness ...)

hejtful:
@Entkenntnis To avoid lengthy discussions, I just pushed a commit with a fix for this. If you'd like me to explain the bug or the fix, please let me know.

hejtful (approved):

Entkenntnis:
@hejtful I have tested your fix, unfortunately it's not working for me: I'm still not able to unselect the link with chromium (v108). In addition, the text plugin crashes some times in chromium. Not able to 100% reproduce, but it involves adding and removing spaces at the end and happens after some tries.

For reference, the branch with the attempted fix: https://frontend-9er7x6hqu-serlo.vercel.app/entity/repository/add-revision/74888

PR #2290, opened 2023-04-12
Fix image pasting undo

LarsTheGlidingSquirrel:
I found another minor bug. 

Steps to replicate: 
- Create a paragraph with two lines of text. Example:
```
one
two
```
- Insert image or video with cursor at the end of `one`
- Undo changes

Now you only see `one`. However, when clicking on the editor the second line reappears. 

See notes in this commit for what I found out so far.


Entkenntnis:
#2291 fixes it for me.

PR #2291, opened 2023-04-12
fix/undo image paste

hejtful:
**Problem**: Undo doesn't work as expected after image paste.

**Status quo**: The `useEffect` hook was firing after the render, therefore not updating the editor state on time.

**Fix**: The `useMemo` hook doesn't wait for the render to be complete, and updates the editor on time.

Additionally, added the README explaining some technical decision related to plugin-text.

### DIFF
--- a/src/serlo-editor-repo/plugin-text/README.md
+++ b/src/serlo-editor-repo/plugin-text/README.md
@@ -1,0 +1,84 @@
+# Text plugin (`TextEditor`)
+
+## Table of contents
+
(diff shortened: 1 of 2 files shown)


Entkenntnis:
@hejtful Changes look good 👍 

@LarsTheGlidingSquirrel Can you test this PR? Is it the same bug as in https://github.com/serlo/frontend/pull/2290?

Entkenntnis (approved):
yes, it fixes the described bug

PR #2297, opened 2023-04-13
fix/latex editor copy paste bug

hejtful:
This PR's scope is bigger than it should be, sorry about that.

- The first commit adds the theme and i18n values to `MathElement`, which passes it to the `plugin-math`.
- The second commit removes the unused `clipboard` Redux store module. I found this module while looking for the cause of the math editor copy/paste bug, but then I realized it's not used anywhere.
- **The third commit is the actual bugfix.** In the old version, the `copy` event of the `MathEditorTextArea` did not propagate all the way to the `Editor`. In the new version, the `copy` event of the `MathEditorTextArea` propagates all the way to the `Editable` component. This PR explicitly stops that propagation.


Entkenntnis (approved):
thank you, this fixes the issue.

You should now be able to merge PR yourself.

PR #2298, opened 2023-04-13
chore/resolve deprecated todos and eslint issues

hejtful:
Note: API extractor was removed when moving the editor code from edtr-io to frontend, therefore workaround for it can be removed.


Entkenntnis (approved):
thanks, looks good

PR #2301, opened 2023-04-13
fix(text-plugin): Auto focus created text plugin

LarsTheGlidingSquirrel:


LarsTheGlidingSquirrel:
I did not see a way to use `autofocusRef` like in other plugins. But `ReactEditor.focus(editor)` seems to be the recommended way to change focus in Slate. 

Testing still due.

Entkenntnis (changes requested):
overlaps with #2302

Entkenntnis:
@LarsTheGlidingSquirrel Can we close this PR? It looks like the previous PR already contains all fixes.

PR #2302, opened 2023-04-13
feat(plugin-text): handle hotkeys

hejtful:
- Insert a new text plugin on enter key press
- Focus the newly inserted text plugin (implementation taken from https://github.com/serlo/frontend/pull/2301 and tweaked a bit to fit for inserting with enter as well)
- Prevent adding line breaks if `noLinebreaks` config flag is active
- Also fixes a bug with inserting an image when selection is at the end (where a text plugin would be inserted after the inserted image)
- Remove an empty text plugin on backspace or delete key press
- Merge text plugin with previous text plugin on backspace if selection at start
- Merge with next text plugin on delete if selection at end

### DIFF
--- a/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
@@ -31,9 +31,11 @@ import {
 } from '../types'
 import {
   emptyDocumentFactory,
+  mergePlugins,
   sliceNodesAfterSelection,
 } from '../utils/document'
 import { markdownShortcuts } from '../utils/markdown'
+import { isSelectionAtEnd, isSelectionAtStart } from '../utils/selection'
 import { HoveringToolbar } from './hovering-toolbar'
 import { LinkControls } from './link-controls'
 import { MathElement } from './math-element'
@@ -96,12 +98,11 @@ export function TextEditor(props: TextEditorProps) {
     // ReactEditor.focus(editor) does not work without being wrapped in setTimeout
     // See: https://stackoverflow.com/a/61353519
     setTimeout(() => {
-      Transforms.select(editor, { offset: 0, path: [0, 0] })
+      const pointAtStart = { offset: 0, path: [0, 0] }
+      Transforms.select(editor, selection || pointAtStart)
       ReactEditor.focus(editor)
     })
-    // Editor should be manually focused only on first render
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [editor, focused])
 
   function handleEditorChange(newValue: Descendant[]) {
     const isAstChange = editor.operations.some(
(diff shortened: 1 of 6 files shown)


elbotho (approved):

Entkenntnis (approved):
looks good! you can merge it into staging

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -116,6 +116,12 @@ export function TextEditor(props: TextEditorProps) {
   }
 
   function handleEditableKeyDown(event: React.KeyboardEvent) {
+    const isEnterKey = isHotkey('enter', event)
+
+    if (config.noLinebreaks && isEnterKey) {
```
Entkenntnis: :+1:

Entkenntnis (approved):
I have tested it a while and wasn't able to break it

The comments in the code are very helpful :+1:

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -97,12 +120,14 @@ export function TextEditor(props: TextEditorProps) {
   }
 
   function handleEditableKeyDown(event: React.KeyboardEvent) {
-    /*
-      Special handler for links. If you move right and end up at the right edge of a link,
-      this handler unselects the link, so you can write normal text behind it.
-    */
+    if (config.noLinebreaks && event.key === 'Enter') {
+      event.preventDefault()
+    }
```
LarsTheGlidingSquirrel, line 125: As far as I understood it, when pressing 'enter', one of these should happen: 
noLineBreaks: false --> Simply insert `\n` and skip a bunch of code. (maybe create a new list element)
noLineBreak: true --> Create new text-plugin, split text, ...

Maybe the code could reflect that in a `handleEnter` function:
```
function handleEnter(...) {
  if (!noLineBreaks) return
  
  ... // event.preventDefault & do all the transformations
}

LarsTheGlidingSquirrel (approved):
Looks good! Just added one suggestion to make the code more clear.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: I understood it differently, from how `noLinebreaks` is used in the code:
- `noLinebreaks === true` --> `Enter` key does nothing
- `noLinebreaks === false` --> `Enter` key adds new lines or new text-plugins etc

An example of where `noLinebreaks` is used is the title of the box plugin.

elbotho: Agree. `noLinebreaks ` should make the plugin "inline" with no possibility to create linebreaks or other block level elements (so also no list elements).

Issue #2299, opened 2023-04-13
SSO: Put single sign-on into an experiment


CodingDive:
SSO should be disabled by default until we need it. See discussion https://github.com/serlo/frontend/issues/2292#issuecomment-1506963969 and https://serlo.slack.com/archives/CEB781NCU/p1681308459785569?thread_ts=1681304582.199229&cid=CEB781NCU

Note that experiments the way they are currently implemented might not be a viable solution. See comment https://github.com/serlo/frontend/pull/2303#issuecomment-1521682746

elbotho:
I guess we don't need this any more.

PR #2303, opened 2023-04-13
Change styling of SSO flow

### LINKED ISSUE #2282: SSO: styling improvements
We can test now the SSO with NBP. It's already deployed in staging. In frontend I found so far the following problems:

- [x] At `auth/login` after hiting 'Enter' you go always to the OIDC flow, even if you want to use the normal login. Find a better solution.
- [x] At `auth/registration` you see the duplication of the consent and 'Sign In with NPB account' is just 'Weiter'. Repetition not needed.
- [x] 'Sign In with NBP Account' occurs also in the German instance (it should be already translated).
- [x] Nice to have: just a better styling

Please check if we need to refactor `components/auth/node.ts` and `components/auth/flow.ts` before going on with this issue.

The most important thing here: It may not change the behavior at production, which is not going to have social login for now!

### LINKED ISSUE #2292: Have hardcoded order in the form for the registration
The order of elements in the staging environment changed after enabling OIDC

### Order of elements in staging
See https://de.serlo-staging.dev/auth/registration, Note that `Passwort` is in the beginning

2023-04-12-105620_830x926_scrot

### Order of elements at production
See https://de.serlo.org/auth/registration
2023-04-12-105608_839x870_scrot

### Expected behavior
Implement a fixed order of elements in the registration form

CodingDive:
Fixes #2292 and most of #2282

Thank you to @elbotho for the suggestion of modifying the element order. I tried to make it as resilient as I could by defining an enum. I think this should be fixed in Kratos though and we shouldn't ideally have the need to change anything on the frontend or modify the order.

Need to figure out how I can make the SSO button get clicked. Right now it triggers the submit of the whole form.

- [x] clean up code
- [x] pass correct onSubmit to the `button` inside the `flow.tsx` or `node.tsx` component (unsure as of now on how to achieve this)


src/components/auth/node.tsx
```diff
@@ -65,13 +65,40 @@ export function Node({
...
-        })
-        return null
+      case 'button': {
+        const label = node.meta.label?.id
+          ? getKratosMessageString(
+              node.meta.label.id,
+              strings.auth.messages,
+              'Undefined'
+            )
+          : 'Could not find translation'
```
elbotho: `'Undefined'` as a fallback makes no sense here. Maybe use `node.meta.label.text`?
Also the label should not contain an error message. I guess if `node.meta.label` is actually not defined we should not render a button at all (any optinally trigger a sentry error).

src/components/auth/node.tsx
```diff
@@ -65,13 +65,40 @@ export function Node({
...
+            <button
+              className="text-xl serlo-button-blue block w-full py-2 mt-10"
+              name={attributes.name}
+              onClick={(e) => {
+                void onSubmit(e, (attributes as { value: string }).value)
+              }}
+              value={(attributes.value as string) || ''}
+              disabled={attributes.disabled || disabled}
+            >
+              {isLoading ? (
```
elbotho: `isLoading` currently results in both buttons showing the animation which is a bit confusing.
Not sure what's the best way to solve that.

src/components/pages/auth/registration.tsx
```diff
@@ -32,9 +61,33 @@ export function Registration() {
 
   const reorderAndSetFlow = (origFlow?: RegistrationFlow) => {
     if (!origFlow) return
-    const { nodes: n } = origFlow.ui
-    origFlow.ui.nodes = [n[0], n[1], n[3], n[2], ...n.slice(4)]
-    setFlow(origFlow)
+    const reorderedNodes = origFlow.ui.nodes
```
elbotho, line 37: did you check back if we can solve this at least partly in kratos?

src/components/pages/auth/registration.tsx
```diff
@@ -32,9 +61,33 @@ export function Registration() {
...
+          console.warn('Unexpected name attribute value found')
+          return 0
+        }
+
+        return aIndex - bIndex
+      })
+    const newFlow = {
+      ...origFlow,
+      ui: { ...origFlow.ui, nodes: reorderedNodes },
+    }
```
elbotho, line 43: basically just taste. It's  more fp but also more code.

src/components/pages/auth/utils.ts
```diff
@@ -6,6 +12,25 @@ export const logoutUrl = '/auth/logout'
 export const recoveryUrl = 'auth/recovery'
 export const VALIDATION_ERROR_TYPE = 'validation-error'
 
+export const changeButtonTypeOfSSOProvider = (uiNode: UiNode) => {
```
elbotho: this is just annoying that kratos flow makes this necessary…

src/data/de/index.ts
```diff
@@ -375,7 +375,8 @@ export const instanceData = {
         code1010002: 'Sign in with NBP Account',
         code1010013: "Weiter",
         code1040001: "Account anlegen",
-        code1040003: "Weiter",
+        code1040002: 'Registriere mit NBP Account',
```
elbotho: this will be removed in i18n. add to en file as well.

elbotho (changes requested):
Thanks @CodingDive, added some questions and suggestions.

src/components/pages/auth/utils.ts
```diff
@@ -1,3 +1,9 @@
+import {
+  UiNode,
+  UiNodeInputAttributes,
+  UiNodeInputAttributesTypeEnum,
+} from '@ory/client'
```
Entkenntnis: putting these imports in this file is increasing bundlesize by 18kb (unfortunately, tree shaking is not able to shake code across different import trees)

Solution would be to move changeButtonTypeOfSSOProvider into a separate file and making it a separate module.

src/components/pages/auth/registration.tsx
CodingDive: Yeah, it's a little bit more resilient to changes from kratos in the future with a little bit more time complexity and code. I believe it's a good trade off, at least one can see the element order from the enum

elbotho: Oh sorry, I just meant the actually marked lines. You could also modify the original object above and then you would not need to spread and modify here.

src/components/auth/node.tsx
CodingDive: I disabled the loading spinner for the SSO provider button

src/components/pages/auth/registration.tsx
CodingDive: I'll talk with Hugo about this

elbotho:
@CodingDive as of now this would block deploys. 
Do you want to take on https://github.com/serlo/frontend/issues/2299 as well or should @Entkenntnis or me do that for now?

CodingDive:
> @CodingDive as of now this would block deploys. Do you want to take on #2299 as well or should @Entkenntnis or me do that for now?

@elbotho the experiments only work for logged in users as far as @Entkenntnis and I can tell. Once you're already logged in, it'd be quite hard to test SSO. Maybe we need another "experiment" solution for this that's not tied to user accounts.

I think it shouldn't block deploys because the "provider" UI element is coming from Kratos and production has a different configuration than staging. As a consequence, the button shouldn't appear in production, only staging (for now).

@Entkenntnis would like to try to deploy it tomorrow evening and see if there are any issues. I'll also be available to fix them.

Entkenntnis:
@CodingDive You can merge this branch into staging after checks are successful.

Please check crowdin and add german translations there as well - you can test it on staging then.

@elbotho As far as I see this should not change production and only difference in staging is an additional button.

elbotho:
Ah nice, if it actually does not change production behaviour we can probably skip that then.


> @elbotho the experiments only work for logged in users as far as @Entkenntnis and I can tell. Once you're already logged in, it'd be quite hard to test SSO. Maybe we need another "experiment" solution for this that's not tied to user accounts.

While the experiments interface is only visible for logged in users the features are actually stored in cookies and don't depend on the user-acc. So if we need this in the future we could either ask people to login once with a regular acc and activate the feature or make the experiment interface available outside of the own profile.

src/components/pages/auth/ory-transforms.tsx
elbotho, line 1: Since this does not use jsx this should be a `ts` file. 
Also: we have a convention where we do not export more than one thing from one file if possible and name the file like the exported function/component (with some exceptions for e.g. helper/utils files).

Maybe this code could  be part of `@/components/pages/auth/utils.ts`?

(backlog: maybe move utils.ts to `@/auth/`)

elbotho (approved):
Thank you @CodingDive, added one more follow up, but I'm fine with testing this in staging.

src/components/pages/auth/ory-transforms.tsx
CodingDive: The reason why we put it in a separate file other than utils was because the bundle size got too big due to an import of @ory/client. I can rename the file to 'ory-helper'?

elbotho: I see, good that you checked bundle size.
`ory-helper.ts` or `kratos-helper.ts` would be fine for now then. thanks.

src/components/auth/node.tsx
```diff
@@ -72,22 +72,13 @@ export function Node({
...
         const label = node.meta.label?.id
           ? getKratosMessageString(
               node.meta.label.id,
               strings.auth.messages,
-              node.meta.label?.text || fallbackLabel
+              node.meta.label?.text ?? strings.auth.messages.code1010002
             )
-          : fallbackLabel
-        if (!label || label === 'Could not find translation') {
-          triggerSentry({
```
elbotho: Missing translations are handled in `getKratosMessageString` already. 
I think it's save enough to trust that label is actually set (my guess is that it's just not an accurate type by kratos here).

PR #2306, opened 2023-04-17
a11y: color contrast

elbotho:
- improved `<ol>` contrast and positioning
- significantly darker header menu text color (and less layout shift between auth/non-auth)
- additional slightly darker blue for smaller texts (e.g. links) and when blue is used on light blue (e.g. buttons)
- more contrast in exercise input field (placeholder)

image
image
image
image


Entkenntnis (approved):
a good solution without changing CI

PR #2312, opened 2023-04-18
Disable markdown shortcuts in box title, image caption & certain table cells

LarsTheGlidingSquirrel:
In this PR, markdown shortcuts will no longer work inside box title, image caption and certain table cells (top-row, left-column or both)

This reproduces the current behavior in production.


Entkenntnis (approved):
Thank you, changes look good and clean.

PR #2313, opened 2023-04-18
feat: import content from uuid

elbotho:
only for newly created entities.

examples:
- https://frontend-git-load-any-revision-serlo.vercel.app/entity/create/Article/200247
- https://frontend-git-load-any-revision-serlo.vercel.app/entity/create/Exercise/23869

needs testing
- [x] applet
- [x] article
- [x] course
- [x] course page
- [x] event
- [x] exercise group
- [x] exercise
- [x] solution (if we find a bug here I just deactivate it)
- [x] video

image

Notes:
- ~course: works but only imports title so might not make sense ⚠️~ (deactivated for course but:)
- ~coursePage: no result ⚠️~ (CoursePage can now be imported when editing a course)
- ~exercise: imports task and feedback but not solution (ok)~ (Solutions can no be imported if needed)
- exercise group: imports task but no child exercises (but manually importing a child ex works fine!) (ok)
- meta_title is still set (ok)

TODO:
- [x] fix course page import
- [x] discuss with #u_begleitung
- [x] code review
- [ ] test on staging


Entkenntnis (approved):
code looks clean and a nice reuse of existing components! + not distracting the UI.

PR #2315, opened 2023-04-19
refactor/merge formatting options

hejtful:
1. Rename `controls` to `formattingOptions` (and `useControls` to `useFormattingOptions` etc)
2. Merge markdown utils into `useFormattingOptions` hook
3. Use `formattingOptions` config to determine if a markdown shortcut should be handled (and remove `disableMarkdownShortcuts` config flag)
4. Merge handling lists shortcuts into `useFormattingOptions` hook
5. Fix a bug where pressing the enter key, when selection is at the end of a list, which is at the end of the editor, would break the list and create a new text plugin (now it just creates a new list item, like on production)


LarsTheGlidingSquirrel (approved):
Nice! Code now communicates more clearly what it is doing. Just added a little bit more info to the readme.

Entkenntnis (dismissed):
~for some reason that I haven't found yet, this PR increases bundle size of the frontend main route `[...slug]` by 19kb.~

~Needs to be invested.~

~update: checking build output, bundle size looks good, could be a problem with the CI, checking ...~

solved, not related to this PR

PR #2323, opened 2023-04-20
refactor link component

### LINKED ISSUE #2310: cleanup link component
remove some code for legacy serlo.

elbotho:
closes #2310 

@Entkenntnis I removed some `path` code since we don't use it currently. Okay with you?

### DIFF
--- a/src/components/comments/meta-bar.tsx
+++ b/src/components/comments/meta-bar.tsx
@@ -43,7 +43,6 @@ export function MetaBar({
       <UserLink
         user={user}
         withIcon
-        path={['comment-user', id]}
         className={clsx(
           'serlo-button text-brand text-lg font-bold',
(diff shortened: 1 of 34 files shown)


Entkenntnis:
removing `path` is fine

elbotho:
@Entkenntnis Should I remove it everywhere then?

Entkenntnis:
I think it was useful for e.g. revision view? Then we maybe need an alternative...

Edit: I'm not currently using it, so it's safe to remove.

Entkenntnis (approved):
Thank you for the refactoring, code looks good.

Some manuell testing an staging would be good here

elbotho:
I remove it where we don't use it atm.

Entkenntnis (approved):
thanks

PR #2324, opened 2023-04-20
single logout: revalidate kratos session

elbotho:
revalidates the kratos session from server:
- on tab change
- on coming online again
- on first opening a page
- not in production

If we stick to this we can probably simplify the tabchange and online check in the future.

### DIFF
--- a/src/auth/auth-provider.tsx
+++ b/src/auth/auth-provider.tsx
@@ -3,7 +3,9 @@ import type { AuthorizationPayload } from '@serlo/authorization'
 import { createContext, ReactNode, useEffect, useState } from 'react'
 
 import { AuthSessionCookie } from './cookie/auth-session-cookie'
+import { fetchAndPersistAuthSession } from './cookie/fetch-and-persist-auth-session'
 import type { createAuthAwareGraphqlFetch } from '@/api/graphql-fetch'
+import { isProduction } from '@/helper/is-production'
 
 export type AuthenticationPayload = {
   username: string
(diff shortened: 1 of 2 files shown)


Entkenntnis (approved):
thanks, can be tested on staging

elbotho:
@hugotiburtino and @kulla merged for testing on staging.

src/auth/auth-provider.tsx
```diff
@@ -3,7 +3,9 @@ import type { AuthorizationPayload } from '@serlo/authorization'
 import { createContext, ReactNode, useEffect, useState } from 'react'
 
 import { AuthSessionCookie } from './cookie/auth-session-cookie'
+import { fetchAndPersistAuthSession } from './cookie/fetch-and-persist-auth-session'
```
Entkenntnis, line 6: @elbotho I have overlooked the bundle size analyzer.

Including this function will load the whole kratos client for every route.

We need to lazy load kratos here ... I would suggest an `import()` here:

https://nextjs.org/docs/advanced-features/dynamic-import#with-external-libraries

elbotho: Hmm I see. 
The trouble is, that we'd actually use the call on all routes.
At least for probably logged in users (and on staging for now).

Entkenntnis: Dynamic import should exclude it from first load JS, which will keep hydration fast and defer it a bit.

It's ok that it's used on all routes. As far as I understood it's only loaded for users that have an active session? Only these users will load the additional JS.

elbotho: nice, if everything goes well it should not even cause additional js loading, since kratos is needed for logging in / out anyway.

PR #2329, opened 2023-04-20
add rp initiated logout redirect

elbotho:
https://serlo.slack.com/archives/C04TTH20JP5/p1682012733496979

### DIFF
--- a/src/components/pages/auth/logout.tsx
+++ b/src/components/pages/auth/logout.tsx
@@ -31,10 +31,17 @@ export function Logout({ oauth }: { oauth?: boolean }) {
   useEffect(() => {
     checkInstance({ redirect: false })
 
-    const redirection = filterUnwantedRedirection({
-      desiredPath: sessionStorage.getItem('previousPathname'),
-      unwantedPaths: [settingsUrl, loginUrl, registrationUrl],
-    })
+    const cookieSession = AuthSessionCookie.parse()
+    const isSSO = cookieSession?.authentication_methods
+      ? cookieSession.authentication_methods[0].method === 'oidc'
+      : false
+
+    const redirection = isSSO
+      ? 'https://aai-dev.nbpdev.de/realms/nbp-aai/protocol/openid-connect/logout'
+      : filterUnwantedRedirection({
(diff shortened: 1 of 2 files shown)


src/components/pages/auth/logout.tsx
```diff
@@ -31,18 +31,25 @@ export function Logout({ oauth }: { oauth?: boolean }) {
   useEffect(() => {
     checkInstance({ redirect: false })
 
-    const redirection = filterUnwantedRedirection({
-      desiredPath: sessionStorage.getItem('previousPathname'),
-      unwantedPaths: [settingsUrl, loginUrl, registrationUrl],
-    })
+    const cookieSession = AuthSessionCookie.parse()
+    const isSSO = cookieSession?.authentication_methods
```
kulla, line 35: We can use `io-ts` here (can also be a follow up :smile: )

src/components/pages/auth/logout.tsx
```diff
@@ -31,18 +31,25 @@ export function Logout({ oauth }: { oauth?: boolean }) {
   useEffect(() => {
     checkInstance({ redirect: false })
 
-    const redirection = filterUnwantedRedirection({
-      desiredPath: sessionStorage.getItem('previousPathname'),
-      unwantedPaths: [settingsUrl, loginUrl, registrationUrl],
-    })
+    const cookieSession = AuthSessionCookie.parse()
```
kulla, line 34: Is `AuthSessionCookie.parse()` the same as `AuthSessionCookie.get()`?

kulla (approved):

src/components/pages/auth/logout.tsx
elbotho: - `get` returns a `string` 
- `parse` returns a `Session` object (and you can optionally supply a cookie yourself)
But the difference is basically just a `JSON.parse`

src/components/pages/auth/logout.tsx
elbotho: Just for line 35-36?
I'd only use it if it's as short tbh :)

PR #2339, opened 2023-04-24
fix(plugin-text): disable tab list shortcut

hejtful:
Description of the bug from beta testing page:
>Moving through table cells by pressing ‘tab’ only works to some extend. As soon as a normal cell is reached ‘tab’ creates an enumeration. Is that a behavior we want to have?

Fix: As the lists package we're using doesn't support choosing which shortcuts to enable, we now manually skip the package's keydown handling if a list is not selected and the shortcut is tab or shift+tab.

### DIFF
--- a/src/serlo-editor-repo/plugin-text/hooks/use-formatting-options.tsx
+++ b/src/serlo-editor-repo/plugin-text/hooks/use-formatting-options.tsx
@@ -193,10 +193,13 @@ export const useFormattingOptions = (
       const isListsOptionEnabled = formattingOptions.includes(
         TextEditorFormattingOption.lists
       )
-      const isTab = isHotkey('tab', event)
-      const isShiftTab = isHotkey('shift+tab', event)
+      if (!isListsOptionEnabled) return
 
-      if (isListsOptionEnabled || isTab || isShiftTab) return
+      const isListActive =
+        isOrderedListActive(editor) || isUnorderedListActive(editor)
+      const isTabShortcut =
+        isHotkey('tab', event) || isHotkey('shift+tab', event)
+      if (!isListActive && isTabShortcut) return
 
       return slateListsOnKeyDown(editor, event)
     },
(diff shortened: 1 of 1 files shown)


src/serlo-editor-repo/plugin-text/hooks/use-formatting-options.tsx
```diff
@@ -190,9 +190,15 @@ export const useFormattingOptions = (
...
-      if (formattingOptions.includes(TextEditorFormattingOption.lists)) {
-        return slateListsOnKeyDown(editor, event)
-      }
+      const isListsOptionEnabled = formattingOptions.includes(
+        TextEditorFormattingOption.lists
+      )
+      const isTab = isHotkey('tab', event)
+      const isShiftTab = isHotkey('shift+tab', event)
+
+      if (isListsOptionEnabled || isTab || isShiftTab) return
```
Entkenntnis: This will also disable moving to second level of list. Is this intended?

grafik

Can we disable this only within table cells?

Entkenntnis (commented):
tabbing through table is working, thanks!

There is a side-effect that may be annoying, see comment.

src/serlo-editor-repo/plugin-text/hooks/use-formatting-options.tsx
hejtful: >Can we disable this only within table cells?

I'm already working on another bug, would you like to try this?

hejtful: Actually, I have a better idea. The increment/decrement list level only worked in the old version if a list is already selected. I will just make it the same now, that will be a simple fix which will align us with the state of the previous version.

hejtful: Updated

Entkenntnis (approved):
thanks, it now works

PR #2341, opened 2023-04-24
fix(plugin-text): center align image placeholder

hejtful:
Original bug report:
>Creating a new picture plugin shows ‘optional caption’ on the left side. Centering would be nice.


Entkenntnis (approved):
thank you, works

PR #2342, opened 2023-04-24
fix: auth provider refresh

### LINKED ISSUE #2338: In staging changing tabs leads in add-revision to reset changed content
It is probably caused after #2324
We had this problem after releasing the new user management: user edits, change tab, when they go back to serlo tab, the changes are gone

elbotho:
- makes sure we only update auth provider when auth actually changes 
- adds prominent warning for edge case (user gets logged out in the background while on `add-revision` route)

(closes #2338)

### DIFF
--- a/src/auth/auth-provider.tsx
+++ b/src/auth/auth-provider.tsx
@@ -42,10 +42,11 @@ export function AuthProvider({
     // careful: when changed this updates most of the components and can reset state in e.g. the editor!
     // use functional update to get the current value of the payload
     // returning same value will skip set state
-    setAuthenticationPayload((authenticationPayload) => {
-      const isChanged = authenticationPayload?.id !== newPayload?.id
-      return isChanged ? newPayload : authenticationPayload
-    })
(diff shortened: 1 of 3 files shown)


Entkenntnis (approved):
logging out in another tabs successfully triggers the notification for me in firefox.

not sure how it relates to #2338, but seems independant

elbotho:
> not sure how it relates to https://github.com/serlo/frontend/issues/2338, but seems independant

first commit in PR…
> …makes sure we only update auth provider when auth actually changes

that is the main problem in https://github.com/serlo/frontend/issues/2338

elbotho:
@Entkenntnis if you are still working on the e2e test here I will not merge it for now.

Entkenntnis:
@elbotho You can go ahead and merge it into staging, will make it easier to test.

I have merged the e2e test, so we can see if it works in staging.

Only open task: Maybe @hugotiburtino can check if this merge is fixing his issue #2338 or if it still persist

PR #2351, opened 2023-04-25
fix: Emphasize last add button in editor document

LarsTheGlidingSquirrel:
This PR fixes and improves a broken feature: The last add button within a section of the editor should be always visible and showing a label next to it. 

Most of the time only one add button will be emphasized this way on the whole page. However, on certain content types more add buttons are emphasized (Course, Exercise, ...). For example in a course we emphasize the last button after each course page. See: http://localhost:3000/entity/repository/add-revision/51979 

Implementation uses useScopedSelector(...) to get all ancestor plugin names from the redux store and then decides depending on the context if the last add button should be emphasized. For example, it should not be emphasized within a box or nested in another RowEditor.


elbotho:
@LarsTheGlidingSquirrel I'm not sure if it works for me, but maybe I also misunderstand what the PR is supposed to do.

Does it work for you in the preview (e.g. [here](https://frontend-git-fix-emphasize-last-add-button-broken-serlo.vercel.app/entity/create/Article/60730) or [here](https://frontend-git-fix-emphasize-last-add-button-broken-serlo.vercel.app/entity/repository/add-revision/1555))?

LarsTheGlidingSquirrel:
> Does it work for you in the preview (e.g. [here](https://frontend-git-fix-emphasize-last-add-button-broken-serlo.vercel.app/entity/create/Article/60730) or [here](https://frontend-git-fix-emphasize-last-add-button-broken-serlo.vercel.app/entity/repository/add-revision/1555))?

@elbotho Oh, no it does not. I thought I checked that, but guess I missed it. Thanks for checking again.

elbotho:
I'll be patient then :)

src/serlo-editor-repo/plugin-rows/editor/index.tsx
```diff
@@ -82,7 +88,19 @@ export function RowsEditor(props: RowsProps) {
 
   if (!props.editable) return <RowsRenderer {...props} />
 
-  const isEditorForRootOfDocument = props.id === 'root'
+  // Prevent add button being visually emphasized when this RowsEditor is contained within certain plugin types.
+  const visuallyEmphasizeLastAddButton =
```
elbotho, line 92: This means any time we introduce a plugin that contains a row plugin we need to think about changing this. 
I see you already tried, but just to be sure: Is there a feasible way identify the root without checking all plugin types?

elbotho (approved):
Thank you!
Added one question but no blocker.

hejtful (commented):
Just an idea:

There is already one `Separator` hardcoded before any rows are rendered.

Would it make things simpler to hardcode one after all the rows as well, and make that one emphasized? And then inside the last row, simply not render its own separator?

Please feel free to ignore if it doesn't make sense here.

src/serlo-editor-repo/plugin-rows/editor/index.tsx
LarsTheGlidingSquirrel: > This means any time we introduce a plugin that contains a row plugin we need to think about changing this.
I see you already tried, but just to be sure: Is there a feasible way identify the root without checking all plugin types?

I tried a few things but no simple solution like `props.id === 'root'` worked in all content types. For example, in type exercise we have two RowsEditors nested like this: 
- "type-text-exercise", "exercise", "rows"
- "type-text-exercise", "type-text-solution", "solution", "rows"

The fact that they are at nested at different levels and that other content types each have their own nesting structure made it difficult for me to find a solution that worked everywhere and does not rely on checking the plugin types on the path to the root.

LarsTheGlidingSquirrel:
> There is already one `Separator` hardcoded before any rows are rendered.
> 
> Would it make things simpler to hardcode one after all the rows as well, and make that one emphasized? And then inside the last row, simply not render its own separator?

Yeah, I agree. To clean it up even more I think we could change it so that `RowEditor` does not contain a `Separator` itself and that they are all created by `RowsEditor`: One `Separator` at the end and for each row one `Separator` and one `RowEditor`. We could have a look into a change like this as a followup if you like.

src/serlo-editor-repo/plugin-rows/editor/index.tsx
elbotho: thanks for checking!

PR #2358, opened 2023-04-26
fix(plugin-text): list at start of editor bug

### LINKED ISSUE #2350: text-plugin: some lists break editor
error message:
```
Error: Cannot get the leaf node at path [0,0] because it refers to a non-leaf node: {"type":"list-item","children":[{"type":"list-item-child","children":[{"text":""},{"type":"a","href":"http://de.serlo.org/115817","children":[{"text":"Belege der Paläontologie (Fossilien)"}]},{"text":""}]}]}
```

It seems to affect some lists.

sexamples:
- https://de.serlo.org/entity/repository/add-revision/1795
- https://de.serlo.org/entity/repository/add-revision/1629
- https://de.serlo.org/entity/repository/add-revision/1685
- https://de.serlo.org/entity/repository/add-revision/1999
- https://de.serlo.org/entity/repository/add-revision/80524

example where lists don't have this problem:
https://de.serlo.org/entity/repository/add-revision/1541

hejtful:
resolves #2350

The issue was in the workaround for focusing the editor when a new editor is added on enter key press.

The fix is to make that workaround more specific, and only run it when the editor is empty, because that's when it's needed.

### DIFF
--- a/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
@@ -101,10 +101,17 @@ export function TextEditor(props: TextEditorProps) {
     // ReactEditor.focus(editor) does not work without being wrapped in setTimeout
     // See: https://stackoverflow.com/a/61353519
     setTimeout(() => {
-      const pointAtStart = { offset: 0, path: [0, 0] }
-      Transforms.select(editor, selection || pointAtStart)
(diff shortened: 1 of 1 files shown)


elbotho (approved):
Thanks for the quick fix. We should probably deploy this soon since it affects a lot of content.

Entkenntnis (approved):
thanks! There is a deployment today evening, so merge into staging and it will be live tomorrow.

PR #2361, opened 2023-04-26
remove obsolete slate hacks

elbotho:
Did a quick test and did not find problems.
All hail the new slate.


Entkenntnis (approved):
one hack less

elbotho:
even two I think :D

PR #2368, opened 2023-04-28
Add translation to turn sign up validation message friendly

### LINKED ISSUE #2367: No friendly message anymore when try to registering with invalid username
image
Note: Umlaute are not allowed from legacy until now.

Before that we had a better error message like "Your username may only contain letters, digits, underscores (_) and hyphens (-). Please use a different one." in the English instance. But we are getting this error message in English also in de.serlo.org. See legacy for [English](https://github.com/serlo/serlo.org-legacy/blob/main/packages/public/server/src/lang/en_US.UTF-8/LC_MESSAGES/default.po#L740) and [German](https://github.com/serlo/serlo.org-legacy/blob/main/packages/public/server/src/lang/de_DE.UTF-8/LC_MESSAGES/default.po#L741) messages.

This is the [place](https://github.com/serlo/frontend/blob/248244886046594a844e8e1ee122d1dda3fe542e/src/components/auth/message.tsx#L29) where it should be working.

Chorme 112, but it should be unrelated to browser.

CodingDive:
Fixes #2367

### DIFF
--- a/src/data/de/index.ts
+++ b/src/data/de/index.ts
@@ -388,6 +388,8 @@ export const instanceData = {
         code4000001: '%reason%',
         code4000002: "%field% bitte noch angeben.",
         code4000005: '%reason%',
+        // Should map to usernameInvalid
+        code4000004: '%reason%',
         code4000006: "Der Benutzername, die E-Mail-Adresse oder das Passwort stimmen so nicht. Bitte überprüfe deine Eingabe.",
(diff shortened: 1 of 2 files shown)


elbotho:
please don't edit `de/index.ts` it really messes with the i18n automation.

Entkenntnis (approved):
translations can be added on staging - in this case this isn't even necessary I thing.

PR #2370, opened 2023-05-02
fix(plugin-math): prevent default on cut

hejtful:
https://trello.com/c/t8wqq0DB/122-bug-cutting-a-part-of-a-formula-not-possible-text-plugin


Entkenntnis (approved):
thank you, this fixes the described bug.

PR #2371, opened 2023-05-02
fix(header-footer): move overflow hidden class to outside wrapper

hejtful:
https://trello.com/c/Qqrd9CUj/121-bug-latex-box-not-fully-visible-half-screen


Entkenntnis (approved):
thanks for investigating, sticky menu works as before, so this is good to go

PR #2379, opened 2023-05-03
fix(plugin-math): visual formula issues

hejtful:
- fix copy and cut functionality in visual formula (it was already fixed for LaTeX formula in previous PRs)

https://trello.com/c/t8wqq0DB/122-bug-cutting-a-part-of-a-formula-not-possible-text-plugin

### DIFF
--- a/src/serlo-editor-repo/math/visual-editor.tsx
+++ b/src/serlo-editor-repo/math/visual-editor.tsx
@@ -95,6 +95,12 @@ export function VisualEditor(props: VisualEditorProps) {
       onChange={(ref) => {
         props.onChange(ref.latex())
       }}
+      onCopy={(event: React.ClipboardEvent) => {
+        event.stopPropagation()
+      }}
+      onCut={(event: React.ClipboardEvent) => {
+        event.stopPropagation()
+      }}
       // @ts-expect-error https://github.com/serlo/serlo-editor-issues-and-documentation/issues/67
       config={mathQuillConfig}
       mathquillDidMount={onMount}


src/serlo-editor-repo/math/visual-editor.tsx
```diff
@@ -95,6 +95,12 @@ export function VisualEditor(props: VisualEditorProps) {
       onChange={(ref) => {
         props.onChange(ref.latex())
       }}
+      onCopy={(event: React.ClipboardEvent) => {
+        event.stopPropagation()
+      }}
+      onCut={(event: React.ClipboardEvent) => {
+        event.stopPropagation()
+      }}
```
LarsTheGlidingSquirrel, line 103: Do you see a need in the near future to let this event bubble up further to the parent event handlers? Maybe when we select and copy/cut some text and inline math elements in one go.

LarsTheGlidingSquirrel (approved):
Works nice! Just added a thought to maybe think about as a followup.

src/serlo-editor-repo/math/visual-editor.tsx
hejtful: I don't really see that need.

In the case you mentioned, the math editor is not focused, so its copy and cut handlers won't be triggered anyway, the whole math element will be copied instead.

LarsTheGlidingSquirrel: I guess the math editor will be focused when the selection ends within a math editor. But I think this use case is not happening very often.

hejtful: The selection currently works so that you cannot select a part of the math editor, when already selecting outside the math editor. Only the whole math element can be selected in that case.

PR #2380, opened 2023-05-04
chore(editor): remove deprecated blockquote config prop from text plugin

### LINKED ISSUE #104: Remove blockquote config property from editor codebase
It is a deprecated property and the blockquote plugin will be removed with #94  anyway

hejtful:
resolves https://github.com/serlo/backlog/issues/104


LarsTheGlidingSquirrel (approved):
Looks good! Seems like it is not used anyway currently. 
Side note: Plugin blockquote still appears in the database, so we should not remove `plugin-blockquote` until after the database migrations.

PR #2383, opened 2023-05-04
chore(plugin-text): resolve lists util todos

hejtful:
- Create an enum with list element types, as in slate-lists package examples
- Just remove the todo to remove castings. Slate's recommended typing approach is to have a union type of all the custom elements, which simply means that sometimes we need to cast in order to specify which type from the union we are returning. If someone has an idea how to do this without casting, please share!


Entkenntnis:
frontend uses this technique alot: https://basarat.gitbook.io/typescript/type-system/discriminated-unions

as a last resort something like this can also avoid casting: https://basarat.gitbook.io/typescript/type-system/typeguard#user-defined-type-guards

Ignore if this is not appliable to your problem.

hejtful:
Unfortunately, it doesn't really help in this case, as we're not trying to use a received union type, but trying to return a new object as one of the union types.

Entkenntnis:
not a complete solution, what I was able to get rid of the type casts by adding a new check:

grafik

It could happen that props e.g. have additional attributes or contains wrong children types. Same goes for other function.

slate-lists documentation is a bit unclear about what to do in these cases

Entkenntnis:
a slightly different schema here: https://github.com/prezly/slate/pull/148

edit: still has the problem that children could be of different types ...

hejtful:
Thanks a lot for the feedback @Entkenntnis !

There are schemas all around, for example [in this one](https://codesandbox.io/s/prezly-slate-lists-user-guide-4-add-onkeydown-handler-5wpqxv?file=/src/MyEditor.tsx:344-451) the Slate declaration looks like this:
```
declare module 'slate' {
    interface CustomTypes {
        Element: { type: Type } & BaseElement;
    }
}

enum Type {
    PARAGRAPH = 'paragraph',
    ORDERED_LIST = 'ordered-list',
    UNORDERED_LIST = 'unordered-list',
    LIST_ITEM = 'list-item',
    LIST_ITEM_TEXT = 'list-item-text',
}
```
but that doesn't really work for us, because we have different custom elements that have additional props which `BaseElement` doesn't have...

In general, looking at all these solutions, I find that casting the objects' types, at the spot where we create them, is the simplest solution. Still, it's not a strong opinion, so if you prefer some other solution, just tell me which one, and I'll try to implement it.

Entkenntnis:
I got something like this to work:

grafik

put I'm struggling to make this work for others ...

It seems that 

```
return { children: [{ text: '' }], ...props, type: nodeType } as List
```

is wrong, because `List` exepects `ListItem` as children, not `CustomText`?

---

If this is possible for all cases this would be a preferable solution, otherwise I'm fine with sticking with casts

hejtful:
@Entkenntnis Thanks a lot for your persistence! The `Partial` props approach worked fine for all of them, the casts are gone. And you're right, `CustomText` was being returned (and this is just a guess, as this is the code I found when I joined) because Slate can interpret it and still create a valid list. But now the correct types are being returned by all functions.

Entkenntnis (approved):
thank you fixing the todo

PR #2384, opened 2023-05-04
fix(plugin-text): clear timeouts in useEffect hooks

hejtful:
Found some timeout that are not cleared on unmount. Although it probably isn't the cause of the performance bug, it's still good to clear those timeouts.


src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -98,8 +98,10 @@ export function TextEditor(props: TextEditorProps) {
 
   useEffect(() => {
     if (focused === false) return
+
     // ReactEditor.focus(editor) does not work without being wrapped in setTimeout
     // See: https://stackoverflow.com/a/61353519
+    let timeout: ReturnType<typeof setTimeout>
     setTimeout(() => {
```
Entkenntnis, line 105: `timeout =` missing? :)

Entkenntnis: `timeout =` missing? :)

Entkenntnis (changes requested):

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: 😅 Thanks!!

Entkenntnis (approved):
thx

PR #2387, opened 2023-05-08
fix(editor): plugin configuration modal issue

hejtful:
Frontend and editor modals use different positioning techniques. For frontend, it's position absolute+transition. For editor, it's flexbox justify content. The frontend CSS was spilling into the editor CSS, so this is just a quick fix to make sure editor modal content isn't transform-pushed to the left.

With bug:
Screenshot 2023-05-08 at 10 45 01

With fix:
Screenshot 2023-05-08 at 10 45 06


elbotho (approved):
thank you 🎉

PR #2394, opened 2023-05-08
fix/selection menu improvements

### LINKED ISSUE #2388: editor: selection menu improvements
### small bugs/problems

- [x] add text plugin as first plugin (activating it just removes the `/`)
- [x] Hover (mouse) and focused (arrow keys) entries both are visually prominent and can be at the same time. Solution: only one should be visible at the same time (hide other focus when hovering with mouse or manage in state)

### wish (optional)
- remove `/` when closing the popover or deselecting the text-plugin

hejtful:
resolves https://github.com/serlo/frontend/issues/2388

### DIFF
--- a/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/text-editor.tsx
@@ -50,7 +50,6 @@ import {
   replace,
 } from '@/serlo-editor-repo/store'
 
-/** @public */
 export type TextEditorProps = EditorPluginProps<
   TextEditorState,
   TextEditorConfig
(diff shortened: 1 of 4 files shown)


src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -98,17 +96,23 @@ export function TextEditor(props: TextEditorProps) {
...
   useEffect(() => {
-    // If the editor is not focused, do nothing
-    if (focused === false) return
+    // Get the current text value of the editor
+    const text = Node.string(editor)
+
+    // If the editor is not focused, remove the suggestions search
+    // and exit the useEffect hook
+    if (focused === false) {
+      if (text.startsWith('/')) {
```
elbotho, line 105: Is it possible to check if the popover is still open at this point?
Currently this would always remove a line like `// this is how a comment looks like`.
Def. an edge case but maybe there are more cases like this (not sure).

Fallback: Check if whitespaces are in text and then return.

elbotho (approved):
One question/request. Rest looks great 👍

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: > Is it possible to check if the popover is still open at this point?

The popover will be closed at this point, because it's only open when `focused === true`.

This logic will change, however. I'll open an issue for that now.

Issue #2369, opened 2023-05-02
Investigate possible memory leak


hejtful:
https://trello.com/c/Z36akMeR/125-bug-performance-problems-editor

Entkenntnis:
I quick profile in chromium shows no obvious problems: there are many allocations, but GC is able to keep usage low.

![grafik](https://user-images.githubusercontent.com/13507950/235681663-58e077b9-aba5-4f89-b00b-6c31d12115b9.png)

it's notable that the editor is quite allocation-heavy though: usage grows to ~80 - 100MB regularly and needs gc-collects to shrink again

hejtful:
When inserting an image plugin, the following error appears. Could be a memory leak, worth investigating and fixing

<img width="1038" alt="Screenshot 2023-05-08 at 13 41 37" src="https://user-images.githubusercontent.com/13198821/236814909-b97802e2-d25f-4e7b-9ee5-f9af0017d342.png">

Edit: Fixed by https://github.com/serlo/frontend/pull/2399

Entkenntnis:
closed in trello

PR #2399, opened 2023-05-08
fix(plugin-image): redux state change during render

hejtful:
Problem: Redux state was being updated during render
Screenshot 2023-05-08 at 13 41 37

Relevant `react-redux` issue: https://github.com/reduxjs/react-redux/issues/1640

This PR could help with https://github.com/serlo/frontend/issues/2369

Note: You might notice that there is an issue with the URL field of the image plugin. This was an already existing bug, not introduced by this fix. Will open a new issue for that.

**Update**: The URL field issue was an existing bug, but it was escalated by this PR, as before only the image plugin crashed, now the whole app crashes. Need to investigate and fix that before merging this PR.
**Another update**: Weirdly, after I merged this branch with staging, the app breaking stopped.


Entkenntnis (approved):
thx for the fix

PR #2408, opened 2023-05-09
refactor(editor): explicit React imports, class components

elbotho:
sorry for the PR spam, local builds and checks work fine and I'm trying to find the problem.

---- 
- used explicit React imports to merge code styles (not really necessary but…)
- refactored some old class components into functional components (the remaining class comps are special cases or quite big and need some testing)


elbotho:
@Entkenntnis I have to conclude that this PR is cursed 👻, maybe you have an idea. 
Locally all checks work fine, but the gh actions throw errors that (I think) include old code.

Entkenntnis (approved):
thanks for cleaning up the code style

hejtful (approved):
Nice, thanks! In case you do something similar in the future, please feel free to remove the "internal", "public" and similar comments, these are leftovers from when edtr-io was supposed to be widely used

elbotho:
> please feel free to remove the "internal", "public" and similar comments, these are leftovers from when edtr-io was supposed to be widely used

I will, thanks!

PR #2409, opened 2023-05-09
Refactor redux selectors

LarsTheGlidingSquirrel:
This pull request tries to make some of our redux selector code clearer. 

Background: While trying to understand the existing redux selector code and writing a new selector we noticed: 
- It was difficult to figure out when and how reselect was used for memoization. 
- It was not clear when a component would re-render when using `useScopedSelector(...)`
- Functions returning functions, returning functions, and so on made it hard to understand what would be executed in the end. 
- `createDeepEqualSelector` is using reselect in a way that is not recommended by the docs: Deriving values in the input selector. See https://redux.js.org/usage/deriving-data-selectors 

Instead of only documenting the existing code, we thought about doing some refactoring. 

Changes: 
- `useScopedSelector(...)` now communicates more clearly that it is basically doing the same as `useSelector(...)` from react-redux, only limited to the current scope. 
    - Parameter `enforceScope` was removed because it is not used anywhere.
    - Optional parameter `equalityFn` was added to make it look like `useSelector(...)` 
- `createDeepEqualSelector` is replaced in two places by directly using `Reselect.createSelector(...)`.
    - Makes it more clear in my opinion that reselect is used
    - Communicates more clearly how memoization is used to avoid unnecessary component re-renders
- Some renaming and comments
- Bumped minor version of reselect to use new feature

More changes could be done following this approach, but first I would like to get some feedback. 

Side note: In this pull request `(scopedSelector) => scopedSelector` is used as an input selector in `Reselect.createSelector`. This is also not recommended in the docs. But we can probably change it to `(scopedSelector) => scopedSelector.documents` or something similar, but that seems like it requires more changes, so I left it out for now.


src/serlo-editor-repo/core/store.tsx
```diff
@@ -77,19 +77,21 @@ function scopeDispatch(dispatch: (action: Action) => void, scope: string) {
...
  * @public
  */
 export function useScopedSelector<T>(
   scopedSelector: (state: ScopedState) => T,
-  enforcedScope?: string
+  equalityFn?: (previous: T, next: T) => boolean
 ) {
-  const scope = useScope(enforcedScope)
-  return useSelector((state) => scopedSelector(getScope(state, scope)))
+  const scope = useScope()
```
hejtful, line 87: It seems like the `enforcedScope` argument of `useScope` is never used, anywhere in the code. Maybe consider removing that as well.

If you don't feel like this change belongs to this PR, feel free to ignore.

src/serlo-editor-repo/store/documents/reducer.ts
```diff
@@ -112,12 +112,20 @@ export const documentsReducer: SubReducer<Record<string, DocumentState>> =
...
-/** @public */
-export const getDocument: Selector<DocumentState | null, [string | null]> =
-  createSelector((state, id) => {
+/**
+ * Creates a selector that gets the document with the given id.
+ *
+ * @param id The id of the document that should be selected
+ * @returns A selector `(ScopedState) => DocumentState`. The returned DocumentState object reference will only change when this document was modified by a redux action.
+ * @example const document = useScopedSelector(getDocument(id)) // Gets the document with the given id and trigger component re-render when this document was modified by a redux action.
+ * @public
```
hejtful: No need for these `@public` comments any more, I believe. They are just remnants of when `edtr-io` was designed as a public package.

src/serlo-editor-repo/store/focus/reducer.ts
```diff
@@ -1,4 +1,5 @@
 import * as R from 'ramda'
+import * as Reselect from 'reselect'
```
hejtful, line 2: Rather use explicit imports like `import { createSelector } from 'reselect'`, to enable better tree shaking.

hejtful (commented):
Just some suggestions. Looking forward to see the final state of the PR!

src/serlo-editor-repo/store/focus/reducer.ts
LarsTheGlidingSquirrel: This was just a quick workaround because there is already a `createSelector` function defined that would result in a name clash. I agree with your suggestion.

Entkenntnis: `import {creatSelector as reselectCreateSelector} from 'reselct'` :)

LarsTheGlidingSquirrel: I would rather remove or rename our custom `createSelector` function. But I'll wait with that for now so that there aren't too may changes.

PR #2413, opened 2023-05-10
handle empty value in plugin-text deserialization

### LINKED ISSUE #2411: editor: crash on certain articles
example: https://de.serlo.org/nachhaltigkeit/270596/veganerin-werden-wie-fange-ich-an

there are errors in the console

Entkenntnis:
The debugging shows that `value` can be  `[]`. Now the deserializer handles this case and it fixes the bug.

example: https://frontend-git-2411-editor-crash-on-certain-articles-serlo.vercel.app//entity/repository/add-revision/270596/270598

### DIFF
--- a/src/serlo-editor-repo/plugin-text/index.tsx
+++ b/src/serlo-editor-repo/plugin-text/index.tsx
@@ -35,7 +35,6 @@ const createTextPlugin = (
       return value
     },
     deserialize(value) {
-      // sometimes, the value is empty, e.g. https://github.com/serlo/frontend/issues/2411
       if (value.length === 0) {
         return emptyDocumentFactory()
       }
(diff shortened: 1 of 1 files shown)


src/serlo-editor-repo/plugin-text/index.tsx
```diff
@@ -35,6 +35,11 @@ const createTextPlugin = (
       return value
     },
     deserialize(value) {
+      // sometimes, the value is empty, e.g. https://github.com/serlo/frontend/issues/2411
```
hejtful: I'm not sure how helpful this comment is, as the production link from the issue will be working when this is deployed.

hejtful (approved):
Minor suggestion, but great work in finding the cause so quickly

src/serlo-editor-repo/plugin-text/index.tsx
Entkenntnis: true, will probably just remove it

PR #2427, opened 2023-05-15
Fix math element disappearing & make slate see math elements as inline/block

### LINKED ISSUE #2410: editor: math element sometimes disappears
When the math element is the last element of the text plugin and is in block mode, it tends to disappear. This only happened for me when the math element is already in the loaded state, not for newly added elements.

Authors reported it as highest priority bug.

https://trello.com/c/t8wqq0DB/122-bug-formula-disappears-during-editing-text-plugin

LarsTheGlidingSquirrel:
This pull requests provides a fix for #2410 / https://trello.com/c/t8wqq0DB/122-bug-formula-disappears-during-editing-text-plugin

This seems to fix the main issue and could be deployed as a quick fix in production. However, introduces some followup issues, see below. 

### Before

During certain actions (for example cut inside latex editor) the slate normalization was triggered. This deleted the math element because of a default rule within slate normalization: A block element and an inline element cannot be siblings. See: https://docs.slatejs.org/concepts/11-normalizing#built-in-constraints
Even though the math element was marked as a block element by the property `inline=false`, slate still recognized it as an inline element because calling `editor.isInline(someMathElement)` always returned true. 

### After

Math elements can now be of type 'inline' or 'block' and are recognized by slate as such. 

Changing a math element from inline to block or the other way around performs the necessary node transformations. See large comment in code with an example of such a transformation. The goal of these transformations is to ensure that the node structure follows the slate rules. 

Small unrelated change about ignoring failed attempts to focus a text editor. 

### Followups

This seems to fix the main issue but there are some issues I found while testing. In my opinion these can be done in followups after releasing a quick fix.

- Making a math element inline inside a list is broken
- HoverOverlayOld is re-rendered too often (but no infinite re-render) when clicking inside the latex textarea. State `nativeSelection` triggers this.
- Pressing up arrow in text below math block to navigate to the block above is broken. Down arrow from the element above works still.
- Existing math elements have a weird invisible line at the bottom that you can navigate to with the keyboard. Pressing enter there makes the math element disappear.


hejtful:
>- HoverOverlayOld is re-rendered too often (but no infinite re-render) when clicking inside the latex textarea. State nativeSelection triggers this.

`nativeSelection` here makes sure that the `HoverOverlayOld` and its content are positioned properly at all times. Just for context. If you have some improvements in mind, that's great!

hejtful (approved):
Great work! It's really easy to understand the code and the comments.

I would just propose a couple of minor refactors:

1. Move `transformNodeToInline` and `transformToBlockNode` to `/utils/math.ts`, and maybe rename them to `transformMathNodeToInline` and `transformMathNodeToBlock`. **OR**
2. Make the above-mentioned functions non-math-specific, and move them to `/utils/documents.ts` without renaming them.

These are just suggestions to make this component a bit cleaner. If you see a better way to do that, that's fine with me.

LarsTheGlidingSquirrel:
> I would just propose a couple of minor refactors:
    1. Move transformNodeToInline and transformToBlockNode to /utils/math.ts, and maybe rename them to transformMathNodeToInline and transformMathNodeToBlock. OR
    2. Make the above-mentioned functions non-math-specific, and move them to /utils/documents.ts without renaming them.

Good idea! I was thinking to look for other places in the code where we do similar node transformations and merge them. We can do that as a followup.

src/serlo-editor-repo/plugin-text/plugins/math.ts
```diff
@@ -4,7 +4,11 @@ export const withMath = (editor: SlateEditor) => {
   const { isInline, isVoid } = editor
 
   editor.isInline = (element) => {
-    return element.type === 'math' ? true : isInline(element)
+    if (element.type !== 'math') {
+      return isInline(element)
+    }
+
+    return element.inline
```
elbotho: just an optinion: I prefer the inline syntax over the is-not logic here.

Entkenntnis (approved):

src/serlo-editor-repo/plugin-text/plugins/math.ts
LarsTheGlidingSquirrel: Good suggestion! Changed it.

PR #2433, opened 2023-05-21
refactor(editor): clean up and upgrade redux store

hejtful:
## Motivation - current state of the Redux store:
- Old Redux with a lot of boilerplate
- Usage of deprecated `createStore`
- Unnecessary "scoped" store, when "main" scope is only ever used
- Configurable options that were useful for etdr-io, but unused by the new editor
- Actions and sagas that are not used anywhere

**Note:** I might very well be wrong about some of the above, please let me know in that case.

## Scope:
- Use Redux Toolkit ([Why?](https://redux.js.org/introduction/why-rtk-is-redux-today))
- Clean up unused code across the store and editor in general

## Follow-up issues:
- https://github.com/serlo/backlog/issues/127
- Deal with Redux Saga, by either:
  1. Cleaning up and typing sagas properly (+ maybe testing the sagas) **OR**
  2. Switching from Redux Saga to Thunks (which come with Redux Toolkit)

## Review
I'm aware that reviewing all the changed files and lines might be too much. If you prefer to have a call where I can show the relevant changes, I'd be happy to do it. Otherwise, you might want to simply switch to the branch and try it out. Any caught bugs would be greatly appreciated.


Entkenntnis:
Really love `immer`, it makes reducers so much simpler and safer to write :heart_eyes:

elbotho:
👍
Did some testing and did not find any obvious bugs other than this:
- when I delete the title of an article and undo the title appears twice (Just checked, also present in production…)

Entkenntnis:
@hejtful In testing, I found that this PR has more lag while typing, compare

https://frontend-git-refactor-clean-up-upgrade-redux-store-serlo.vercel.app/entity/repository/add-revision/1553

with staging. Is there maybe an issue with the selectors, e.g. the cache size is exceeded? (the article is quite big ...)

hejtful:
@Entkenntnis That's a great find, thanks a lot! I managed to reproduce it, but couldn't catch anything obvious with React Profiler. I'll try to get to the bottom of it tomorrow.

hejtful:
@Entkenntnis Quite a few components optimized, it's noticeably less laggy for me, it would be amazing if you could check again!

Entkenntnis (approved):
Thank you! There is no noticable regression in lag anymore.

I think we can continue to test this on staging.

src/serlo-editor-repo/store/helpers.ts
```diff
@@ -1,139 +1,15 @@
...
-  Selector,
-  ActionCreatorWithPayload,
-  ActionCreatorWithoutPayload,
-  InternalScopedState,
-  InternalSelector,
-  InternalState,
-} from './storetypes'
+export const createDeepEqualSelector = createSelectorCreator(defaultMemoize, {
+  resultEqualityCheck: R.equals,
+  maxSize: 50,
```
LarsTheGlidingSquirrel, line 6: I think I would prefer using [selector factories](https://redux.js.org/usage/deriving-data-selectors#selector-factories) and get rid of this `maxSize` value. This number is something additional we have to think about when working on the code. Using a selector factory, we could create a selector for each `id` with a cache size of 1. Right now, I don't see any downside with my suggestion, but I might be missing something. If you want, I can make these changes but we can also leave it like this.

src/serlo-editor-repo/store/focus/selectors.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    return {
+      id: root,
+      children,
+    }
+  }
+)
+
+export const selectParent = createSelector(
+  [(state: State) => state, (_state, id: string) => id],
```
LarsTheGlidingSquirrel, line 52: It would be great if we could rewrite this selector (and others in the same file) to not rerun on every action. They do not need to rerun, when the focus changes for example. But we can work on this as a followup.

LarsTheGlidingSquirrel (approved):
Really awesome that you took the time to clean up this part of the editor code and were brave enough to also remove a lot. The clarity of the code definitely improved for me. I like how the redux selectors became much more clear.

src/serlo-editor-repo/store/helpers.ts
hejtful: I don't see any downside either, and I'd love if you'd get involved in the Redux code straight away, as I think it would give you an opportunity to get even more ideas for improvement 😄 Please feel free to add those changes to this PR, or as an improvement in a next PR.

src/serlo-editor-repo/store/focus/selectors.ts
hejtful: For sure, I was thinking about it for some time. In general, the `selectFocusTree` is used all around, not only for focusing feature. And it has to select state from documents and plugins, which makes me wonder if we should re-think the whole state structure.

In any case, if you think of a way to rewrite this selector without changing the state structure, please feel free to add the change, whenever. Otherwise, we can think about the problem a bit deeper in the future.

PR #2434, opened 2023-05-22
Prevent block-type math elements in lists & fix bug when merging math element with siblings

LarsTheGlidingSquirrel:
### Before

1. Bug: When the user toggles a math element to be inline and there is another block-type math element before or afterwards, merging with this nodes siblings will fail. 
2. Users can create a block-level math element in lists, effectively creating line breaks in lists. 

### After

1. Merging is skipped if the element before or after is not a "paragraph"
2. Checkbox to toggle inline inside math editor is hidden when the math element is inside a list. 

### Refactorings

- Introduced function `existsInAncestors` and used it for checking if A) an element is within a list, B) the current selection is within a list
- Renamed and introduced new functions to make it more clear what the function does. Example: `isOrderedListActive` became `isSelectionWithinOrderedList`

### DIFF
--- a/src/serlo-editor-repo/plugin-text/components/math-element.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/math-element.tsx
@@ -14,6 +14,7 @@ import type {
   Paragraph,
   TextEditorPluginConfig,
 } from '../types'
+import { isElementWithinList } from '../utils/list'
 import { MathFormula } from './math-formula'
 
 export interface MathElementProps {
@@ -38,17 +39,7 @@ export function MathElement({
   const preferences = useContext(PreferenceContext)
 
   const isInsideListElement = useMemo(() => {
-    const path = ReactEditor.findPath(editor, element)
-    const ancestorNodes = [...Node.ancestors(editor, path)]
-    return ancestorNodes.some((ancestor) => {
-      const ancestorPath = ancestor[1]
-      const ancestorNode = Node.get(editor, ancestorPath)
-      return (
-        'type' in ancestorNode &&
-        (ancestorNode.type === 'ordered-list' ||
-          ancestorNode.type === 'unordered-list')
-      )
-    })
+    return isElementWithinList(element, editor)
   }, [editor, element])
 
   const shouldShowMathEditor =
--- a/src/serlo-editor-repo/plugin-text/components/math-element.tsx
+++ b/src/serlo-editor-repo/plugin-text/components/math-element.tsx
@@ -63,7 +63,7 @@ export function MathElement({
   }
 
   /**
-   * Applys slate node transformations when MathElement is changed from inline to block type and backwards.
+   * Applies slate node transformations when MathElement is changed from inline to block type and backwards.
    */
   function handleInlineChange(newInlineValue: boolean) {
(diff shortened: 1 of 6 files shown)


Entkenntnis (approved):
thank you - code changes look good

as again, quite involved to test manually and catch all edge cases :) I hope we can add more e2e tests in the future

src/serlo-editor-repo/plugin-text/components/math-element.tsx
```diff
@@ -37,6 +37,20 @@ export function MathElement({
   const selected = useSelected()
   const preferences = useContext(PreferenceContext)
 
+  const isInsideListElement = useMemo(() => {
```
hejtful, line 40: Maybe you can just use `isOrderedListActive(editor) || isUnorderedListActive(editor)` here instead? You can get those helpers from `'../utils/list'`.

src/serlo-editor-repo/plugin-text/components/math-element.tsx
```diff
@@ -37,6 +37,20 @@ export function MathElement({
...
+      const ancestorPath = ancestor[1]
+      const ancestorNode = Node.get(editor, ancestorPath)
+      return (
+        'type' in ancestorNode &&
+        (ancestorNode.type === 'ordered-list' ||
+          ancestorNode.type === 'unordered-list')
+      )
+    })
+  }, [editor, element])
+
```
hejtful:
```typescript
import { isOrderedListActive, isUnorderedListActive } from '../utils/list'
```

```suggestion
  const isInsideListElement = useMemo(() => {
    return isOrderedListActive(editor) || isUnorderedListActive(editor)
  }, [editor])
```

Just tried it, seems to work. What do you think?

LarsTheGlidingSquirrel: Nice suggestion! I see the potential to reuse or combine the code. However, looking into `isOrderedListActive` I find myself putting more work into understanding what it does than the current solution. For example, I have to additionally think about selection stuff. I'll try to figure something out that is more clear, but inspired by your suggestion.

src/serlo-editor-repo/plugin-text/utils/list.ts
```diff
@@ -1,32 +1,58 @@
...
+  editor: SlateEditor,
+  listType?: 'unordered-list' | 'ordered-list'
+) {
+  if (!editor.selection) return false
+
+  return existsInAncestors(
+    (element) => {
+      return listType
+        ? element.type === listType
+        : element.type === 'ordered-list' || element.type === 'unordered-list'
```
hejtful: Maybe consider using the enums from [this file](https://github.com/serlo/frontend/blob/staging/src/serlo-editor-repo/plugin-text/types/text-editor.ts#LL31C16-L31C16) instead of string literals.

(I know string literals were used before, but that's old code 😄 )

hejtful (approved):
Looks great! Thanks for the fix and for the refactor, I also find it more organized and understandable now

src/serlo-editor-repo/plugin-text/utils/list.ts
LarsTheGlidingSquirrel: I agree, that will be less vulnerable to sneaky typos. :)

PR #2439, opened 2023-05-23
editor(text-plugin): link tool with search

elbotho:
often requested feature to make linking serlo content easier.
first iteration.
background: https://miro.com/app/board/uXjVP0cDuaI=/?moveToWidget=3458764548239622696&cot=14
preview: https://frontend-git-editor-link-rework-serlo.vercel.app/entity/create/Article/60730

I added a flag to the config so the search function can be disabled when the editor is used outside of serlo or in other instances of serlo.org where the quickbar data is not available. 
The component should work fine, but without suggestions obviously.

image

image

### todos:

- [x] Quickbar data only available for de.serlo.org! how should we handle other instances?
- [x] (maybe i18n: see above)
- [x] ⚠️ currently it does not work with keyboard navigation! (because the popover is at the very end of the tab order – portal maybe?)
- [x] long titles can overflow the popover atm (`2137` is a good example)

### possible follow ups / improvements
- filter by subject or show subject in suggestions
- make link validation and distinction between internal and external links more solid
- reintroduce the arrow triangle / improve contrast between content and popover
- maybe suggest content from same taxonomy first or recently used links
- is quickbar fetcher now hosted on serlo servers?
- Icons: I'd like to add type info to quickbar data (for type icons)
- edge case: maybe find a way to save last query (when popover is closed without a selection and the query makes sense as href)
…


src/components/navigation/google-search-wrapper.tsx
```diff
@@ -31,7 +30,6 @@ export function GoogleSeachWrapper() {
 
   //init
   useEffect(() => {
-    submitEvent('search-showing-results')
     loadSearchScripts()
     activateSearch()
   })
```
CodingDive, line 34: This effect will run on every render. Add an empty dependency array to only run it on init.

src/components/navigation/quickbar.tsx
```diff
@@ -63,17 +69,14 @@ export function Quickbar({ subject, className, placeholder }: QuickbarProps) {
 
   setOverlayPosition()
```
CodingDive, line 69: The `setOverlayPosition` function is directly mutating the style of a DOM node (referenced by a ref). In blocking rendering (the default before concurrent mode), each render always resulted in updating the UI, so this wasn't a problem in the past. However, concurrent mode (modern React) can start a rendering that might not commit fully, or it could be interrupted before finishing. This means the code that's running during render might run multiple times without the UI actually updating. Coupled with `getBoundingClientRect` which is a good use-case for a layout effect, this seems like a liability.

Wrapping this in a `useLayoutEffect` would be safer I believe. Right now it's being called every render (as it should as far as I understand) but the mutation on the ref is unsafe in concurrent react.

src/components/navigation/quickbar.tsx
```diff
@@ -212,78 +214,84 @@ export function Quickbar({ subject, className, placeholder }: QuickbarProps) {
...
-        })
+export async function fetchQuickbarData(subject?: string) {
+  const req = await fetch('https://de.serlo.org/api/stats/quickbar.json')
+  const data = (await req.json()) as QuickbarData
 
-        const subjectLower = subject?.toLowerCase() ?? ''
+  data.forEach((entry) => {
+    entry.pathLower = entry.path.map((x) => x.toLowerCase())
+    entry.titleLower = entry.title.toLowerCase()
+  })
```
CodingDive, line 227: I think this'd look better being more functional. I don't like mutations in `forEach` as they are generally harder to see. Just a matter of preference I suppose.

```suggestion
  data.map((entry) => ({...entry, 
    pathLower: entry.path.map((path) => path.toLowerCase()),
    titleLower:  entry.title.toLowerCase()
  }))
```

src/components/navigation/quickbar.tsx
```diff
@@ -38,7 +37,14 @@ export function Quickbar({ subject, className, placeholder }: QuickbarProps) {
...
 
   useEffect(() => {
-    if (query && !data) fetchData()
+    const effectFetch = async () => {
+      const fetchedData = await fetchQuickbarData(subject)
+      if (fetchedData) setData(fetchedData)
+    }
+
+    if (query && !data) {
+      effectFetch().catch(console.error)
```
CodingDive: I don't like the `effectFetch` name as a function should convey what it's doing and not in which context it's invoked. I propose a name like "fetchAndSetQuickbarData" or as it's only used in one place and you are using a similar name in the closure below, just get rid of the function.

```suggestion
   if (query && !data) {
     fetchQuickbarData(subject).then((fetchedData) => fetchedData && setData(fetchedData)).catch(console.error)
   }
```

src/components/navigation/quickbar.tsx
```diff
@@ -212,78 +214,84 @@ export function Quickbar({ subject, className, placeholder }: QuickbarProps) {
...
       }
     }
+    if (score > 0) {
+      score += Math.log10(entry.count)
+      results.push({ entry, score })
+      results.sort((a, b) => b.score - a.score)
+      results = results.slice(0, 7)
+    }
   }
+  return results
```
CodingDive, line 297: For performance reasons, I believe it's better to do the sorting once at the end (outside of the forof loop) 

```suggestion
     }
   }
  results.sort((a, b) => b.score - a.score)
  results = results.slice(0, 7)
  return results
```

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
```diff
@@ -0,0 +1,273 @@
...
+          </div>
+          {renderResults()} <br />
+        </>
+      ) : (
+        <>{renderSetEntry()}</>
+      )}
+    </div>
+  )
+
+  function renderInput() {
```
CodingDive: Should be just a component `<Input />` instead of `renderInput()`. Same with all the other presentational, functional components (renderResetButton / renderResults, etc.)

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
```diff
@@ -0,0 +1,273 @@
...
+  useEffect(() => {
+    const effectFetch = async () => {
+      const fetchedData = await fetchQuickbarData()
+      if (fetchedData) setData(fetchedData)
+    }
+
+    if (query && !data) {
+      // eslint-disable-next-line no-console
+      effectFetch().catch(console.error)
+    }
```
CodingDive: See comment above of function naming (either change the name and make it more descriptive, or just inline the function as suggested above)

CodingDive (commented):
Hey @elbotho. Really good code.
I added some comments and suggestions but overall, this is a much needed feature and I like your implementation. :raised_hands: 
I have checked out your branch and will do some local testing and will report back / comment if I find anything else.

src/components/navigation/quickbar.tsx
elbotho: While I agree with your reasoning about naming and unnecessary functions: wrapping async calls in a function is annoyingly the correct way of using async code inside of `useEffect` as far as I know. If you know a cleaner way without the linter screaming at me I'm very interested ;)

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
elbotho: well… this is a code style I adopted while working at serlo and we use this across the whole frontend. 
So changing that would be a massive refactor. 
I now actually like it, since it makes parsing a component a lot easier without spreading code across the repo that is only used in one place.

elbotho:
@CodingDive thanks for the review and the suggestions. I answered some above.
The rest does not directly concern this PR and would be more suited to a follow up (basically refactor quickbar) imo.
What do you think?

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -162,6 +162,9 @@ export function TextEditor(props: TextEditorProps) {
   }
 
   function handleEditableKeyDown(event: React.KeyboardEvent) {
+    // stop slate from loosing focus when using esc
+    if (event.key === 'Escape') return false
```
elbotho: > The only quickfix that I can make right now is that we blur completely on Escape.

Hej @hejtful,
following you comment I tried this for testing and it improved my keyboard editing experience a lot actually. Do you know of any cases where this might cause problems?

elbotho: (moved it further down so it should only stop the list plugin from stealing focus)

src/components/navigation/quickbar.tsx
hejtful: Hey @CodingDive, I didn't know about this, thanks for sharing! Could you maybe share a link to a docs page or a PR where it's explained in detail?

src/components/navigation/quickbar.tsx
hejtful: Yes, let's avoid mutations, they are a pain to figure out when they become bugs later 😄

src/components/navigation/quickbar.tsx
CodingDive: Somewhat old blog post from the React team https://legacy.reactjs.org/blog/2022/03/29/react-v18.html 

I can try to find a link where it's explained better but most of my React knowledge I learned from reading PRs of https://github.com/andarist :mage_man:

src/components/navigation/quickbar.tsx
CodingDive: Just using promises without async/await is the cleanest (edited my suggestion just now), but generally I also like async/await syntax better. I think in this example it fits well enough into a oneliner.

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
CodingDive: Although I don't like this code style too much, here in particular, I'm referring to the naming. It should be just a component that you can create even within other components. 
The React convention is to have `PascalCased` component names and it's clear when returning JSX that we're dealing with a React component. I'd only use the camelCased `renderComponentName` for render props.

hejtful: > and we use this across the whole frontend

While I fully appreciate this, it's not what we use across the editor codebase. We should either agree to respect that the editor codebase has a different style of component splitting, or refactor the whole editor codebase to use this `render*` approach.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: I don't see any problems with this, it's actually a super smooth solution, and I'm ashamed I didn't think of it myself 😄

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
```diff
@@ -0,0 +1,283 @@
...
+    )
+  }
+  function renderResetButton() {
+    return (
+      <div
+        className={clsx(
+          'absolute top-4 right-6 flex items-center justify-center w-8 h-8',
+          'serlo-button-editor-secondary',
+          'cursor-pointer text-gray-700'
+        )}
```
hejtful: Just out of curiosity, what is the need for using `clsx` here?

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
elbotho: Nice! Don't worry, you found all the pieces :)

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -59,19 +59,15 @@ export type TextEditorProps = EditorPluginProps<
 >
 
 export function TextEditor(props: TextEditorProps) {
-  const [hasSelectionChanged, setHasSelectionChanged] = useState(0)
-  const [isLinkNewlyCreated, setIsLinkNewlyCreated] = useState(false)
```
hejtful, line 62: Quick explanation of this workaround:

When updating Slate, the idea was to keep the editor functioning as it was before the Slate update. Before the update, the input inside the link controls overlay would only be focused when the link was newly created. If you click on an already existing link, that input would not be focused. Without the workaround, that input will be focused whenever a user clicks on a link element, or navigates to it using the keyboard.

While I hate this workaround and would love to see it go, I have to ask, are you sure you want to introduce this change? 😄

elbotho:
resolved comments are now linked in this issue: 
https://github.com/serlo/backlog/issues/133

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -157,7 +153,7 @@ export function TextEditor(props: TextEditorProps) {
         ({ value }) => ({ value, selection: previousSelection.current })
       )
     }
-    setHasSelectionChanged((selection) => selection + 1)
+    setSelectionChanged((selection) => selection + 1)
```
hejtful: Pls don't change this 🙏 

`selectionChanged` sounds like an event, like `mouseClicked` or `stateUpdated`

`hasSelectionChanged` sounds like a boolean, like `isButtonRed`. If you really dislike `has-` prefix, I'm ok with `isSelectionChanged`, but I'd really like to keep it obvious that this is a boolean flag.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -192,6 +188,26 @@ export function TextEditor(props: TextEditorProps) {
...
+            Object.hasOwn(parent, 'type') &&
+            parent.type === 'a' &&
+            Object.hasOwn(node, 'text') &&
+            node.text.length === offset
+          ) {
+            Transforms.move(editor)
+            event.preventDefault()
+          }
+        }
+      }
```
hejtful: For me, this doesn't work like the comment describes it. The link will be split, and the new line will also have a link. If I didn't understand it well, maybe just the comment could be a bit more specific about the behavior.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -358,32 +377,28 @@ export function TextEditor(props: TextEditorProps) {
...
+              config={config}
+            />
+            <HoveringToolbar
+              editor={editor}
+              config={config}
+              controls={toolbarControls}
+              focused={focused}
+            />
+          </>
+        )}
```
hejtful: Just out of curiosity, what does moving this block after the `Editable` do?

src/serlo-editor-repo/plugin-text/hooks/use-formatting-options.tsx
```diff
@@ -67,13 +65,7 @@ const isRegisteredTextPlugin = (
...
 }
 
-const toggleLinkAndFlag =
-  (setIsLinkNewlyCreated: SetIsLinkNewlyCreated) => (editor: SlateEditor) => {
-    toggleLink(editor)
-    setIsLinkNewlyCreated(true)
-  }
-
-const registeredHotkeys = (setIsLinkNewlyCreated: SetIsLinkNewlyCreated) => [
+const registeredHotkeys = () => [
```
hejtful: If you're sure we can remove the workaround, then this can be an array, it doesn't have to a function any more.

hejtful (commented):
In general, if we agree to keep the editor code style different from the rest of frontend, I'm also ok with refactoring the code style myself, so I didn't complain about that too much. 😄

src/components/navigation/quickbar.tsx
elbotho: While I try to avoid `.then`, I agree it's more concise here. Thanks.

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
elbotho: well we talked about keeping the tailwind lines from becoming super long and `clsx` is an easy way to do that.
But tbh. this is more a leftover when I had more rules in this className.

elbotho: (and at least I'm not at all consistent with that, some lines above is a super long className…)

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
elbotho: as discussed synchronously: I simplified this to rely on the `value`. If a link does not have a `href` set it is considered a new link and the input is focused. 
Let's see what testing shows, but I feel like this is as good or even better since we want people to either set a `href` or remove the link.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
elbotho: okay, glanced at some blog posts and I'm convinced :)
`isSelectionChanged` it is then.

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
elbotho: Before: 
When using `Tab` while the link popover is open (but not focused) the focus went to the settings button.
After 15 minutes of digging I found out that `Shift+Tab` would have set the focus to the link popover.

After:
Now when using `Tab` it moves the focus to the popover as expected. (Without needing JS to control the focus, instead relying on the normal tab order)

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
elbotho: Nice catch, thank you. I forgot to handle all cases. Could you the behaviour again if it works now and the comments are clear?

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: Amazing!

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
hejtful: It's clear now and works as described, thanks for updating!

src/serlo-editor-repo/plugin-text/components/text-editor.tsx
```diff
@@ -176,10 +176,7 @@ export function TextEditor(props: TextEditorProps) {
 
         if (node && parent) {
           if (Object.hasOwn(parent, 'type') && parent.type === 'a') {
-            if (
-              Object.hasOwn(node, 'text') &&
-              node.text.length - 1 === offset
-            ) {
+            if (Object.hasOwn(node, 'text') && node.text.length - 1 <= offset) {
```
elbotho: @hejtful I ran into an edge case here. I think when the link is the last item in a line i could not move out of the link with `ArrowRight`. This way works for me. Does my change makes sense to you?

src/serlo-editor-repo/plugin-text/components/link/link-content-search.tsx
elbotho: > it's not what we use across the editor codebase

That's a good point. Let's talk about it sometime. 
In general I love to save LOC and really dislike pushing values around the component tree that's why I like this unusual approach but it's also just a habit.

I wanted to split up the big component anyway and did so with **actual** components this time 😸. 
What do you think, code style close enough for now?

elbotho: > I wanted to split up the big component anyway and did so with actual components this time 😸.
What do you think, code style close enough for now?

ping @hejtful and @CodingDive

CodingDive: Yes, looks good to me! :+1: 

> In general I love to save LOC and really dislike pushing values around the component tree that's why I like this unusual approach but it's also just a habit.

Note that you can still define a component inside another component to reap the benefits of the closure (even though it's usually still better to create it outside especially when the outer component rerenders a lot e.g for input fields). I was mostly referring to the camelCase and render* component name instead of sticking with the React convention. Thank you for addressing and fixing it either way.

src/serlo-editor-repo/plugin-text/components/link/edit-mode/link-overlay-edit-mode.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <EditModeResultEntry
+              selectedIndex={selectedIndex}
+              chooseEntry={chooseEntry}
+              title={query}
+              pathHeader={config.i18n.link.customLink}
+              index={results.length}
+              isCustomLink
+            />
+          </div>
+          {serloLinkSearch && ( // de.serlo.org only for now
```
hejtful: `// de.serlo.org only for now` - If this is documented in more detail anywhere, it would be great to have a link in the comment as well. I'm just imagining that I joined Serlo today and read this comment, this "for now" would be confusing.

hejtful (approved):
Great work!

PR #2445, opened 2023-05-25
chore: add polyfill for array.at

### LINKED ISSUE #2442: missing polyfill for array.at
not supported by next.js or polyfilled - causes page crash in older browsers, found in donation-banner

elbotho:
@Entkenntnis do you have an easy way to test this? :) 

closes #2442


Entkenntnis (approved):
thanks. i stumbled across it on a very old browser. I haven't got a good idea yet how to test that ...

Entkenntnis:
using tools like browserling could help - tested it, and it works at least with chrome 72 (2019)

PR #2449, opened 2023-05-27
chore: move to node 18

elbotho:
node 16 is EOL in september:
https://nodejs.org/en/blog/announcements/nodejs16-eol


Entkenntnis (approved):
thank you

PR #2450, opened 2023-05-27
preview of prettier plugin tailwindcss

elbotho:


Entkenntnis:
@elbotho I have tested it out and it works after a small tweak.

I have also found a better way of splitting tailwind classes across lines with template literals:

grafik

the plugin is able to sort them, even across lines :tada: 

Conditional classes will stay with `clsx`, but it clsx is only used to split long lines, I will replace it with a template literal

src/helper/tw.ts
```diff
@@ -0,0 +1,11 @@
+// use this tagged template literal to enable formatting outside of className
+// works in vs code with this config:
+/*
+
+"tailwindCSS.experimental.classRegex": ["tw`([^`]*)"]
```
elbotho, line 7: does this mean everybody needs to add this? 
so we should probably put it in the wiki, right?

elbotho: very nice abstraction!

Entkenntnis: good idea, I have added it to https://github.com/serlo/frontend/wiki/VS-Code

elbotho:
okay then, let's create some merge conflicts 😉

Entkenntnis:
it would be best to first merge https://github.com/serlo/frontend/pull/2453, then merge staging into this branch

Entkenntnis:
I looks like we would need `tw` also for multi-line to avoid new lines ending up in the source code:

grafik

It's not breaking things, but it generates noise

Entkenntnis:
In thhe end, this is kinda a trade-off between improved DX (sorted multi-line tailwind, auto-complete + sorted anywhere outside of `className`) and bundle-size (not 100% sure what causes it, but the additional whitespace and template literals are added about +1kb)

This is not perfect yet, but an overall improvement, so I think it's ok for the moment.

Entkenntnis (approved):
self-reviewing :)

elbotho:
@Entkenntnis you manually added tabs to the template literals lines, right?

src/components/comments/dropdown-menu.tsx
```diff
@@ -61,10 +61,10 @@ export function DropdownMenu({
...
   return (
     <div
-      className={clsx(
-        'text-right bg-brand-50 py-3 pr-4 pl-2.5',
-        'shadow max-w-65 rounded-lg -mt-4'
-      )}
+      className={`
+        -mt-4 max-w-65 rounded-lg bg-brand-50
+        py-3 pr-4 pl-2.5 text-right shadow  
+      `}
```
elbotho: just to make sure I understand: did you forget the tw here or is it not necessary?

elbotho: In the comment of the helper it says it's just for use `outside of className` but you added it almost everywhere now. 
I'm confused :)

src/components/comments/comment-form.tsx
```diff
@@ -83,13 +83,10 @@ export function CommentForm({
         onKeyDown={onKeyDown}
         placeholder={placeholder}
         minRows={1}
-        className={clsx(
-          tw`
+        className={tw`
             serlo-input-font-reset w-full resize-none border-0 bg-transparent
-            text-lg text-black placeholder-brandgreen focus:!outline-none
-          `,
-          reply ? 'pr-14 pl-4' : 'pr-14 pl-4'
```
Entkenntnis: nice catch :)

src/components/comments/dropdown-menu.tsx
Entkenntnis: I forgot it here, thanks for finding this.

Adding it everywhere improves the rendered dom.

the newlines around the template literals are added manually. I haven't investigated it - but there could be other plugins to automatically format it ...

src/helper/tw.ts
```diff
@@ -1,4 +1,6 @@
 // use this tagged template literal to enable formatting outside of className
+// and to remove whitespace in multi-line mode
```
elbotho, line 2: thanks!

elbotho:
Okay, it's probably a bit confusing for new ppl, ~but so is tailwind~ but it's also easy to learn so… ;) 
Merge whenever you want.

PR #2453, opened 2023-05-31
Refactoring for edu-sharing integration

LarsTheGlidingSquirrel:
This pull request ...

- Moves the content in `tailwind.config.cjs` to `shared/tailwind-base.cjs` in order to make it available to the edu-sharing integration as a tailwind preset. Both frontend and edu-sharing integration use this preset as a base in their individual `tailwind.config.cjs`. However, for now they both do not overwrite this base config & both are styled the same. 

- Adds a config option to the box plugin to limit the allowed plugin types within the box. For example, in the edu-sharing integration we don't allow image plugins. 

- Adds a config option to the table plugin to prevent inserting images in the table cells. In the edu-sharing integration we don't allow images. 

- Restores the editor component `Renderer` that was deleted during the recent redux refactoring but is needed in the edu-sharing integration for rendering non-editable content. 

All these changes should have no effect on serlo.org. However, there might be some things that were affected. Would be great if you could double check if ...

- styling on serlo.org broke somewhere
- some tooling broke (e.g. autocomplete of serlo-* classes)


tailwind.config.js
```diff
@@ -1,229 +1,5 @@
...
-    while ((m = regex.exec(css)) !== null) {
-      // The result can be accessed through the `m`-variable.
-      components[m[0]] = {}
-    }
-
-    return components
-  } catch (error) {
-    // don't run on client, no problem
-  }
+  content: ['./src/**/*.{js,ts,jsx,tsx}', './shared/tailwind-base.cjs'],
```
Entkenntnis: @elbotho Do you remember why we included the tailwind config into the config path? I thought it was about some editor colors - but looking through the config this looks obsolete.

I think we can remove it.

Entkenntnis: digging through file history, I found out that the first versions looked like this: https://github.com/serlo/frontend/blob/81572b71c0f39127de61d59ed010b8b0ca5bd3dd/tailwind.config.js

I moved this out of the file, and now it's not necessary anymore - prefer `serlo-tailwind.css`.

Entkenntnis (approved):
thanks for this change, from frontend-perspective, everything works as expected (auto-complete, styles, ...)

I haven't looked to closely into the editor-part. @hejtful probably has a better overview there.

shared/README.md
elbotho, line 1: will there be a lot more? 
Otherwise I'd probably prefer a comment at the top of the file(s).

src/serlo-editor-repo/renderer/index.tsx
elbotho, line 1: maybe add a comment that this file is currently used in the edu sharing integration (so it's not removed again 😄)

elbotho (approved):
Looks great, thank you.
Added two small comments about comments.

However I did not check the tooling yet.

elbotho:
> for now they both do not overwrite this base config & both are styled the same.

What is the suggested workflow for changes/updates to this file?

tailwind.config.js
LarsTheGlidingSquirrel: Thanks for checking!

LarsTheGlidingSquirrel:
> What is the suggested workflow for changes/updates to this file?

@elbotho For now, we can simply make all changes & additions in `shared/tailwind-base.cjs`. If we want to style the edu-sharing editor differently at some point and cannot use overwrites, we can go ahead and move certain stuff to the `tailwind.config.cjs`. But for the foreseeable future I think this is not necessary.

src/edtr-io/plugins.tsx
```diff
@@ -211,7 +211,7 @@ export function createPlugins({
         plugin: 'text',
       },
     }),
-    box: createBoxPlugin(editorStrings),
+    box: createBoxPlugin({ editorStrings: editorStrings }),
```
elbotho: could be shortened to:  `box: createBoxPlugin({ editorStrings }),`

elbotho:
Ah since I worked on both of those plugins: The editor changes look fine.

Entkenntnis:
@LarsTheGlidingSquirrel you can go ahead an merge this if you are ready

shared/README.md
LarsTheGlidingSquirrel: Not right now. I changed it like you suggested.

PR #2455, opened 2023-06-01
fix(plugin-text): make slate readonly when text plugin is not editable

hejtful:


LarsTheGlidingSquirrel (commented):
Works like it is supposed to! Thanks! :)

LarsTheGlidingSquirrel (approved):

PR #2456, opened 2023-06-01
fix(editor): row and box styling in readonly mode

### LINKED ISSUE #150: Fix box styling issue in edu-sharing integration
The edu-sharing content can be opened in editor & read-only view, controlled by flag `mayEdit={true, false}` in url. 

The box styling is different between these two views: 

mayEdit=true 

Image

mayEdit=false

Image

For some reason this line renders additional html nodes that add margin when `mayEdit=true`: 
https://github.com/serlo/frontend/blob/d9b4917f084bc77bca8361600942972323b576a3/src/edtr-io/plugins/box/renderer.tsx#L76

hejtful:
resolves https://github.com/serlo/backlog/issues/150

### DIFF
--- a/src/edtr-io/plugins/box/renderer.tsx
+++ b/src/edtr-io/plugins/box/renderer.tsx
@@ -66,7 +66,7 @@ export function BoxRenderer(props: BoxProps) {
       <figcaption className="pt-1 flex text-lg">
         {isBlank ? null : (
           <div>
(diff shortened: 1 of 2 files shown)


Entkenntnis (approved):
thanks

LarsTheGlidingSquirrel (approved):
Thanks! Fixes the issue in the edu-sharing integration.

PR #2458, opened 2023-06-02
fix: Make box plugin only show empty content warning message when editable is true and box is empty.

LarsTheGlidingSquirrel:
### Before

Box showed this warning even when box is not editable in edu-sharing integration: 
image

Component used `useState` to show/hide this message and updated the state on key presses. However, deleting a plugin did not update this state -> Message should have been displayed but was not. 

### After

Warning is hidden when props.editable is false. 

Use `useAppSelector` instead of component state to check if the box content is empty and subscribe to re-render if the state changes.

### DIFF
--- a/src/edtr-io/plugins/box/renderer.tsx
+++ b/src/edtr-io/plugins/box/renderer.tsx
@@ -1,7 +1,6 @@
 import { selectIsEmptyRows } from '@edtr-io/plugin-rows'
-import { store } from '@edtr-io/store'
+import { useAppSelector } from '@edtr-io/store'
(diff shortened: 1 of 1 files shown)


Entkenntnis:
:tada: @elbotho you have worked on this, could be interesting for you

elbotho:
yes! way nicer. thank you.

elbotho (approved):

elbotho:
There are still some inconsistencies but they might be caused by `selectIsEmptyRows`.
e.g. when I delete a plugin the warning does not update.

PR #2473, opened 2023-06-09
replace shortid with uuid

elbotho:
for https://github.com/serlo/backlog/issues/176


kulla (approved):

src/serlo-editor-repo/plugin/child.tsx
```diff
@@ -1,5 +1,5 @@
-import * as R from 'ramda'
-import { generate } from 'shortid'
+import { mergeDeepRight } from 'ramda'
+import { v4 } from 'uuid'
```
LarsTheGlidingSquirrel, line 2: Suggestion: 
`import { v4 as uuidv4 } from 'uuid';` like it can be found in the docs.

LarsTheGlidingSquirrel (commented):
Inspected the store with the new ids and it seems to work fine. Only potential issue I can think of is that we now have "-" in the id string. I don't know for sure if that might cause issues somewhere, but probably not.

PR #2474, opened 2023-06-09
editor: persisting plugin id, expose id to frontend

elbotho:
@hejtful not a clean implementation at all but I think this would work for my purposes in https://github.com/serlo/backlog/issues/176.
What do you think?


Entkenntnis (approved):
I'm fine with these changes

elbotho:
try on staging

PR #2476, opened 2023-06-09
inelegant fix for isEmpty bug

### LINKED ISSUE #2475: editor: text with only math -> isEmpty is true
Text Plugin `isEmpty` returns true when the checked content consists of only a math element (inline or block).
As soon as a normal text is added it works as expected.

check:
[frontend/index.tsx](https://github.com/serlo/frontend/blob/staging/src/serlo-editor-repo/plugin-text/index.tsx#L56)

maybe related:
[frontend/math-element.tsx](https://github.com/serlo/frontend/blob/staging/src/serlo-editor-repo/plugin-text/components/math-element.tsx#L98)

elbotho:
https://github.com/serlo/frontend/issues/2475

### DIFF
--- a/src/serlo-editor-repo/plugin-text/index.tsx
+++ b/src/serlo-editor-repo/plugin-text/index.tsx
@@ -1,4 +1,4 @@
-import { Node } from 'slate'
+import { Editor as SlateEditor, Node } from 'slate'
 
 import { EditorPlugin, serializedScalar } from '../plugin'
 import { TextEditorProps, TextEditor } from './components/text-editor'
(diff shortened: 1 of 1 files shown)


LarsTheGlidingSquirrel (approved):
Thanks! Seems to work for now. 
I think it would be great to use [`Editor.isEmpty(...)` from slate](https://docs.slatejs.org/api/nodes/editor#manipulation-methods) if possible instead of doing it manually here. As far as I understood it, we could give it the entire editor object and it will go through the nodes checking each for isEmpty. However, this requires some more thought. For now, the fix works.

elbotho:
Nice, maybe that's what I was looking for.
> accounting for void nodes

sounds very promising.

elbotho:
@LarsTheGlidingSquirrel Since I could not find a solution to pass the `editor` into the `isEmpty` function without bigger changes I'll go ahead and merge the quickfix.

LarsTheGlidingSquirrel:
> @LarsTheGlidingSquirrel Since I could not find a solution to pass the `editor` into the `isEmpty` function without bigger changes I'll go ahead and merge the quickfix.

Yeah! I had the same issue when I tried a little. For now, it works and can be used.

PR #2482, opened 2023-06-12
refactor editor i18n, styles & icons

elbotho:
Based on [merge-serlo-plugins](https://github.com/serlo/frontend/tree/merge-serlo-plugins).

- simplifies config and handing down props
- also: some style refactoring (mostly replacing `style={{` and some `styled` with tailwind)
- refactor box plugin

([example editor link](https://frontend-git-refactor-plugins-serlo.vercel.app/entity/repository/add-revision/1555))


src/serlo-editor/plugins/injection/index.tsx
```diff
@@ -7,14 +7,11 @@ import { OverlayInput } from '@/serlo-editor/core'
...
   PreviewOverlay,
 } from '@/serlo-editor/editor-ui'
 import { EditorPluginProps, string, EditorPlugin } from '@/serlo-editor/plugin'
 import { Icon, faNewspaper } from '@/serlo-editor/ui'
 
-// this plugin is activly used in the serlo frontend
-
 /* global */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 declare const Common: {
```
elbotho: @Entkenntnis Do you know if this is still needed?

Entkenntnis: it's a leftover from the legacy integration and can be safely removed

elbotho:
Sorry @LarsTheGlidingSquirrel and @Entkenntnis another unwieldy one

src/serlo-editor/plugins/box/index.tsx
```diff
@@ -28,32 +21,35 @@ export function createBoxState(
...
+  allowedPlugins,
+} = defaultConfig): EditorPlugin<BoxPluginState> {
   return {
-    Component: BoxRenderer,
+    Component: BoxEditor,
+    state: createBoxState(allowedPlugins!),
     config: {},
-    state: createBoxState(editorStrings, allowPluginsWithin),
   }
 }
```
LarsTheGlidingSquirrel: Suggestion for getting rid of the not-null operator (!) and make typescript know that it cannot be null: 
``` typescript
const defaultAllowedPlugins = [
  'text',
  'image',
  'equations',
  'multimedia',
  'serloTable',
  'highlight',
]

export function createBoxPlugin({
  allowedPlugins = defaultAllowedPlugins,
}: BoxConfig): EditorPlugin<BoxPluginState> {
  return {
    Component: BoxEditor,
    state: createBoxState(allowedPlugins),
    config: {},
  }
}
```

src/serlo-editor/plugins/rows/components/rows-editor.tsx
```diff
@@ -53,14 +59,19 @@ export function RowsEditor(props: RowsProps) {
 
   const isDocumentEmpty = props.state.length === 0
 
+  const plugins = getPluginRegistry(
+    config.parentType,
+    editorStrings,
+    config.allowedPlugins
+  )
```
LarsTheGlidingSquirrel: I think this does not work for `serlo-editor-for-edusharing`. There we have additional plugins (for example the edusharing-asset) and define our own registry `serlo-editor-for-edusharing/src/frontend/plugins/registry.ts`. `getPluginRegistry` here cannot return the edusharing plugin. I guess we either need to add an entry for the edu-sharing asset also to the registry in `frontend` or provide an option to extend / overwrite the registry when calling `createRowsPlugin` like before. Unsure right now, what I would go for.

LarsTheGlidingSquirrel (commented):
Nice cleanup! Removing all the `i18n` properties and baking them in is great. However, I might have found an issue regarding `serlo-editor-for-edusharing`, see comment.

src/serlo-editor/plugins/rows/components/rows-editor.tsx
elbotho: Hmm, we could go back to the way it was, but tbh. I really like the simplified version :) 

Should we provide the registry as a prop to the `Editor` like the plugins?
``` tsx
<Editor  plugins={plugins}  pluginRegistry={…} {…} >
```

LarsTheGlidingSquirrel: > Should we provide the registry as a prop to the Editor like the plugins?

That would be great I think! At some point we could even change it so that we only pass one prop that contains everything.

LarsTheGlidingSquirrel: Great! I just tested this in `serlo-editor-for-edusharing` and it seems to work. However, I noticed a few weird things: 
- Plugin drag & drop does not work in the outermost scope. Within a box, it works. 
- Multimedia plugin has more padding and an outline when focused. 

I need to have a closer look into what is happening there.

elbotho: thanks for testing! 
I guess what you report is caused by my changes and does not have anything to do with the integration.

LarsTheGlidingSquirrel: In case you want to have a look, my try to get it working can be found here: https://github.com/serlo/serlo-editor-for-edusharing/tree/use-new-config-and-registry-and-more

elbotho: @LarsTheGlidingSquirrel could you confirm that the multimedia plugin issue also happens on staging and is not related to this PR?

elbotho: drag & drop problm should be fixed now, good catch!

LarsTheGlidingSquirrel: > @LarsTheGlidingSquirrel could you confirm that the multimedia plugin issue also happens on staging and is not related to this PR?

Its the same on staging, so it is unrelated to this pull request.

Entkenntnis (approved):
thank you for merging serlo-editor and frontend more and more.

only thing I noticed is `useLoggedInData()!`, but this is fine for the moment (maybe throwing within the hook could be another option ... in both cases, there would be an exception) - it's better than returning null though!

elbotho:
> only thing I noticed is useLoggedInData()!, but this is fine for the moment (maybe throwing within the hook could be another option ... in both cases, there would be an exception) - it's better than returning null though!

Yeah, In the frontend it might makes sense since we sometimes return null until the loggedInData is available (which is preferable to crashing) but here it does not make sense. 
Our `serlo-editor.tsx` already makes sure that this is available so maybe I add a `useEditorStrings` hook. Yeah I like that :)

elbotho:
@Entkenntnis thanks for the push, way nicer code now.

Entkenntnis:
Agreed!

PR #2488, opened 2023-06-15
editor: merge geogebra and video renderers

elbotho:
Another try to unify our renderers a bit more on two simple plugins.

- using same renderer for editor and frontend
- a little bit improved editor styles and ux (but that's not the main goal here)
- using different wrappers (privacy vs no interaction wrapper)


elbotho:
@Entkenntnis do you think a refactor of all plugins in this way makes sense as a step away from completely separated renderers?

PR #2493, opened 2023-06-16
Fix error message of h5p type and allow ImagePair plugin

CodingDive:
I have yet to get an okay from Tina or @Entkenntnis if it's okay to allow the ImagePair plugin.

However, I'd also like to fix the bug that the article can be saved while errors are in it. For this, I'll probably need some guidance but right now the local `error` state is not checked and one can save entities with an errored h5p plugin just fine.

Not sure if we should create another PR for this, or if it's a small change, add it to this one.

### DIFF
--- a/src/edtr-io/plugins/h5p.tsx
+++ b/src/edtr-io/plugins/h5p.tsx
@@ -17,6 +17,7 @@ const h5pLibraryWhitelist = [
   'H5P.Blanks',
   'H5P.DragText',
   'H5P.ImageHotspotQuestion',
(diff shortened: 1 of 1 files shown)


Entkenntnis:
don't forget to add imagePair to the overview list

CodingDive:
@Entkenntnis please review again and feel free to merge it as it is.
The fix for blocking `save` and displaying error messages in the modal when a plugin has an error will be added in a subsequent PR after I fixed some bugs and tested it thoroughly.

PR #2507, opened 2023-06-23
editor: refactor multimedia plugin

elbotho:
- adds shared renderer for frontend and editor
- simplifies editor code:
   - remove unused important config
   - remove drag&drop resizer ( another 10kb compressed gone from editor routes)
   - only two sizes are allowed (accessible via settings)

ping @LarsTheGlidingSquirrel I think you might need to change `plugins` to `allowedPlugins` in edusharing, sorry ☀️

[Editor-Preview](https://frontend-git-refactor-multimedia-serlo.vercel.app/entity/repository/add-revision/1555)


Entkenntnis:
looks like you forgot to run ´yarn`, lockfile outdated :)

src/serlo-editor/plugins/multimedia/renderer.tsx
```diff
@@ -0,0 +1,50 @@
+import clsx from 'clsx'
+
+export interface MultimediaRendererProps {
+  mediaWidth: number // 25 | 50 | 75 | 100 Percent
```
LarsTheGlidingSquirrel, line 4: Is this still relevant if the user can only set two sizes in the settings?

LarsTheGlidingSquirrel (approved):
Looks way clearer now. Also works in edu-sharing integration. Thank you!

src/serlo-editor/plugins/multimedia/renderer.tsx
elbotho: unfortunately yes, since this does not change currently stored values.

Entkenntnis (approved):
Thank you! Haven't noticed anything off, good to go.

PR #2508, opened 2023-06-23
editor: refactor spoiler

elbotho:
- shared renderer between frontend and editor
- improve styling

[example](https://frontend-git-refactor-spoiler-serlo.vercel.app/entity/repository/add-revision/1555)

### DIFF
--- a/src/assets-webkit/styles/serlo-tailwind.css
+++ b/src/assets-webkit/styles/serlo-tailwind.css
@@ -126,7 +126,7 @@
     @apply mx-side;
   }
   .serlo-spoiler-body {
-    @apply border-l-8 border-brand pt-6 pb-3.5;
+    @apply pt-6 pb-3.5;
   }
   .serlo-solution-box {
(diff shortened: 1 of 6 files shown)


Entkenntnis (approved):
looks good. I assume it's intended that you can't close the spoiler in editor?

elbotho:
> I assume it's intended that you can't close the spoiler in editor?

Yes, after fiddling with focus stuff for a while I simplified to "open when active, closed when not" and nothing else for now.

elbotho:
> looks good. I assume it's intended that you can't close the spoiler in editor?

Hmm thanks,
I think I intended it to close when not focused. I'll investigate again.

Entkenntnis:
I find it actually less distracting if it always stays open, because we have less layout shifts - so I thought it was a feature :)

elbotho:
well, I remember playing around with `focused` a bit and found it too unreliable (again)… so it's a feature after all 😂

PR #2512, opened 2023-06-23
editor: refactor highlight

elbotho:


Entkenntnis (approved):
thank, one duplication less

PR #2514, opened 2023-06-24
editor: replace deprecated with unsupported

### LINKED ISSUE #2489: editor: replace deprecatedPlugin
Current state:
- when editor encounters a plugin that is not supplied via `plugin` it just crashes
- pretty sure `deprecatedPlugin` is not in use.

Idea:
New `unsupported` plugin that gets displayed when the editor loads a plugin that is not supplied via `plugins`. Remove `deprecatedPlugin`.

in preparation of wiping the old table plugin :)

elbotho:
image

### Notes
- displays in place of any plugins that are not provided to editor
- instead of just crashing the editor it provides some user feedback
- Also removes `ErrorPlugin` that does basically the same with an error message related to migration

### Small downside
currently the actual state of the of the original plugin is not retrieved / displayed in `unsupported`.
That might be helpful for manual migrations but is not important at this point.

(closes #2489)

@LarsTheGlidingSquirrel you'll probably need to update the plugins in edusharing as well


Entkenntnis (approved):

Entkenntnis:
I hope we will not deprecate the image plugin :)

PR #2515, opened 2023-06-24
remove duplicate css from editor hacks

elbotho:
hopefully we can get rig of more soon, but for now this simplifies it quite a bit

### DIFF
--- a/src/assets-webkit/styles/serlo-tailwind.css
+++ b/src/assets-webkit/styles/serlo-tailwind.css
@@ -266,74 +266,40 @@
   }
 
   .serlo-editor-hacks {
+    @apply text-lg leading-cozy;
+
     h1 {
-      @apply mx-side mb-9 mt-4 p-0 text-3.5xl font-bold special-hyphens-auto;
+      @apply serlo-h1 special-hyphens-auto;
     }
     h2 {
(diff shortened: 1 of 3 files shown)


src/assets-webkit/styles/serlo-tailwind.css
```diff
@@ -266,74 +266,32 @@
...
-      }
-      & > li > ul,
-      & > li > ol {
-        @apply mt-2 !mb-4;
-      }
-    }
-
     blockquote {
-      @apply my-block mx-side border-l-6 border-brand-200 p-2.5;
+      @apply serlo-blockquote;
```
Entkenntnis: @elbotho Have you tested if this is working? I remembered that there where some limitations with applying custom classes.

You can add the class to the `blockquote-renderer` directly, though - or just delete it ...

Entkenntnis (approved):
Finally we can set classes to slate elements directly! this is making everything so much more pleasant.

src/assets-webkit/styles/serlo-tailwind.css
elbotho: added it directly, will do more cleanup here after this PR and also [this](https://github.com/serlo/frontend/pull/2527) is merged.

PR #2524, opened 2023-06-27
editor: refactor exercise plugin

elbotho:
- the fun part is in the template plugins so this has no shared renderer for now
- removing donations banner


Entkenntnis (approved):
jup, not much to share here atm

PR #2525, opened 2023-06-27
fix(geogebra-plugin): Make button 'activate' clickable when editable=false

LarsTheGlidingSquirrel:
## Before

When content is rendered by the editor with `editable: false` the button to 'activate' the plugin appears but cannot be clicked. 

This was an issue in `serlo-editor-for-edusharing`


elbotho (approved):
Thank you!

PR #2540, opened 2023-07-03
fix(plugin-text): merge plugins when list selected

hejtful:
**Bug**: When trying to merge a text plugin that has only a list, with a previous text plugin, by pressing "Backspace", an error is thrown and the editor breaks.

**Solution**: Prioritize removing list on Backspace over merging plugins, when a list is selected.


elbotho (approved):
second iteration works great 🎉
testing on staging now

PR #2541, opened 2023-07-03
feat(plugin-text): upgrade slate lists plugin dependency

hejtful:
Unfortunately, not a lot has been improved in the new version of slate lists plugin.

1. Blur on Esc key **has been removed**, so we can remove the workaround from our code.
2. Starting a list on Tab key has not been removed, so we still need the workaround in our code.
3. Slate normalization is still triggered on each keydown, so this plugin still causes problems for our undo/redo behavior.


elbotho (approved):
thanks for updating!

PR #2545, opened 2023-07-04
editor: refactor equations

elbotho:
Second time lucky. 

**Shares renderer between editor and frontend.**
Editing mode still uses "old" renderer because it differs quite a bit already and the whole editor code needs refactoring anyway. The focus code is pretty rough tbh.

Drag & Drop replaced by simple "move up" button again.

Some visual improvements.

(caveat: I think in the editor "preview" empty steps are not filtered as well as they are filtered in the frontend right now so here we might have a slight rendering difference, but it's a lot more accurate than before)


Entkenntnis (approved):
playaround a bit and looks good (no obivous regressions)

PR #2549, opened 2023-07-04
refactor(editor): use plugins from context in sagas

hejtful:


elbotho:
@hejtful: For me the perceived performance (without profiler etc.) is exactly the same between this branch and current staging.
I tested current versions of Chrome and FF on mac on both a [new article](http://localhost:3000/entity/create/Article/60730) and [an existing one](http://localhost:3000/entity/repository/add-revision/1555).

The performance is great when there is no/little content, and somewhat sluggish with some existing content.
That alone is worth investigating, but for me at least it's not at all related to this PR.

LarsTheGlidingSquirrel (approved):

elbotho (approved):

PR #2552, opened 2023-07-05
fetch new quickbar data via proxy

### LINKED ISSUE #2550: use new quickbar updater data
- add proxy to avoid passing ip adress to github
- test for regressions (data should be the same)
- check for all occurences in the frontend (quickbar, editor)
- make sure response is gzipped

https://github.com/serlo/quickbar-updater is mostly complete now

Entkenntnis:
caching is a bit problematic, stick with cf-worker instead: https://github.com/serlo/cloudflare-worker/pull/439

### DIFF
--- a/src/components/navigation/quickbar.tsx
+++ b/src/components/navigation/quickbar.tsx
@@ -218,7 +218,7 @@ export function Quickbar({ subject, className, placeholder }: QuickbarProps) {
 }
 
 export async function fetchQuickbarData(subject?: string) {
-  const req = await fetch('https://de.serlo.org/api/stats/quickbar.json')
+  const req = await fetch('/api/frontend/quickbar-data')
   const data = (await req.json()) as QuickbarData
 
   data.forEach((entry) => {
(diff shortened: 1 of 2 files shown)


elbotho:
> caching is a bit problematic

Just curious :) 
What went wrong? too much or too little caching?

Entkenntnis:
no support to cache `fetch` in api, so API would have to refetch every time, something like this ttl:

```
fetch(quickbarUrl, { cf: { cacheTtl: 24 * 60 * 60 } })
```

haven't really investigated it though ..

elbotho:
for the future :) 
- [serverless](https://vercel.com/docs/concepts/functions/serverless-functions/edge-caching) 
- [or with edge](https://vercel.com/docs/concepts/functions/edge-functions/edge-caching)

Entkenntnis:
I think we will switch to a frontend proxy when we add language versions, because otherwise it will be a bit annoying to sync the two implementations ..

PR #2553, opened 2023-07-05
bugfix: only show deleteOwnComment when the user is allowed to

MoeHome:
You are only able to delete a comment when it is an answer to another comment. But not if it is the start of a thread.

### DIFF
--- a/src/components/comments/dropdown-menu.tsx
+++ b/src/components/comments/dropdown-menu.tsx
@@ -90,7 +90,7 @@ export function DropdownMenu({
               : strings.comments.archiveThread}
           </>
         )}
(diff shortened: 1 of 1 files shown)


MoeHome:
In the API it is possible to delete your own comment also, when it is the start of a thread if there are no answers to it or all the answers are yours. Is there a good way of implementing that here?

Edit: Actually I do have an idea.

MoeHome:
And I don't know why the vercel error occurs.

Entkenntnis (approved):
thanks for the bugfix.

I have restarted vercel build and it should run fine now

Entkenntnis:
I'll merge it, as it's a fix for staging

PR #2556, opened 2023-07-06
Load iframe resizer script when rendering content-only

CodingDive:
This script is needed to communicate to the parent window (edusharing) and tell about it's size. This should fix the bug of the Serlo content being cut off at the bottom (https://github.com/serlo/serlo-editor-for-edusharing/issues/262) because the height would not adjust in some cases.
 
Even though I managed to lazy load the script and there should not be a bigger bundle size, I don't like this solution. I tried to adjust the edusharing code to not use the injection plugin, and instead render the content by itself which would potentially also fix some other problems that Lars and I have identified.
As it took me too long to figure out how to fetch and render the Serlo content directly, I fell back to this solution. As a consequence, the different-origin problems will remain but with some guidance, I'm happy to tackle it and use the renderer instead of the injection plugin.


src/components/content/lazy.tsx
```diff
@@ -21,7 +32,12 @@ export function Lazy(props: LazyProps) {
 
   // disable lazy load if content-only and probably within iframe
   if (router.pathname.startsWith('/content-only/')) {
-    return <>{props.children}</>
+    return (
+      <>
+        <LazyIframeResizer />
```
Entkenntnis: can you put this into `src/pages/content-only/[...slug].tsx`?

this place is only a fix for another problem and not directly related to the content-api. Adding it in the route is more reliable

Entkenntnis (changes requested):

Entkenntnis:
is this change sufficient to fix the bug?

CodingDive:
> is this change sufficient to fix the bug?

Lars and I were just talking about it. We can not test it within the PR deployment as the `content-only` does not work here.
To be honest, it may not fix the bug on first try, in which case we'll need to make 1-2 small changes that I have already lined up. It's really difficult to test but the importance is that we get the change(s) into frontend staging, so that we can test it from within the local edusharing integration (by pointing the iframe URL to staging).

Entkenntnis:
You could test it with https://frontend-git-fix-import-script-for-edusharing-serlo.vercel.app/content-only/1553 ... anyways, I have deployed, so you can merge this into staging

src/pages/content-only/[...slug].tsx
```diff
@@ -50,7 +50,7 @@ export default renderedPageNoHooks<SlugProps>(({ pageData }) => {
       authorization={pageData.authorization}
     >
       <Script
-        src="/_assets/iframeResizer.contentWindow.min.js"
+        src="iframe-resizer/js/iframeResizer.contentWindow.min.js"
```
Comment (Entkenntnis) [inline src/pages/content-only/[...slug].tsx]: this will attempt to load https://frontend-git-fix-import-script-for-edusharing-serlo.vercel.app/content-only/iframe-resizer/js/iframeResizer.contentWindow.min.js and fail

Entkenntnis (changes requested):

src/pages/content-only/[...slug].tsx
Comment (CodingDive) [reply src/pages/content-only/[...slug].tsx]: Nice catch, thank you! I didn't know you can't serve node_modules from within the <Script> tag. I reverted my change and now solve it via dynamic import which gives a little bit more flexibility to set up message listeners in the useEffect that I may need if it doesn't work already.

src/components/content/lazy.tsx
```diff
@@ -11,6 +12,16 @@ export interface LazyProps {
   noPrint?: boolean
 }
 
+export const LazyIframeResizer = dynamic(
```
Entkenntnis: Prefer putting unrelated components into separate files - e.g. lazy-iframe-resizer.tsx

Entkenntnis (changes requested):

CodingDive:
> You could test it with https://frontend-git-fix-import-script-for-edusharing-serlo.vercel.app/content-only/1553 ... anyways, I have deployed, so you can merge this into staging

I tested the edusharing against the feature branch. Thank you for the great suggestion. Now works 100% of the time, up from about 10-20% :tada: :partying_face:

LarsTheGlidingSquirrel (approved):

LarsTheGlidingSquirrel:
Thank you both for finding a way to fix this issue! :heart:

Entkenntnis (approved):

PR #2557, opened 2023-07-06
fix overlap in text-exercise template

### LINKED ISSUE #2555: regression: save bar is overlapping hovering toolbar
https://de.serlo-staging.dev/entity/create/Exercise/23869:

grafik

works in [production](https://de.serlo.org/entity/create/Exercise/23869).

@elbotho looks like some z-index issue ...

Entkenntnis:
fixes #2555


elbotho (approved):
Thank you for the fix. Also looks better imho.

PR #2562, opened 2023-07-07
geogebra: replace iframe

elbotho:
Trying out the direct route: instead of using the Iframe embedding this imports the applet directly via the script. 
Upsides: we can directly set some settings and sizing is more flexible.

example: https://frontend-git-geogebra-direct-load-serlo.vercel.app/43563

### DIFF
--- a/src/components/content/geogebra.tsx
+++ b/src/components/content/geogebra.tsx
@@ -19,7 +19,7 @@ export function Geogebra({ id }: GeogebraProps) {
         embedUrl={url}
         className="print:hidden"
       >
-        <GeogebraRenderer url={url} />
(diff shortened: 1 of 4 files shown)


Entkenntnis:
Direct load is great and gives us more control 👍 

want do you think about the behaviour on https://frontend-git-geogebra-direct-load-serlo.vercel.app/278642?

The size is still fixed, isn't it?

elbotho:
> The size is still fixed, isn't it?

The behaviour in this examples is like expected and discussed with Felix. 
The idea is to never scale up the applets since that is not intended by the authors. The applets should however not use the available space and scale down when needed.

Entkenntnis (approved):
Looks good!

PR #2564, opened 2023-07-09
refactor: editor plugin types

elbotho:
- remove some duplicated types
- Editor Plugin and Frontend Node types (e.g. 'article'): only use enum values instead of raw strings
- Use same strings for Editor Plugin and Frontend Node types
- `FrontendImgNode`  -> `FrontendImageNode`
- update to `typescript > 5` to use enum values in enum definition
- use editor state directly for plugins that don't have converted children (e.g. `anchor`, `geogebra`…)

@LarsTheGlidingSquirrel and again… please test if this breaks edusharing somehow. we can easily use looser types somewhere this is mostly in preparation to use editor state to render the frontend as well.
(and to make it more obvious what kind of `type` we are dealing with)


LarsTheGlidingSquirrel (approved):
Seems to work fine in serlo-editor-for-edusharing. However, working with the types makes it very clear that if we use `EditorPluginType` in more places (e.g. `allowedPlugins`) we need to figure out a way to define this type differently in `frontend` and `serlo-editor-for-edusharing`.

elbotho:
@LarsTheGlidingSquirrel as discussed I never changed the expected types to only `EditorPluginType`, it's either `EditorPluginType | string` or just `string` in some places so you should not have issues there.

Also I removed almost all of the raw type/plugin strings with either a `EditorPluginType` or a `FrontendNodeType` value.

LarsTheGlidingSquirrel (approved):
Nice! Still works with the new changes and is cleaner now. Just added a few type declarations in some places.

Entkenntnis (approved):
thanks! mostly enum stuff :)

PR #2567, opened 2023-07-10
fix(editor-toolbar): Improve z-index sorting of undo/save toolbar & text editor toolbar & tooltips & latex editor

LarsTheGlidingSquirrel:
## Before
Undo/save bar overlays elements & makes them look broken. 

Screenshot from 2023-07-10 16-26-12
Screenshot from 2023-07-10 12-58-21
Screenshot from 2023-07-10 16-31-30
Screenshot from 2023-07-10 19-43-43

## Changes
1) More margin at the top to prevent issue 1. However, it still happens in certain scroll positions of the content. Adding a high z-index to the tooltips does not work sadly because they are constrained by the z-index of a parent. 
2) Reduce z-index of undo&save bar to move it below text toolbar & latex editor to prevent issues 2-4

### DIFF
--- a/src/components/pages/add-revision.tsx
+++ b/src/components/pages/add-revision.tsx
@@ -120,7 +120,7 @@ export function AddRevision({
         )}${id ? ` (${id})` : ''}`}</title>
       </Head>
       {renderBacklink()}
-      <div className="controls-portal sticky top-0 z-[100] bg-white" />
+      <div className="controls-portal sticky top-0 z-[90] bg-white" />
       <div className="edtr-io serlo-editor-hacks mx-auto mb-24 max-w-[816px]">
         <SerloEditor
           entityNeedsReview={entityNeedsReview}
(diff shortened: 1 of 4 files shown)


elbotho:
Hej @LarsTheGlidingSquirrel,
this should be solved already (see https://github.com/serlo/frontend/issues/2555 and linked PR). 
Or did you find another case?

LarsTheGlidingSquirrel:
> Or did you find another case?

@elbotho I'll check if this is a new case. I found the issue on the following url but just now realized that this uuid 1000 is of type `user`. So testing here makes probably no sense at all. 
http://localhost:3000/entity/repository/add-revision/1000

LarsTheGlidingSquirrel:
@elbotho Pushed new attempt to improve this, see description at the top.

elbotho:
@LarsTheGlidingSquirrel thanks for investigating this. Let's just merge this to try.
It's somewhat likely this change will break something else but let's just see for now :)

PR #2571, opened 2023-07-11
feat(editor): plugin toolbar

### LINKED ISSUE #189: Problems with two consecutive multimedia plugins
**Problem**: when in the editor you have two consecutive multimedia plugins, it's quite difficult to add something in between. For example, when adding a text plugin in between, the text plugin overlaps a bit with the second multimedia plugin

**Request**: plugins are not overlapping and easy to recognize

**Source**: old backlog document

hejtful:
### testing
- [x] hand off to testing group (@elbotho)
- [x] run editor e2e tests against this branch and see if we need to adapt something when we merge to stagin

### general
- [x] visual bug: toolbar overlapps tooltips
- Refactor old toolbar
  - [x] find a way to remove settings icon for plugins with new toolbar (e.g. return null on renderIntoSettings…)
  - [x] find a different name for the "old" toolbar since we will still use it for dragging and maybe adding
- review nesting
  - [x] especially with the multimedia plugin
  - [x] build some weird nestings and see what happens :) 
  - [x] ~try fixed positioning~
- check template plugin types that use our child plugins:
    - [x] video
    - [x] applet
    - …?
- improve positioning / styling
- [x] remove setting icon and modal code

### image (done in https://github.com/serlo/frontend/pull/2632)
- [x] add toolbar.tsx
- [x] move all settings to new modal
- [x] add simple settings button for now
- [x] leave inline src for init state…

### multimedia
- ~how should toolbars look if they are restricted in width?~
- ~what nestings are allowed at all (best case)?~
- [x] could we leave them just overlapping for now?
- [x] move settings to toolbar (just a modal for now?)
- [x] add toolbar.tsx
- [x] simple parent indicator 

### box
- [x] retire generate id feature

### anchor
- [x] move input field into toolbar

**all plugins**
- make sure there is at least a toolbar with the plugin Name and ...-menu so users can remove plugins again (exerimental plugins, page plugins, …) (@AsterMiha)
 
## not todo (follow ups)
- no general parent indicator for now (let's see if we need it at all)
- all plugins that are not insertable in `page` or `article` (like the template plugins, …)
- Remove the old toolbar code (removed settings but we still need toolbar)
- `page-layout` needs parent indicator and same toolbar hacks as `multimedia`

## Status

Plugin toolbar applied to the following plugins so far:
- `Text`
- `Video`
- `Image` (isInlineChildEditor children)
- `Geogebra`
- `Highlight`
- `Injection`
- `Box`
- `Spoiler`
- `Anchor`
- `Equations` (isInlineChildEditor children)
- `SerloTable` (isInlineChildEditor children)
- `PageTeam`
- `PagePartners`
- `PageLayout`
- `Unsupported`
- `PasteHack`
- `InputExercise`
- `ScMcExercise`

## Bugs

…

## [Preview Link](https://frontend-git-feat-plugin-toolbar-serlo.vercel.app/entity/repository/add-revision/277232)


Entkenntnis:
@hejtful small note due to latest refactorings: `text-leaf-renderer` is now used in the frontend. Your https://github.com/serlo/frontend/pull/2571/commits/dcfd4c16c197c5bea7d4074309c9af255d76deb7 is causing a bundle size increase, because now the renderer is importing `utils/color.ts` which has a dependency on slate + co. (and tree-shaking within modules is not working too well)

You can put the colors into a separate file to enable better tree-shaking, this should remove the bundle size increase.

hejtful:
@Entkenntnis I just started wondering where does the size increase come from, and you already found the cause, amazing! Thanks a lot, I took your advice and moved it out to a constants file.

src/serlo-editor/plugins/multimedia/toolbar.tsx
```diff
@@ -105,7 +105,7 @@ export const MultimediaToolbar = ({
         </>
       }
       pluginControls={<PluginDefaultTools pluginId={id} />}
-      className="-top-[35px] left-[21px] w-[calc(100%-37px)]"
+      className="-top-[35px] left-[21px] !w-[calc(100%-37px)]"
```
elbotho: @elbotho investigate.

Entkenntnis:
To prepare you for some debugging: currently 11 out of 18 serlo editor e2e tests fail 🙈 

Most of the time it's about losing focus. Maybe it's not bad to start early, create a production build and test it locally a bit to get a feeling for it 🤞

Entkenntnis:
Luckly, most failures are caused by https://github.com/serlo/frontend/issues/2645 and fixing this would make most test cases workable again.

elbotho:
@hejtful found another image bug and decided to revert the absolutely positions inline-controls. 
I thought I could sneak this in but we just need a bigger refactoring here. Sorry for the extra loop.

elbotho (approved):

PR #2575, opened 2023-07-12
Update tailwind, replace custom classes with tw arbitrary values

elbotho:


Entkenntnis:
note: https://github.com/serlo/frontend/pull/2583 is happening in parallel - shouldn't cause too much trouble but keep it in mind and merge at some point

Entkenntnis (approved):
tailwind has come a long way :) thanks for removing the special classes with built-in ones

elbotho:
merged into https://github.com/serlo/frontend/pull/2583

PR #2583, opened 2023-07-13
Update dependencies (first batch)

LarsTheGlidingSquirrel:
## 1. Two major version bumps: 

**@next/bundle-analyzer**
Seems to work fine on pull request page. Any other place to check?  

**@svgr/webpack**
New config recommended by documentation of svgr. Tested and confirmed that svgs appear the same as before.


next.config.js
```diff
@@ -32,6 +44,7 @@ module.exports = withBundleAnalyzer({
   // TODO: reactStrictMode with react18 breaks edtr.io atm (inside react-hotkeys)
   reactStrictMode: false,
   productionBrowserSourceMaps: true,
+  transpilePackages: ['ramda'],
```
elbotho: why is this necessary?

elbotho (approved):
Thanks!

next.config.js
Entkenntnis: for context: https://github.com/vercel/next.js/issues/40183

elbotho: thanks!

PR #2593, opened 2023-07-14
RFC: use slate paragraphs and add soft breaks

elbotho:
I think this is the way we should have used paragraphs in slate in the first place ;) 
Not sure if changing this now makes sense, but I think the approach is cleaner.

Before: 
- We render the native slate paragraphs as `div` without margin and therefore basically use it as a line-container
- `shift+enter` triggers slate behaviour:p new Paragraph that gets displayed as new line.
- `enter` we create a new text-plugin after the current one

This PR:
- native slate paragraphs are rendered as `<p class="serlo-p">…` with proper spacing etc.
- `shift+enter` and `enter`:
   - if the current line is empty: new paragraph (basically pressing enter twice)
   - if it is not empty: soft break using `\n`
- `\n` should get serialized as `<br/>` (needs testing)
This way people can create Paragraphs (semantically and visually correct) and new lines without knowing about the difference between `enter` and `shift+enter`.

### DIFF
--- a/src/serlo-editor/plugins/text/components/text-editor.tsx
+++ b/src/serlo-editor/plugins/text/components/text-editor.tsx
@@ -239,8 +239,10 @@ export function TextEditor(props: TextEditorProps) {
             // soft break
             event.preventDefault()
             editor.insertText('\n')
+          } else {
+            // native behaviour (adds new paragraph) but remove empty line first
+            editor.deleteBackward('character')
           }
-          // otherwise native behaviour (adds new paragraph)
 
           // TODO: test <br/> serializer somehow
           // TODO: add placeholder (with add element that can split text plugin)
@@ -484,7 +486,7 @@ export function TextEditor(props: TextEditorProps) {
         )
       }
       return (
-        <p {...attributes} className="serlo-p">
+        <p {...attributes} className="serlo-p border border-gray-200">
           {children}
         </p>
       )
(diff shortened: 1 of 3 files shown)


Entkenntnis:
it feels good to have all paragraphs within one plugin.

one thing I would expect: If i press enter in an empty like and after the paragraph is created, I would expect the previous empty line to be removed, so from

grafik

to

grafik

elbotho:
> one thing I would expect: If i press enter in an empty like and after the paragraph is created, I would expect the previous empty line to be removed

good catch!
should be fixed now

Entkenntnis:
@elbotho you need to reopen and close the PR for new changes to take effect 

correction: vercel will still deploy, but github will not show further commits 😅

Entkenntnis:
reopen it to make it more visible :)

Entkenntnis:
Feels quite natural to me and the implementation looks not too complicated 👍

elbotho:
> Feels quite natural to me and the implementation looks not too complicated 👍

Thank you! Do you see any problems with the frontend rendering / spacing part (`FrontendSlateContainerNode` and `FrontendSlatePNode` etc)?


> reopen it to make it more visible :)

Since we focused on the toolbar in the meantime I'm okay with it being pending a bit :)

Entkenntnis:
I would have to think it through in detail one day - but it can't really get more complicated than it already is 😓 we will find a solution for it :)

elbotho:
> but it can't really get more complicated than it already is 

Good to hear ;) I think it might actually be easier this way.

Entkenntnis:
ok, closing again to reduce clutter

PR #2596, opened 2023-07-17
Make quickbar render on top of save button

CodingDive:
Instead of defining a z-index of 100 as I did here :zany_face: I'd like to know why the bar with the save button and undo/redo has such a high z-index. Inside the add-revision.tsx: `controls-portal sticky z-[90]`

**Before**

Screenshot from 2023-07-17 01-13-00

**After**

image


Entkenntnis (approved):
thanks, moving it up is working well

PR #2599, opened 2023-07-17
feat(plugin-image): add plugin-toolbar

hejtful:
One obvious bug: Plugin toolbar text control buttons active state is not consistent.


src/serlo-editor/plugins/text/components/hovering-toolbar-controls.tsx
```diff
@@ -31,10 +33,14 @@ export function HoveringToolbarControls({
...
             onMouseDown={(event) => {
               event.preventDefault()
+              event.stopPropagation()
               isNestedControlButton(control)
                 ? setSubMenu(index)
                 : control.onClick(editor)
             }}
+            onClick={(event) => {
+              event.stopPropagation()
+            }}
```
elbotho: not sure if both `stopPr…` are needed, did you test is further?
Also a short comment why we need this would help future me :)

src/serlo-editor/plugins/text/components/headless-text-editor.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }
+}
+
+export type HeadlessTextEditorProps = EditorPluginProps<
+  TextEditorState,
+  HeadlessTextEditorConfig
+>
+
+// TODO: Think of a better name, "headless" doesn't really make sense.
+// Something like "ChildTextEditor" or similar
```
elbotho: Have not thought it through, but so far we only use it for inline texts. Maybe `InlineTextEditor` would work?

elbotho (approved):
Nice 🎉
Helpful todos and comments as well.

Follow ups:
- Maybe we should write the "remove italic" todo down somewhere? (backlog issue?)
- Link Popover aligned to center right now
- block math elements are allowed

image
image

PR #2600, opened 2023-07-17
Some code documentation comments

LarsTheGlidingSquirrel:
## Before
The component name `DocumentEditor` can be very confusing: 
- It is below `InnerDocument`, `Editor` and `SubDocumentEditor` in the component hierarchy. 
- Is sounds like much but is just a container for plugin editor, toolbar and settings. 

## Changes
- Add code documentation comments


Entkenntnis (approved):
👍 definitively more expression :)

elbotho:
I like the change, thank you. 
Since @hejtful and me are working on the Plugin Toolbar right now over at https://github.com/serlo/frontend/tree/feat/plugin-toolbar maybe we will do the merging, okay?

LarsTheGlidingSquirrel:
> Since @hejtful and me are working on the Plugin Toolbar right now over at https://github.com/serlo/frontend/tree/feat/plugin-toolbar maybe we will do the merging, okay?

@elbotho Alright!

hejtful:
What do you think about `PluginSettings`, instead of `SettingsAndToolbar`?

For me, `SettingsAndToolbar` is quite general, and without context I'd guess that it means "editor settings". And as Botho mentioned, plugin toolbar will probably be moved out of this component soon.

LarsTheGlidingSquirrel:
> What do you think about PluginSettings, instead of SettingsAndToolbar?
For me, SettingsAndToolbar is quite general, and without context I'd guess that it means "editor settings". And as Botho mentioned, plugin toolbar will probably be moved out of this component soon.

@hejtful Having one component called `PluginToolbar` and one `PluginSettings` would be even better! If you are working on it right now, do you want to continue with this pull request? Or should I make the change?

hejtful:
> Having one component called PluginToolbar and one PluginSettings would be even better! If you are working on it right now, do you want to continue with this pull request? Or should I make the change?

I can do it after we finish with the PluginToolbar, but it won't be so soon. Maybe we can keep this PR open as a reminder, and in case something changes with the PluginToolbar, I guess we can then merge this PR as is. What do you think?

LarsTheGlidingSquirrel:
> I can do it after we finish with the PluginToolbar, but it won't be so soon. Maybe we can keep this PR open as a reminder, and in case something changes with the PluginToolbar, I guess we can then merge this PR as is. What do you think?

I reverted the renaming and added a TODO comment instead.

elbotho:
@LarsTheGlidingSquirrel little thing: We use ` TODO: … ` not `@TODO`
This way they also get picked up by eslint.

LarsTheGlidingSquirrel:
> @LarsTheGlidingSquirrel little thing: We use TODO: … not @TODO
This way they also get picked up by eslint.

Ah, good to know. I'll change them.

PR #2602, opened 2023-07-18
upgrade graphql related dependencies

### LINKED ISSUE #2601: upgrade graphql related dependencies
some examples:

grafik

elbotho:
somehow there problems with anonymous queries/mutations but everything else seems to work fine.
@LarsTheGlidingSquirrel maybe you just merge this into you other upgrades?


LarsTheGlidingSquirrel:
> somehow there problems with anonymous queries/mutations but everything else seems to work fine.

Does this mean, it can not be merged like it is right now? Do we have to put more work into making the anonymous stuff work?

Entkenntnis:
I see a 21kb bundle size increase that we may have to investigate. The only really relevant package is `graphql-request`, but I can't see any bigger changes ...

I will investigate it a bit

elbotho:
> Does this mean, it can not be merged like it is right now? Do we have to put more work into making the anonymous stuff work?

No, giving them a (arbitrary) name did the trick. 
The Bundle size is more of a blocker… Thank you @Entkenntnis for investigating.

elbotho:
@Entkenntnis at least according to https://bundlephobia.com/ there are no mentionable size increases in `graphql-request` or `graphql`.

Entkenntnis:
the solution could be simpler - we have `@graphql-codegen/typescript-graphql-request` configured, but are not using it at all. This is causing the bundle size increase.

I'm removing it and preparing a commit.

@elbotho @LarsTheGlidingSquirrel Would you mind merging this PR into staging directly (and rebasing the other PR)? I would like to run e2e tests against it (+ let more people use it  and let issues surface faster)

Entkenntnis (approved):
Fine for me

PR #2605, opened 2023-07-19
refactor(editor): move plugin toolbar controls to editor-ui

hejtful:
Plugin toolbar text controls (ex `HoveringToolbarControls`) will be used in multiple plugins, like Text, Image, Table, Box etc. Therefore, moving them to editor-ui directory and renaming them makes sense.


elbotho (approved):
Thanks!

PR #2611, opened 2023-07-19
refactor(editor): move plugins context one level up

hejtful:
In order not to re-create the plugins object after creating it, moving the plugins context one level up, so it wraps the `InnerDocument` component, makes sense.


CodingDive (approved):
Looks great!

elbotho (approved):

PR #2617, opened 2023-07-20
remove last styled components 🎉

elbotho:
based on `feat/plugin-toolbar`.
Sorry for polluting the PR. 

Tailwind is never pretty, but it's especially ugly here. Don't look too close 😉
Especially in parts of the Editor that we will replace anyway I removed styled quite ruthlessly.


Entkenntnis (approved):
thanks for the PR, didn't expect it to happen so fast

elbotho:
> didn't expect it to happen so fast

Yeah, procrastinating on easy stuff instead of tackling the hard stuff ;)

hejtful (approved):
The room is so clean 🥹

PR #2638, opened 2023-07-23
feat(editor): anchor link tool for plugin toolbar

elbotho:
todo: 
- test that it does not display in edusharing and newly created content (without uuid)
- test new and existing [pages](/entity/repository/add-revision/18922/269138)
- …


src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const serloEntityId = useEntityId()
+  const editorStrings = useEditorStrings()
+  const { strings } = useInstanceData()
+
+  if (
+    !serloEntityId ||
+    !navigator.clipboard ||
+    !window.location.href.includes('add-revision')
+  )
+    return null
```
hejtful: I really dislike omitting curly braces when it goes to a new line like this, and I'd really appreciate adding curly braces here. The reason is that it's so easy to implement a bug by adding a line before the return statement line.

Or maybe:
```suggestion
  const isAddRevisionPage = window.location.href.includes('add-revision')
  if (!serloEntityId || !navigator.clipboard || !isAddRevisionPage) return null
```

But not a deal breaker, for sure.

src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <DropdownButton
+      onClick={() => {
+        const url = `https://serlo.org/${serloEntityId}#${
+          pluginId.split('-')[0]
+        }`
+        void navigator.clipboard.writeText(url)
+        showToastNotice(strings.share.copySuccess, 'success')
+        showToastNotice('👉 ' + editorStrings.edtrIo.anchorLinkWarning)
```
hejtful, line 36: A more accessible way to show emojis in React would be something like:
```
<span role="img" aria-label={editorStrings.handWithIndexFingerPointingRight}>👉</span>
```
but that would require adjusting the `showToastNotice` to accept HTML, so I guess this is alright for now.

hejtful (approved):

src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
elbotho: Interesting. Could you point me to a source for that?
I know this approach for icons, but never saw it for emoji. It was my understanding that emoji have an alt text baked in already (that might not fit the purpose).
Also the emoji in the toasts are mostly presentational so I'm not sure if it actually would help with a11y in those cases.

elbotho: I'd rather hide them with `aria-hidden="true"` but as you said converting notices to html is out of scope here ^^

hejtful: The best "official" source I can provide is: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#using_roleimg_to_confer_meaning_that_is_obscured_or_implied

So the issue for us would be that it's in English. But I agree with you, in this case, just `aria-hidden="true"` would probably be the best, the emoji is not crucial for understanding the message.

And also agree that it's out of scope, so I'll resolve the comment.

elbotho: thank you for the input!

Issue #2639, opened 2023-07-24
open solutions in review


Entkenntnis:
you need to open the solution https://de.serlo.org/entity/repository/compare/64282/279491 in review.

expected: automatically open the spoiler

PR #2641, opened 2023-07-24
Deployment

Entkenntnis:
- fix: start solution unfolded in revisions ([issue](https://github.com/serlo/frontend/issues/2639))
- fix: pasting images, images urls or geogebra urls adds the plugins again
- fix: close plugin suggestions menu on esc
- add input exercise event


src/serlo-editor/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -140,6 +141,13 @@ export const useSuggestions = (args: useSuggestionsArgs) => {
     SELECT_UP: handleSelectionChange('up'),
     SELECT_DOWN: handleSelectionChange('down'),
     INSERT: handleSuggestionInsert,
+    ESCAPE: () => {
+      // We delete the line with the slash and therefore, close the suggestions
+      // upon next render
+      if (closure.current.showSuggestions) {
+        editor.deleteBackward('line')
+      }
+    },
```
hejtful, line 150: To stay in line with the current code style, I'd prefer to declare this as a function, and then the comment would also not be necessary.

```suggestion
    ESCAPE: handleCloseSuggestions,
```

hejtful (approved):

PR #2642, opened 2023-07-24
refactor(text-plugin): Change focus and cursor placement slightly

LarsTheGlidingSquirrel:
This hopefully resolves the issues with the e2e tests failing after updating dependencies. 

The e2e test that failed all had to do with the cursor placement when creating a new text plugin. After updating the dependencies, the cursor was wrongfully placed before the `/` resulting in some failed tests. 

Locally, the e2e test now pass with these changes and updated dependencies. I don't understand it fully but it may have been a weird race condition that was somehow affected by updating our dependencies.

Concerning my changes: I figured it makes most sense to first focus the text editor and then move the cursor. Also, everything wrapped in setTimeout to make `ReactEditor.focus(editor)` work.
 
Merging this can be postponed until other work on the editor (toolbar changes) is completed. 

@hejtful Would be great if you could have a look at it.

### DIFF
--- a/src/serlo-editor/plugins/text/components/text-editor.tsx
+++ b/src/serlo-editor/plugins/text/components/text-editor.tsx
@@ -101,47 +101,48 @@ export function TextEditor(props: TextEditorProps) {
 
   // Workaround for setting selection when adding a new editor:
   useEffect(() => {
-    // Get the current text value of the editor
-    const text = Node.string(editor)
-
-    // If the editor is not focused, remove the suggestions search
-    // and exit the useEffect hook
-    if (focused === false) {
-      if (text.startsWith('/')) {
-        editor.deleteBackward('line')
-      }
-      return
-    }
-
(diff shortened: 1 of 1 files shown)


CodingDive (approved):
Just tested it locally and works great for me. :tada: Thanks!

Only focus issue that I ran into that has nothing to do with your changes is that when creating a plugin, then undoing, the focus is at the start of the line and not at the end (basically the undo breaks focus and the cursor is placed to the left of the / again).

hejtful (approved):

Entkenntnis (approved):
Change looks good to me. @elbotho Do you see if this would conflict with the current work on the plugin-toolbar?

If not we can proceed and merge this.

elbotho:
Hmmm, I don't expect a conflict here 👍

hejtful:
I also don't think we changed this part, so this particular PR probably won't create conflicts 😄

PR #2643, opened 2023-07-24
refactor(multimedia-plugin): split plugin toolbar into smaller components

hejtful:


src/serlo-editor/plugins/multimedia/toolbar/toolbar-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ReactNode, useState } from 'react'
+
+import { FaIcon } from '@/components/fa-icon'
+import { ModalWithCloseButton } from '@/components/modal-with-close-button'
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
+import { PluginToolbar } from '@/serlo-editor/editor-ui/plugin-toolbar'
+import { PluginDefaultTools } from '@/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-default-tools'
+import { EditorPluginType } from '@/serlo-editor-integration/types/editor-plugin-type'
+
+interface Props {
```
elbotho: take it or leave it. We stick quite well to the pattern that the interface for `ToolbarWrapper` is called `ToolbarWrapperProps`

src/serlo-editor/plugins/multimedia/toolbar/toolbar-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+  id: string
+  children: ReactNode
+  strings: ReturnType<typeof useEditorStrings>
```
elbotho: is there an upside to passing it on?
I usually prefer to use the hook again in the child component so we don't have to pass too much stuff down but that might also be a matter of taste.

src/serlo-editor/plugins/multimedia/toolbar/toolbar-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const ToolbarWrapper = ({ id, children, strings }: Props) => {
```
elbotho: Ah, I expected `ToolbarWrapper` to wrap something around the toolbar.

elbotho: Instead it wraps the toolbar around the content of the settings modal, right?

src/serlo-editor/plugins/multimedia/toolbar/size-select-button.tsx
```diff
@@ -0,0 +1,35 @@
+import clsx from 'clsx'
+
+import { MultimediaProps } from '..'
+import { tw } from '@/helper/tw'
+
+interface Props {
```
elbotho: see other comment

src/serlo-editor/plugins/multimedia/toolbar/type-select.tsx
```diff
@@ -0,0 +1,76 @@
...
+  selectSerializedDocument,
+  store,
+  useAppSelector,
+} from '@/serlo-editor/store'
+
+interface Props {
+  allowedPlugins: MultimediaProps['config']['allowedPlugins']
+  state: MultimediaProps['state']['multimedia']
+  strings: ReturnType<typeof useEditorStrings>
+}
```
elbotho: see other comments

elbotho (approved):
Thank's for splitting this 🪵.
I added some comments but it's just code style stuff (take it or leave it).

src/serlo-editor/plugins/multimedia/editor.tsx
```diff
@@ -53,24 +53,31 @@ export function MultimediaEditor(props: MultimediaProps) {
...
     <div className="group/multimedia">
-      {editable && focused ? (
-        <MultimediaToolbar
-          {...props}
-          showSettingsModal={showSettingsModal}
-          setShowSettingsModal={setShowSettingsModal}
-        />
-      ) : null}
-      {editable && !focused ? (
+      {editable && focused && (
```
elbotho: ah, missed that one. 
I always liked the brevity of this but got convinced by I think Jonas to always use the `isTrue ? <>…</> : null` pattern instead just to never run into annoying bugs ([some random article on it](https://www.crocoder.dev/blog/react-conditional-rendering/)). 
In this case it should never matter so I don't mind, but in general this is why I stick with the longer ternary.

elbotho: 🐊

hejtful: Oh, good to know! I'll keep that in mind and try to stick to the longer version in the future!

Issue #2646, opened 2023-07-24
e2e testing (plugin-toolbar): Setting url of image plugin fails


Entkenntnis:
link: https://frontend-git-feat-plugin-toolbar-serlo.vercel.app/entity/create/Article/1337

![grafik](https://github.com/serlo/frontend/assets/13507950/1ca66bf4-81c3-4fff-a59f-3804def39120)

Typing any letter in this input field leads to unexpected behaviour: The plugin loses focus, I can't edit url anymore:

![grafik](https://github.com/serlo/frontend/assets/13507950/4c90a0f2-f6aa-4644-9fcc-4777325a80a6)

elbotho:
nice one botho 🤦

hejtful:
@elbotho Added a proposal PR, but if you have another idea, that's totally fine with me.

PR #2647, opened 2023-07-25
fix(plugin-image): display src field idea

hejtful:
Fix proposal for https://github.com/serlo/frontend/issues/2646

With this fix, 16 tests pass and 2 fail.

### DIFF
--- a/src/assets-webkit/styles/serlo-tailwind.css
+++ b/src/assets-webkit/styles/serlo-tailwind.css
@@ -115,6 +115,17 @@
   .serlo-img {
     @apply inline h-auto max-w-full rounded-md;
   }
+  .serlo-img-error-placeholder {
+    @apply relative min-h-[100px] min-w-[100%];
+  }
+  .serlo-img-error-placeholder::before {
+    @apply absolute left-0 top-[50%] h-full w-full translate-y-[-50%] border-2 border-editor-primary-100 bg-slate-100;
+    content: '';
+  }
+  .serlo-img-error-placeholder::after {
+    @apply absolute left-0 top-1 top-1 block h-full w-full;
+    content: '\2639'' ' attr(alt);
+  }
   .serlo-math-wrapper {
     @apply px-4 py-0.5 text-left;
     @apply overflow-x-auto overflow-y-hidden text-lg leading-browser;
--- a/src/assets-webkit/styles/serlo-tailwind.css
+++ b/src/assets-webkit/styles/serlo-tailwind.css
@@ -115,17 +115,6 @@
   .serlo-img {
     @apply inline h-auto max-w-full rounded-md;
   }
-  .serlo-img-error-placeholder {
-    @apply relative min-h-[100px] min-w-[100%];
-  }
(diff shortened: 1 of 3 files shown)


elbotho:
> because we currently display image toolbar inside of image plugin, even when it's a child of multimedia. (is this a bug or intended behavior?)

Yes this is intended. We could create logic that combines multimedia + media-child toolbar buttons but then we will run into problems with space I'm afraid (image toolbar will grow in the future :) )

src/serlo-editor/plugins/image/renderer.tsx
```diff
@@ -33,7 +33,7 @@ export function ImageRenderer({
         {wrapWithLink(
           placeholder ?? (
             <img
-              className="serlo-img"
+              className="serlo-img serlo-img-error-placeholder"
```
Entkenntnis: you can add tailwind directly here.

src/assets-webkit/styles/serlo-tailwind.css
```diff
@@ -115,6 +115,17 @@
...
+    @apply relative min-h-[100px] min-w-[100%];
+  }
+  .serlo-img-error-placeholder::before {
+    @apply absolute left-0 top-[50%] h-full w-full translate-y-[-50%] border-2 border-editor-primary-100 bg-slate-100;
+    content: '';
+  }
+  .serlo-img-error-placeholder::after {
+    @apply absolute left-0 top-1 top-1 block h-full w-full;
+    content: '\2639'' ' attr(alt);
+  }
```
Entkenntnis: the styles here are mainly for components that are used in multiple places and need overriding attributes. For one time usage you can directly use tailwind.

Entkenntnis (commented):
small hint

src/serlo-editor/plugins/image/renderer.tsx
elbotho: @hejtful I'm pretty strict about adding editor stuff into the renderer. The renderer should remain as "pure" as possible.
What we need in the long term is a smarter Image component that parses the url, tries to load the image etc…
In the short term something like that would be okay, but maybe you could add a comment about that.

src/assets-webkit/styles/serlo-tailwind.css
elbotho: maybe relevant: https://tailwindcss.com/docs/content  :)

src/serlo-editor/plugins/image/renderer.tsx
elbotho: @hejtful adding it in this div here should also work maybe?
https://github.com/serlo/frontend/blob/c3fb04da40b1b16e6a61c41dc1ed69b1451626df/src/serlo-editor/plugins/image/editor.tsx#L57

hejtful: > adding it in this div here should also work maybe?

I couldn't make this work, unfortunately. Added a comment, though.

Entkenntnis (approved):
tailwind looks good now

PR #2649, opened 2023-07-25
fix(editor): text plugin image paste immutable

hejtful:
Got an idea how to fix [this bug](https://trello.com/c/4wgdq5FD/190-pasting-image-url-in-text-plugin-does-not-create-image-any-more) easily. A bit ashamed I didn't think of it straight away 😄 

The issue was that Redux state was being mutated by direct assignment. As we already have setter methods defined, I just used those instead.

\+ some Mac specific yarn cache files committed.


Entkenntnis (approved):
nice fix, it works for me and it get's to upload the image

elbotho:
love it :D

PR #2650, opened 2023-07-25
fix(text-editor): pasting image and geogebra urls creates plugins again

elbotho:


hejtful (approved):

Entkenntnis (approved):
Changes look good - haven't tested it manually though.

(maybe a candidate for a e2e test, because this feature seems to break every now and then)

elbotho:
yeah, I added it to our testing document for now. If @yanncloud is done with the current tests he could add them maybe.

hejtful:
Nice, I had no idea this was ever a feature or even possible

elbotho:
Found it in the old edtr-io text plugin, but it was as bit less obvious there :)

PR #2653, opened 2023-07-26
feat/restrict plugin nesting

hejtful:


src/serlo-editor/plugins/rows/utils/nesting.ts
elbotho, line 1: consider renaming to `check-is-allowed-nesting.ts`

elbotho (approved):
this turned out great 🎉
nice and understandable

PR #2598, opened 2023-07-17
Hotkey lib migration

CodingDive:
Closes https://github.com/edtr-io/edtr-io/issues/381

Attempt to make #2595 slimmer. Ctrl+Z / Ctrl+Y does not work in the editor so I'm marking it as a draft until I have found the fix.

PR #2655, opened 2023-07-26
Close plugin suggestion menu upon pressing escape

CodingDive:
With the new keyboard library https://github.com/serlo/frontend/pull/2598 this code should get even nicer as we don't need the "handleHotKeys" code that is just there to prevent default behavior of the event, anymore.


elbotho (approved):
nice, code looks good and works as expected for me.
Thanks!

PR #2656, opened 2023-07-26
style(plugin-text): restructure use suggestions hook code

hejtful:
@CodingDive Just a minor code style suggestion to:
- Not split code into inline hotkey handler *and* handler function - keep it all inside the handler function instead
- Declare hotkey package config as a constant
- A few super super minor changes

Feel free to close the PR if you don't like it, of course.


CodingDive:
Thank you! :tada:

PR #2659, opened 2023-07-27
fix(editor): unfocus plugins on click outside of editor

### LINKED ISSUE #255: Plugin selection menu can't be easily closed by clicking outside of the component
Steps to reproduce
1. Get the plugin suggestion menu to open
2. Try to click outside of it
3. It sometimes jumps visually but it takes a double click or the newly implemented ESC key to really close it

Tested locally, staging and in production, all environments seem to have the same issue.

plugin_suggestion

hejtful:
resolves https://github.com/serlo/backlog/issues/255


CodingDive (approved):
Works great for me! Thank you for the quick fix :tada: 
As the null focus resets the focus entirely, upon pressing tab afterwards, it starts the beginning of the page with the first tabIndex item (usually the navigation bar). I think this is to be expected with the current solution, but let's keep it in mind to reset the focus within the editor itself upon reworking the focus logic.

elbotho (approved):
Thanks for the PR.
I agree with @CodingDive point about tab order.
since this will mostly affect different users I guess it should seldomly lead to problems.

PR #2662, opened 2023-07-27
Fix equations editor unnecessary rerender

LarsTheGlidingSquirrel:
Just two small fixes: 
1. Equations editor no longer rerenders when focus tree changes. Instead focus tree is obtained only when needed. 
2. Change `@TODO` to `TODO:` so that eslint catches it


elbotho (approved):

hejtful (approved):

elbotho:
@LarsTheGlidingSquirrel thank you for the changes, makes sense!
@hejtful obviously thought so aus well and fixed the same thing in `feat/plugin-toolbar` already. Closing in favour of #2571 😄

PR #2666, opened 2023-07-27
Change contact person to Alexandra

CodingDive:
Upon request of Anne.

Is it true that CrowdIn will adjust the other translation files, or should I change them manually?


Entkenntnis (changes requested):
Thank you for tackling this. All german translations are managed by [crowdin](https://crowdin.com/project/serlo) - you should be able to add the translations there.

Never change de files manually :)

PR #2671, opened 2023-07-31
fix(editor): serlo-table can't be focused

hejtful:


elbotho (approved):
not pretty, but works nicely :)

PR #2672, opened 2023-07-31
Remove problematic page with id "49982" from SSG

CodingDive:
This page is cursed, having a Unicode character (ellipsis … unicode U+2026) that crashes the database layer or api (when reading from Redis). Will probably need to write a proper migration for it :crying_cat_face: 

By commenting out the lines, at least the frontend gets to build and we can continue developing and testing until we find a proper fix for it :shrug:


AndreasHuber:
Not required because the cache in staging now has the UUID with the problematic character changed.

PR #2680, opened 2023-08-01
Add data-qa attributes

CodingDive:
Small changes to the frontend as discussed with Botho and outlined here https://github.com/serlo/documentation/wiki/Best-Practices#testing-1 to add qa classNames to certain elements that we are clicking on in the serlo-editor tests. 

Depends on https://github.com/serlo/frontend-e2e-tests/pull/54. Please merge both at the same time, or we'll have some failing tests.

### DIFF
--- a/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
+++ b/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
@@ -255,7 +255,7 @@ function createToolbarControls(
         isActive: isHeadingActive(level),
         onClick: toggleHeading(level),
         renderIcon: () => (
-          <span className={`qa-plugin-toolbar-heading-${level}`}>H{level}</span>
+          <span data-qa={`plugin-toolbar-heading-${level}`}>H{level}</span>
         ),
       })),
     },
--- a/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
+++ b/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
@@ -249,12 +249,14 @@ function createToolbarControls(
       isActive: isAnyHeadingActive,
       renderIcon: () => <EdtrIcon icon={edtrText} />,
       renderCloseMenuIcon: () => <EdtrIcon icon={edtrClose} />,
-      children: ([1, 2, 3] as const).map((level) => ({
+      subMenuButtons: ([1, 2, 3] as const).map((level) => ({
         name: TextEditorFormattingOption.headings,
         title: `${textStrings.heading} ${level}`,
         isActive: isHeadingActive(level),
         onClick: toggleHeading(level),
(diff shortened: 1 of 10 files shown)


elbotho:
Looks good, thank you! It's definitely more sane that usind the tailwind classes as selectors in the tests!

I since learned that at least @hejtful (and maybe others) would prefer the `data-qa` approach, so maybe we have to discuss this in the team before we make it a best practice.

Entkenntnis (approved):
thanks

src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
```diff
@@ -269,7 +271,7 @@ function createToolbarControls(
         return <ColorTextIcon color={color} />
       },
       renderCloseMenuIcon: () => <EdtrIcon icon={edtrClose} />,
-      children: [
+      subMenuButtons: [
```
hejtful, line 274: Thanks for this!

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -426,7 +426,7 @@ export function TextEditor(props: TextEditorProps) {
               <TextLeafRenderer {...props} />
             </span>
           )}
-          className="[&>[data-slate-node]]:mx-side [&_[data-slate-placeholder]]:top-0" // fixes placeholder position in safari
+          className="qa-plugin:text-editor [&>[data-slate-node]]:mx-side [&_[data-slate-placeholder]]:top-0" // fixes placeholder position in safari
```
hejtful: This colon here already seems like you have some convention in mind. Could you maybe explain in the [Best Practices page](https://github.com/serlo/documentation/wiki/Best-Practices#testing-1) why you sometimes use a colon and sometimes only dashes?

hejtful (approved):
As @elbotho mentioned, I prefer `data-qa`, and the teams in [Cypress](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) and [Codecept](https://codecept.io/best/#locators) also seem to prefer it 😄 However, I'm ok with this as well, no need to change it all over again if there is no obvious problems with this already implemented approach.

CodingDive:
> As @elbotho mentioned, I prefer data-qa, and the teams in [Cypress](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) and [Codecept](https://codecept.io/best/#locators) also seem to prefer it smile However, I'm ok with this as well, no need to change it all over again if there is no obvious problems with this already implemented approach.

I'm actually changing this now, because if we don't do it early on, it'll be such a hassle to refactor later and the customLocator you sent me is too sexy to miss out on https://codecept.io/plugins/#customlocator.

Thanks for the feedback! I'll ping the two of you, once I have made the changes.

src/serlo-editor/plugins/text/components/text-editor.tsx
CodingDive: Nice find. As this is the only plugin I gave a data-qa attribute too for now, I just changed it to a hyphen.

elbotho (approved):

PR #2681, opened 2023-08-01
Fix/arrow keys in link suggestion overlay

### LINKED ISSUE #2670: editor: link suggestion arrow keys don't work any more
image

e.g. https://de.serlo-staging.dev/entity/repository/add-revision/1555 but also in production.

Up/down arrows don't do anything. Pretty sure this worked at some point.

CodingDive:
Fixes https://github.com/serlo/frontend/issues/2670 

There were actually two small issues. I broke it by not disabling the global arrow keys (thinking of finding a better solution for this), but also: the highlighting didn't work with the arrow keys, making it seem like nothing is selected.  

I also want to add an e2e test for this.

### DIFF
--- a/src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
+++ b/src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
@@ -24,7 +24,7 @@ export function EditModeResultEntry({
       className={clsx(
         'serlo-link flex cursor-pointer bg-white px-side py-2',
         'hover:!bg-editor-primary-100 hover:!no-underline',
-        index === selectedIndex && 'bg-editor-primary-100'
+        index === selectedIndex && '!bg-editor-primary-100'
       )}
       onClick={() => {
         chooseEntry(index)
--- a/src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
+++ b/src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
@@ -23,8 +23,8 @@ export function EditModeResultEntry({
       key={index}
       className={clsx(
         'serlo-link flex cursor-pointer bg-white px-side py-2',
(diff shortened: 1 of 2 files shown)


src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
```diff
@@ -23,8 +23,8 @@ export function EditModeResultEntry({
       key={index}
       className={clsx(
         'serlo-link flex cursor-pointer bg-white px-side py-2',
-        index === selectedIndex && 'bg-editor-primary-100 group-hover:bg-white',
-        'hover:!bg-editor-primary-100 hover:!no-underline'
+        'hover:!bg-editor-primary-100 hover:!no-underline',
+        index === selectedIndex && '!bg-editor-primary-100'
```
CodingDive, line 27: I removed the `group-hover:bg-white` as I didn't really understand what it does. Also had to add the !, but I couldn't find exactly where the styles get overwritten. Possibly the `serlo-link` but it's the first className added, so definitely shouldn't have any precedence over the crucial `bg-editor-primary-100` :thinking:

CodingDive:
I don't know if this was related to any of my changes, but in a moment I had some z-index 0 text heading appear over the z-index 120 link suggestion modal. :grimacing: Thought I'd mention it here. 

Screenshot from 2023-08-01 20-28-12

src/serlo-editor/plugins/text/components/link/edit-mode/edit-mode-result-entry.tsx
elbotho: that's a tailwind thing. 
Ther is a lib calles `tailwind-merge` that has an explainer: https://github.com/dcastil/tailwind-merge/blob/v1.14.0/docs/what-is-it-for.md
So far we did okay without it but it explains the problem :)

elbotho (approved):
Thank you!

PR #2684, opened 2023-08-02
fix(editor): conditionally enable Redux serializableCheck

hejtful:


src/serlo-editor/store/store.ts
```diff
@@ -20,7 +20,9 @@ export const store = configureStore({
   middleware: (getDefaultMiddleware) =>
     getDefaultMiddleware({
       // https://github.com/serlo/backlog/issues/127
```
elbotho, line 22: Maybe we should update the comment here? Or would you keep the issue open?

elbotho (approved):
👍

src/serlo-editor/store/store.ts
hejtful: It would be good to keep it open, but update it once we merge this approach to staging, so that we continue striving to remove even the exceptions.

elbotho: > so that we continue striving to remove even the exceptions

did some digging yesterday. Quite hard to understand passing around of functions, but we can give it a go.

PR #2685, opened 2023-08-02
fix(editor): use blur on outside click hook triggered by click on modal

hejtful:


elbotho (approved):
thanks, good enough until we find a more permanent solution 👍

PR #2687, opened 2023-08-03
fix(editor): huge amount of `HoverOverlay` re-renders

hejtful:
First win of React strict mode: we got a warning about a huge amount of re-renders of the `HoverOverlay` component!

It happened when opening the suggestions overlay, and then typing a bit, like first typing "Text", but then backspacing and typing "Multimedia" instead. With each keystroke, the amount of re-renders would exponentially increase, creating very obvious performance issues.

The removed code was my quick fix for a link settings overlay positioning issue. However, probably due to all the refactors done by @elbotho , we no longer have any positioning issues that require the code removed in this PR.

### DIFF
--- a/src/serlo-editor/editor-ui/hover-overlay.tsx
+++ b/src/serlo-editor/editor-ui/hover-overlay.tsx
@@ -16,21 +16,6 @@ export function HoverOverlay(props: HoverOverlayProps) {
 
   const windowSelection = window.getSelection()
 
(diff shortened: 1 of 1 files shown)


CodingDive (approved):
Amazing! Maybe wait for Bothos thoughts before merging if the `selectionChange` listener is not needed indeed.

elbotho (approved):
Seems to work fine, thank you!

PR #2688, opened 2023-08-03
refactor(editor): simplify selectAncestorPluginIds

elbotho:
not super importatn but since we don't use it…


hejtful (approved):
giphy (2)

PR #2689, opened 2023-08-03
test(editor): add data-qa and aria attributes for testing multimedia …

hejtful:
…plugin settings

### DIFF
--- a/src/serlo-editor/plugins/multimedia/toolbar/size-select.tsx
+++ b/src/serlo-editor/plugins/multimedia/toolbar/size-select.tsx
@@ -14,7 +14,7 @@ export const MultimediaSizeSelect = ({
 }: MultimediaSizeSelectProps) => (
   <div className="mt-8">
     <b className="mb-4 ml-0 mt-6 block text-lg font-bold">{title}</b>
-    <ul className="flex pb-8">
+    <ul className="flex pb-8" role="listbox">
       {mediaColumsSizes.map((percent: number) => (
         <li key={percent}>
           <MultimediaSizeSelectButton state={state} percent={percent} />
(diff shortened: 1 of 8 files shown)


CodingDive (approved):
Neat!

src/serlo-editor/plugins/multimedia/toolbar/size-select.tsx
```diff
@@ -14,7 +14,7 @@ export const MultimediaSizeSelect = ({
 }: MultimediaSizeSelectProps) => (
   <div className="mt-8">
     <b className="mb-4 ml-0 mt-6 block text-lg font-bold">{title}</b>
-    <ul className="flex pb-8">
+    <ul className="flex pb-8" role="listbox">
```
CodingDive, line 17: Why is this a listbox? Should it have keyboard navigation too? I can only see two items in that list and dont really know if it would be considered a listbox. If it is, we could think of adding `aria-selected="true/false"` to the `MultimediaSizeSelectedButton`

hejtful: What the user chooses here is how big the multimedia section in the Multimedia plugin will be.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role#description
> The listbox role is used to identify an element that creates a list from which a user may select one or more static items

By this definition, it seems to me like it is a listbox, because the user can select one or more static items. Maybe I misunderstood it, in that case, could you please explain my mistake?

And `aria-selected` is already added to the buttons, [inside their own files](https://github.com/serlo/frontend/pull/2689/files#diff-c69a277f14e52206cf91bd62b7b66dade45926c0c684de4503f5d482c8e9c0bdR31).

CodingDive: Perfect!

PR #2693, opened 2023-08-03
editor: replace store focus

elbotho:
New `domFocusWithin` props for plugins
- uses state instead of store
- (currently runs completely in parallel to focus in store)
- uses `onFocus` & `onBlur` to make sure it works with keyboard events as well
- plugin-toolbar modals are now children of the plugin itself to simplify focus
- removes some annoying focus hacks

**issue / todos**
- [ ] serloTable: bluring by outside click does not work when cell's text editor is focused (no blocker)
- [ ] Inline text editor: Enter does not work any more 🤦 (shift+enter does)
- [ ] make sure new plugins get correct focus
- [ ] make sure entities get correct focus after loading
- [x] rows: find simple solution for visibility of add button ([move add row button from rows editor to sub-document editor PR](https://github.com/serlo/frontend/pull/2771))
- [x] multimedia+box: when clicking the gray parent button the plugin toolbar stays active even when selecting other plugins ([proposed fix PR](https://github.com/serlo/frontend/pull/2770))
- [x] multimedia: should use `domFocusWithin` to show the "focus parent" button as well
- [x] serloTable: clicking on inactive table does not immediately focus clicked cell (somehow this worked before)
- [x] serloTable: activate slate on tab again
- [x] serloTable: use `selectIsFocused` in `cell-switch-button.tsx`
- [x] useEnableEditorHotkeys does currentyl not work with new `domFocus`
- [x] focusing images (and sometimes equations, …?) does not work rn (broke somwhere in the last few commits)
- [x] ~just found a critical bug. Activating slate sometimes takes two clicks or looses cursor position.
=/~

**differences:**
- ~multimedia currently still uses old `focused` for now (bugs or special nesting case)~
- ~there are some double toolbars now, but that's actually okay for the moment and changing it should be easy~ added some parent select buttons


src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -74,6 +76,23 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
...
+    ? document.plugin.startsWith('type-') ||
+      [EditorPluginType.Article, EditorPluginType.Rows].includes(
+        document.plugin as EditorPluginType
+      )
+    : true
+
+  const handleDomFocus = useCallback(() => {
+    const target = containerRef.current
+    if (!target) return
+    setTimeout(() => {
```
hejtful: >Activating slate sometimes takes two clicks or looses cursor position.

Removing this `setTimeout` wrapper resolves this bug. What problem does this `setTimeout` solve? Maybe we can think of another solution to that problem?

hejtful: Just saw you changed it, nice 👍

elbotho:
closing this while we figure out how to best procede.

elbotho:

PR #2694, opened 2023-08-03
fix(editor): remove HoverOverlayOld

hejtful:


elbotho (approved):
best solution, thank you! 💚

PR #2695, opened 2023-08-03
docs(plugin-text): README update

hejtful:


elbotho (approved):
thanks for writing this AND keeping up to date. 
may we all be a bit more like @hejtful 😅

PR #2696, opened 2023-08-04
fix(plugin-text): avoid TextEditor re-render when Slate re-renders

hejtful:
Please see the removed README section for why a workaround was implemented.

When working on https://github.com/serlo/frontend/pull/2695, I got inspired again to try and find a better solution for that problem.
Turns out that we can use `useSlate` hook in all `Slate` component's children safely*, and avoid passing down the `editor` object as props. This way the `LinkControls` component renders based on `Slate` renders, meaning that we can remove the `isSelectionChanged` workaround, which forces the `TextEditor` component to render when its `Slate` child renders. 

End result: less code and less renders.

\* The `useSlate` hook uses React Context, so it will use the nearest Context Provider, which will always be its parent `Slate`'s context.


elbotho (approved):
Thanks you. Great improvement 🎉
I was wondering at some point why we don't use useSlate at all, probably it did not exist when edtr was written :)

PR #2697, opened 2023-08-04
Close math editor through ESC and new button

CodingDive:
Trying to understand slate transformations. I don't get how I can move the cursor, or why the X button needs a small timeout around `Editor.focus(editor)` for the selection to be reset, and therefore, the modal with the LaTeX formula to be closed. :face_with_spiral_eyes: 

- [x] Clean up code
- [x] Test more edge cases
- [x] Create a card in Trello for it 
- [x] Remove 2px margin with !important of the texarea
- [x] Maybe in follow-up PR: Move components to separate files and renderFns to distinct components
- [x] Figure out why the Redo test is failing for me locally (might be unrelated to my changes here) :crying_cat_face:

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -97,7 +97,6 @@ export function MathEditor(props: MathEditorProps) {
     Key.Escape,
     (event) => {
       event.preventDefault()
-      // close overlay
       props.closeMathEditorOverlay()
     },
     {
@@ -297,7 +296,7 @@ export function MathEditor(props: MathEditorProps) {
             {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
           </p>
           <button
-            onClick={() => props.closeMathEditorOverlay()}
+            onClick={props.closeMathEditorOverlay}
             className="mr-0.5 mt-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
             aria-label="Close math formula editor"
             data-qa="plugin-math-close-formula-editor"
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -51,7 +51,7 @@ const MathEditorTextArea = (props: MathEditorTextAreaProps) => {
   return (
     <EditorTextarea
       className={tw`
-        !m-0.5 h-24 !w-[80vw] !max-w-[600px] rounded-md !border-2
+        mx-0 my-1 h-24 !w-[80vw] !max-w-[600px] rounded-md !border-2
         border-transparent text-black !shadow-none focus:border-editor-primary
       `}
       onChange={parentOnChange}
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -1,5 +1,5 @@
 import { faCheckCircle, faCircle } from '@fortawesome/free-regular-svg-icons'
-import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons'
+import { faQuestionCircle, faXmark } from '@fortawesome/free-solid-svg-icons'
 import clsx from 'clsx'
 import { useState, useCallback, createRef, useEffect } from 'react'
 import { createPortal } from 'react-dom'
@@ -77,8 +77,8 @@ export interface MathEditorProps {
   onEditorChange(visual: boolean): void
   onInlineChange?(inline: boolean): void
   onChange(state: string): void
-
-  onMoveOutRight: (options?: { closeThroughModal?: boolean }) => void
+  closeMathEditorOverlay: () => void
+  onMoveOutRight: () => void
   onMoveOutLeft(): void
   onDeleteOutRight?(): void
   onDeleteOutLeft?(): void
@@ -98,7 +98,7 @@ export function MathEditor(props: MathEditorProps) {
     (event) => {
       event.preventDefault()
       // close overlay
-      props.onMoveOutRight()
+      props.closeMathEditorOverlay()
     },
     {
       enableOnFormTags: true,
@@ -297,11 +297,12 @@ export function MathEditor(props: MathEditorProps) {
             {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
           </p>
           <button
-            onClick={() => props.onMoveOutRight({ closeThroughModal: true })}
-            className="mr-0.5 mt-1 text-sm font-bold text-gray-600"
+            onClick={() => props.closeMathEditorOverlay()}
+            className="mr-0.5 mt-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
+            aria-label="Close math formula editor"
             data-qa="plugin-math-close-formula-editor"
           >
-            x
+            <FaIcon icon={faXmark} />
           </button>
         </div>
         {!isVisualMode && (
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -297,8 +297,8 @@ export function MathEditor(props: MathEditorProps) {
           </p>
           <button
             onClick={props.closeMathEditorOverlay}
-            className="mr-0.5 mt-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
-            aria-label="Close math formula editor"
+            className="serlo-button-editor-secondary py-0"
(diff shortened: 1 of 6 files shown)


src/serlo-editor/plugins/text/components/math-element.tsx
```diff
@@ -149,50 +185,28 @@ export function MathElement({
...
       Transforms.delete(editor, { unit, reverse })
     }
 
-    ReactEditor.focus(editor)
+    // When calling this function when the 'x' button of the modal is clicked,
+    // a small timeout is needed to reset the selection. Transforms.deselect()
+    // was not needed
+    if (closeThroughModal) {
+      setTimeout(() => {
+        ReactEditor.focus(editor)
```
CodingDive: How could I move the cursor to the right (end of line)? I tried to many things in slate but none of them worked. Now the cursor goes to the left side of the formula which I think is okay, but to the right side of the formula seems more fitting and would be the same behavior as when pressing the ESC key.

elbotho: (ping @hejtful maybe? 🙃)

src/serlo-editor/math/editor.tsx
```diff
@@ -89,7 +93,19 @@ export function MathEditor(props: MathEditorProps) {
 
   const { visual, readOnly, state, disableBlock } = props
 
-  const useVisualEditor = visual && !hasError
+  useHotkeys(
+    Key.Escape,
+    (event) => {
+      event.preventDefault()
+      // close overlay
+      props.onMoveOutRight()
```
elbotho, line 101: we already destructure props above, maybe add it here and use `onMoveOutRight()`?

src/serlo-editor/math/editor.tsx
```diff
@@ -260,16 +280,24 @@ export function MathEditor(props: MathEditorProps) {
...
+          <p className="mr-0.5 mt-1 text-right text-sm font-bold text-gray-600">
+            {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
+          </p>
+          <button
+            onClick={() => props.onMoveOutRight({ closeThroughModal: true })}
+            className="mr-0.5 mt-1 text-sm font-bold text-gray-600"
+            data-qa="plugin-math-close-formula-editor"
+          >
+            x
+          </button>
```
elbotho: Sorry, a11y veto here :) 

If you really want to add this button:
- please add hover state
- add [hidden label](https://tailwindcss.com/docs/screen-readers) for screen readers

Also for consistency I'd prefer to use an icon here instead of `x` like we do for the other close buttons.

src/serlo-editor/plugins/text/components/math-element.tsx
```diff
@@ -31,6 +31,12 @@ export function MathElement({
   const editor = useSlate()
   const selected = useSelected()
   const preferences = useContext(PreferenceContext)
+  const visualModePreferences = !!preferences.getKey(visualEditorPreferenceKey)
+  const [isVisualMode, setIsVisualMode] = useState(visualModePreferences)
```
elbotho: could you explain why you moved this into state?

elbotho (changes requested):
code looks good, added one question and one change request.

src/serlo-editor/plugins/text/components/math-element.tsx
CodingDive: Oops, I'll revert that change. This was when I thought the visualMode is actually the one to render the LaTeX modal. Nice catch! It can remain a computed property

src/serlo-editor/plugins/text/components/math-element.tsx
hejtful: On it, I'll try to find a solution and either give a proposal here or open a sub-PR if I make a working solution while digging through Slate docs 😁

hejtful: None of my experiments worked, sorry. Here is an interesting discussion on exactly this topic, maybe you'll have more luck @CodingDive : https://github.com/ianstormtaylor/slate/issues/3412

src/serlo-editor/math/editor.tsx
```diff
@@ -74,8 +77,9 @@ export interface MathEditorProps {
   onEditorChange(visual: boolean): void
   onInlineChange?(inline: boolean): void
   onChange(state: string): void
-  onMoveOutRight?(): void
-  onMoveOutLeft?(): void
+
+  onMoveOutRight: (options?: { closeThroughModal?: boolean }) => void
```
hejtful: I'd prefer if you would define a new handler, like `onMathEditorClose` or similar, and then use that new one for the Esc hotkey and close button. Reasons:
1. `onMoveOutRight` is a very specific handler for MathQuill package
2. A new handler could have a descriptive name that explains what it handles, instead of the vague `onMoveOutRight`

Will continue this comment in another place.

src/serlo-editor/plugins/text/components/math-element.tsx
```diff
@@ -149,50 +185,28 @@ export function MathElement({
   function transformOutOfElement({
```
hejtful: If you defined a new handler (or two) for strictly using Esc and close button to close the modal, you could avoid the long comment down there, a third argument to this function, and instead of one clunked function, you would have multiple smaller ones which do simple things that are easy to explain in one function name.

This `transformOutOfElement` function was too big already before this PR, as it handles left arrow key, right arrow key, backspace key, and delete key. And all of that for not one, but two different components. 😅  So making it do even more seems like a bad idea to me.

hejtful (commented):
Just one (long) comment, hope you agree with my reasoning there

src/serlo-editor/plugins/text/components/math-element.tsx
elbotho: (seems to be solved by removing all additional hacks 😅)

src/serlo-editor/math/editor.tsx
```diff
@@ -98,7 +98,7 @@ export function MathEditor(props: MathEditorProps) {
     (event) => {
       event.preventDefault()
       // close overlay
-      props.onMoveOutRight()
+      props.closeMathEditorOverlay()
```
hejtful: No need for the `// close overlay` comment any more, I think.

src/serlo-editor/math/editor.tsx
```diff
@@ -297,11 +297,12 @@ export function MathEditor(props: MathEditorProps) {
             {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
           </p>
           <button
-            onClick={() => props.onMoveOutRight({ closeThroughModal: true })}
-            className="mr-0.5 mt-1 text-sm font-bold text-gray-600"
+            onClick={() => props.closeMathEditorOverlay()}
```
hejtful:
```suggestion
            onClick={props.closeMathEditorOverlay}
```

Just to avoid creating a new function on each render

hejtful (approved):
Two super minor comments just for clean-up, but looks good to me already!

src/serlo-editor/math/editor.tsx
```diff
@@ -267,19 +286,28 @@ export function MathEditor(props: MathEditorProps) {
...
-        </p>
-        {!useVisualEditor && (
+        <div className="flex items-center justify-between">
+          <p className="mr-0.5 mt-1 text-right text-sm font-bold text-gray-600">
+            {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
+          </p>
+          <button
+            onClick={props.closeMathEditorOverlay}
+            className="mr-0.5 mt-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
+            aria-label="Close math formula editor"
```
elbotho: follow up: i18n or borrow shorter close string from e.g. `strings.share.close`.

src/serlo-editor/math/editor.tsx
```diff
@@ -267,19 +286,28 @@ export function MathEditor(props: MathEditorProps) {
...
-          {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
-        </p>
-        {!useVisualEditor && (
+        <div className="flex items-center justify-between">
+          <p className="mr-0.5 mt-1 text-right text-sm font-bold text-gray-600">
+            {hasError ? mathStrings.onlyLatex : mathStrings.latexEditorTitle}
+          </p>
+          <button
+            onClick={props.closeMathEditorOverlay}
+            className="mr-0.5 mt-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
```
elbotho: follow up: use and extend `serlo-button-editor-secondary` instead

elbotho (approved):

src/serlo-editor/math/editor.tsx
CodingDive: Added the English translation, the rest is done automagically through crowdin?

PR #2701, opened 2023-08-07
Remove placeholder when not editing

CodingDive:
Fixes https://trello.com/c/Uy0Ur13Q/246-rlp-integration-multimedia-plugin once the edusharing repo updates the git submodule.

As it's only my second PR on the frontend related to edusharing, let me know if there is an easier way to fix it. As it doesn't overwrite the `text-plugin` already, this little change directly in the frontend seems like the easiest fix.


elbotho (approved):
Thanks!
Nice and simple.

kulla:
@CodingDive Thank you :heart:

PR #2702, opened 2023-08-07
editor(math): visual active states

### LINKED ISSUE #261: editor math: no outline in latex mode

elbotho:
(fixes https://github.com/serlo/backlog/issues/261)

- overwrite outline style for visual mode
image

- add visual active state for latex editor
image

@CodingDive sorry for messing with the same files, but I want this in the next deployment. I can solve the conflicts with your PR.

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -200,7 +200,14 @@ export function MathEditor(props: MathEditorProps) {
             />
           </div>
         ) : (
-          <MathRenderer {...props} ref={anchorRef} />
+          <div
+            className={clsx(
+              props.inline ? 'inline-block' : '',
+              'rounded-md bg-editor-primary-200'
+            )}
+          >
+            <MathRenderer {...props} ref={anchorRef} />
+          </div>
         )}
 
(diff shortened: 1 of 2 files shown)


hejtful (approved):
neat

CodingDive (approved):
Works great for me! Great job!

Could you add a tiny bit more padding (especially horizontally)? 

image

elbotho:
> Could you add a tiny bit more padding (especially horizontally)?

We can make that a follow up.
I did not add it so far because it would introduce an additional difference between editing and rendering modes.

elbotho:
@CodingDive Do you know why the formula text is red in your example?

CodingDive:
I don't think I colored it red, let me check 😬

CodingDive:
The backslash causes line breaks and the color to go from black to red :grin: 

image

image

Entkenntnis:
text ist red becaues these are unknown latex commands, if you hover/right click you should see a katex warning

PR #2704, opened 2023-08-07
editor(math): stop latex overlay from closing on double/triple click

### LINKED ISSUE #2700: Double click into the LaTeX editor closes the modal
It's common to double click to highlight some text. However, when doing that within the Math editor, the modal is closed.

double_click_in_LaTeX_closes_menu

elbotho:
for #2700
Tbh. I'm not sure what the root cause is here, but this works well for me.

@CodingDive could you confirm that this solved the problem for you as well?


hejtful (approved):

CodingDive (approved):
Works great for me. I would have never thought of that fix to define `onClick` handler on a div. Super weird bug though and I don't know what caused it either!

PR #2705, opened 2023-08-07
Split math code into multiple files

CodingDive:
Now that we have made some nice bug fixes and visual improvements with the math editor, I just took a few components/render fns and extracted them into their own files/components. 

Also put the DOM element selection into an effect for concurrent mode :tada:

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -58,6 +58,13 @@ export function MathEditor(props: MathEditorProps) {
       targetRef.current = document.querySelector<HTMLDivElement>(
         '.toolbar-controls-target'
       )
+
+      if (!targetRef.current) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'MathEditor: Could not find toolbar-controls-target to create portal'
+        )
+      }
     }
   }, [])
 
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -1,14 +1,14 @@
 import { faCheckCircle, faCircle } from '@fortawesome/free-regular-svg-icons'
-import { faQuestionCircle, faXmark } from '@fortawesome/free-solid-svg-icons'
+import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons'
 import clsx from 'clsx'
 import { useState, createRef, useEffect, useRef } from 'react'
 import { createPortal } from 'react-dom'
 import { useHotkeys } from 'react-hotkeys-hook'
 import { Key } from 'ts-key-enum'
 
+import { MathEditorOverlay } from './math-editor-overlay'
 import { MathHelpModal } from './math-help-modal'
 import { MathRenderer } from './renderer'
-import { MathEditorTextarea } from './textarea'
 import { VisualEditor } from './visual-editor'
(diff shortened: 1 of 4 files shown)


hejtful (approved):
Nice!

src/serlo-editor/math/editor.tsx
```diff
@@ -104,85 +53,30 @@ export function MathEditor(props: MathEditorProps) {
     }
   )
 
-  const isVisualMode = visual && !hasError
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
```
elbotho: useEffect only runs on client so the check is not needed any more right?

src/serlo-editor/math/editor.tsx
```diff
@@ -262,7 +156,7 @@ export function MathEditor(props: MathEditorProps) {
             )}
             {isVisualMode && (
               <button
-                onMouseDown={() => setHelpOpen(true)}
+                onMouseDown={() => void setIsHelpOpen(true)}
```
elbotho, line 157: why do you prefer this?

src/serlo-editor/math/editor.tsx
```diff
@@ -271,43 +165,52 @@ export function MathEditor(props: MathEditorProps) {
...
-            aria-label="Close math formula editor"
-            data-qa="plugin-math-close-formula-editor"
-          >
-            <FaIcon icon={faXmark} />
-          </button>
-        </div>
-        {!isVisualMode && (
-          <MathEditorTextArea {...props} defaultValue={state} />
-        )}
+function MathEditorOverlayPortal({
```
elbotho: Why is this called `…OverlayPortal`? 
Also I think I'd prefer to have this in an own file now.

src/serlo-editor/math/math-help-modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const MathHelpModal = ({
+  isHelpOpen,
+  setIsHelpOpen,
+}: {
+  isHelpOpen: boolean
+  setIsHelpOpen: (isOpen: boolean) => void
+}) => {
+  const mathStrings = useEditorStrings().plugins.text.math
+  return (
+    <Modal
```
elbotho: maybe follow up: using `ModalWithCloseButton` here would probably be a lot simpler.

elbotho (changes requested):

src/serlo-editor/math/editor.tsx
CodingDive: No, that's useLayoutEffect. useEffect works on the server too

src/serlo-editor/math/editor.tsx
CodingDive: I just like to prefix booleans with "is", "has", etc. The `void` I like to symbolize that there is no return value.

src/serlo-editor/math/math-help-modal.tsx
CodingDive: Oh, cool! Will take a look

src/serlo-editor/math/editor.tsx
elbotho: > Effects only run on the client. They don’t run during server rendering.

https://react.dev/reference/react/useEffect

elbotho (approved):
thanks for the improvement!

PR #2716, opened 2023-08-09
fix: eslint typescript errors

hejtful:
Changes:
- Add [eslint-import-resolver-typescript](https://github.com/import-js/eslint-import-resolver-typescript) in order to make the `eslint-plugin-import` work for TS files
- Remove the `import/no-internal-modules` as we didn't stick to it anyway (discussed with @elbotho)

Manual fixes to enforce these changes:
- Change type imports from `import ...` to `import type ...`, or from `import { Type, var } from ...` to `import { type Type, var } from ...`
- Extract utility functions and enums, used by both parent and child components, into a new file
- Replace direct import of `csstype` with import of `CSSProperties` from `react`

Follow-up: The `verbatimModuleSyntax` TS rule could be enabled, so that imports used as types would be auto-imported as `import type ...`, but it brings even more manual changes (would need to import types explicitly as types, even when there are no cyclical imports), so I decided that it's out of scope of this PR.


CodingDive (approved):
Works great for me locally. No errors, neither in vscode, nor when running `yarn lint` through the repl :tada:

elbotho (approved):
thanks @hejtful

PR #2718, opened 2023-08-09
chore(eslint): update packages

elbotho:
I decided to deactivate those rules for now:
```
@typescript-eslint/no-unsafe-enum-comparison
@typescript-eslint/no-explicit-any
@typescript-eslint/no-redundant-type-constituents
```

We can opt-in again at any point if we decide we want those checks.

Rest of the new problems are fixed or ignored in this PR.

PS:
tried https://github.com/un-es/eslint-plugin-i instead of `eslint-plugin-import` but it was not noticeably faster.


hejtful (approved):
Looks good to me!

PR #2719, opened 2023-08-09
fix(plugin-table): pass isInlineChildEditor to inactive cells

### LINKED ISSUE #2709: Plugin toolbar layout breaks within table + image plugin
As usual, the nested plugins are always tricky to get right. The layout of the plugin toolbar breaks when having an image within a table.

image

hejtful:
resolves #2709


elbotho (approved):
nice and simple, thank you!

PR #2721, opened 2023-08-09
chore: remove deprecated eslint rule disables

hejtful:


elbotho (approved):
your call :)

hejtful:
Thanks, I think it's a positive change, as we'll have a smaller eslint config and stricter checks. We can always bring back some of these, if they get annoying.

PR #2723, opened 2023-08-09
fix(serlo-table): restore switch image cell to text cell button

hejtful:
Changes:
- Image is now considered empty if it only has caption text and no actual image
- The button for switching the cell between Image and Text plugins extracted into a component file
- The`updateHack` for showing the button on KeyUp changed, so that it doesn't change focus, but instead just re-renders the `SerloTableEditor` component (result: less code, less dependency imports, less re-renders)

Caveats:
- The "switch image cell to text cell" button is not shown when caption is focused, as in that case, the cell is technically not focused (because cell === Image plugin in this case, and caption's Text plugin is focused).

### DIFF
--- a/src/serlo-editor/plugins/image/index.ts
+++ b/src/serlo-editor/plugins/image/index.ts
@@ -71,8 +71,7 @@ export function createImagePlugin(
       return (
         (!serializedState.src.value || isTempFile(serializedState.src.value)) &&
         (!serializedState.link.defined || !serializedState.link.href.value) &&
-        (!serializedState.alt.defined || !serializedState.alt.value) &&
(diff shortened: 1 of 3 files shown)


elbotho (approved):
thanks. This whole thing is a bit wonky for my taste but your PR fixes the bug nicely.

elbotho:
unrelated: I just saw that the last fix introduced a new bug in the inactive state.
image
Maybe you can quickly fix this in this PR by basically passing the same config in both cell renderers (if that works)?

hejtful:
Thanks, passing empty placeholder config prop to inactive text cell fixed it, as you proposed.

elbotho:
nice, thanks!

Issue #2010, opened 2023-01-09
styled components rely on process.env polyfill


Entkenntnis:
`fallbackNodePolyfills` would be really nice, but unfortunately styled-components is relying on this

related: https://github.com/styled-components/styled-components/pull/3907

-> try again if PR is merged

Entkenntnis:
PR is merged, will be released with v6

Entkenntnis:
postponing this issue because:

- upgrading styled components to v6 could cause issues with editor
- we also need to upgrade https://github.com/Entkenntnis/tmp-serlo-legacy-editor-to-editor
- it's still a experimental feature
- swc compiler has already provided good bundle size decrease

PR #2725, opened 2023-08-10
chore(next): activate fallbackNodePolyfills

elbotho:
Now that we do not use styled-components any more this might work.
(And I think we don't use [tmp-serlo-legacy-editor-to-editor](https://github.com/Entkenntnis/tmp-serlo-legacy-editor-to-editor) and more)

What is fallbackNodePolyfills: https://nextjs.org/blog/next-12-3#other-improvements
Old issue on why we could not use it: https://github.com/serlo/frontend/issues/2010

(edit: oh wow, bundle size difference is quite underwhelming this time around…)


hejtful (approved):
Huh, amazing that you remembered this 😁

Issue #2699, opened 2023-08-06
Page load leads to ory and api recompilation and full reload when running dev server


CodingDive:
Steps to reproduce 
1. Run `yarn dev` 
2. Visit an article like http://localhost:3000/entity/repository/add-revision/277232
3. Make changes to the source code
4. Observe fast refresh/hot module replacement in the browser
5. Trigger a new page load / refresh
6. Ory and parts of the API recompile, triggering multiple page reloads and a compilation cascade.

The terminal looks like this
```
- wait compiling...
- event compiled client and server successfully in 3.3s (3537 modules)

// trigger new page load here
- wait compiling /api/frontend/preview (client and server)...
- wait compiling /api/locale/[lang] (client and server)...
- wait compiling /api/.ory/[...paths] (client and server)...
- event compiled successfully in 1740 ms (1011 modules)
- wait compiling...
- event compiled client and server successfully in 3.1s (3431 modules)
- wait compiling /[...slug] (client and server)...
- event compiled client and server successfully in 3.5s (3538 modules)
```

One can see that after the api recompiles, the slug is next which causes 2 page reloads and everything in total takes more than 10 seconds. 

I have identified this as the main reason why the tests typically fail upon running them against the dev server. Some page reloads or visits cause ory and other parts of the API to recompile.

Entkenntnis:
I wouldn't mind if you can improve the situation here and make it possible to run tests agains dev server (more reliably)

elbotho:
not quite sure but this probably the same issue that slows down dev work in the editor quite a bit (especially when I'm just fine-tuning design stuff.

Any ideas on how to fix this?

CodingDive:
> Any ideas on how to fix this?

I have done a little bit of googling and the best bet right now seems to be precompile the pages that you're going to visit. I have added a primitive fix for this https://github.com/serlo/frontend-e2e-tests/pull/60 here (first commit) but it would require us to hardcode the pages upfront. Therefore even when running a single test with .only, you end up precompiling multiple pages that your test may not even visit :upside_down_face: 
[...]

PR #2727, opened 2023-08-10
fix(dev): move variables into api auth route

elbotho:
for #2699

@CodingDive could you try if that solves the problem for you?

inspiration:
https://stackoverflow.com/a/70682329

### DIFF
--- a/src/pages/api/.ory/[...paths].ts
+++ b/src/pages/api/.ory/[...paths].ts
@@ -1,7 +1,19 @@
-import { COOKIE_DOMAIN } from '@/auth/cookie/cookie-domain'
 import { createApiHandler, config } from '@ory/integrations/next-edge'
 
 export { config }
+
+const COOKIE_DOMAINS = {
+  production: process.env.NEXT_PUBLIC_SERLO_DOMAIN_PRODUCTION,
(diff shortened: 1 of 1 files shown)


CodingDive:
It did not work for me. I still see recompilation when visiting a different page :crying_cat_face: 

It's hard to judge, if it improved anything, but when I first checked out the branch and made some changes in the editor, it felt a little bit snappier until the window refreshed. Are there some other imports that we could try this with? The [...id] pages always seem to recompile, meanwhile the locale / ory and preview parts of the api only sometimes.

```
- event compiled client and server successfully in 4.6s (2561 modules)
- wait compiling /entity/repository/add-revision/[...id]...
- event compiled client and server successfully in 5.2s (3593 modules)
- wait compiling /api/frontend/preview (client and server)...
- wait compiling /api/locale/[lang] (client and server)...
- wait compiling /api/.ory/[...paths] (client and server)...
- event compiled successfully in 2.4s (1109 modules)
```

For the tests, I also couldn't notice a difference. I didn't see ory recompile once (that couild be a win?) but /entity/add-revision[...id], and api/frontend/preview were recompiling a few times.

elbotho:
merge so we can try on staging etc

PR #2728, opened 2023-08-10
editor(slate): use paragraphs and linebreaks

elbotho:
first step of #2593

I suggest we isolate this change first and test it properly and then move to the next feature (spliting text-plugins).

added a hacky frontend part that can be cleaned up after a db migration.

TODO: 
- [x] ~check how saved content like this is displayed in the frontend and see if we need to change something~
- [ ] markdown shortcuts don't work any more if not in first line of paragraph
- [ ] linebreaks work in inline text (e.g. caption)
- [ ] (what exactly should happen when you press enter while text is selected?)
- [ ] testing

> 
> I think this is the way we should have used paragraphs in slate in the first place ;) 
> Not sure if changing this now makes sense, but I think the approach is cleaner.
> 
> Before: 
> - We render the native slate paragraphs as `div` without margin and therefore basically use it as a line-container
> - `shift+enter` triggers slate behaviour:p new Paragraph that gets displayed as new line.
> - `enter` we create a new text-plugin after the current one
> 
> This PR:
> - native slate paragraphs are rendered as `<p class="serlo-p">…` with proper spacing etc.
> - `shift+enter` and `enter`:
>    - if the current line is empty: new paragraph (basically pressing enter twice)
>    - if it is not empty: soft break using `\n`
> - `\n` should get serialized as `<br/>` (needs testing)
> This way people can create Paragraphs (semantically and visually correct) and new lines without knowing about the difference between `enter` and `shift+enter`.


elbotho:
still too many bugs for a first step, trying with #2733 now.

PR #2733, opened 2023-08-12
editor(text-plugin): no new text plugins on enter

elbotho:
third try (after #2728) this time without any changes to the slate format or frontend.
As little change as possible.

### DIFF
--- a/src/serlo-editor/plugins/text/components/text-editor.tsx
+++ b/src/serlo-editor/plugins/text/components/text-editor.tsx
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
         }
 
         // Create a new Slate instance on "enter" key
-        if (isHotkey('enter', event) && !isListActive) {
-          const document = selectDocument(store.getState(), id)
-          if (!document) return
-
-          const mayManipulateSiblings = selectMayManipulateSiblings(
-            store.getState(),
-            id
-          )
-          if (!mayManipulateSiblings) return
-
-          const parent = selectParent(store.getState(), id)
-          if (!parent) return
-
-          event.preventDefault()
-
-          const slicedNodes = sliceNodesAfterSelection(editor)
+        if (isHotkey(['enter', 'shift+enter'], event) && !isListActive) {
+          // prevent two consecutive empty lines
+          // wrapped in setTimeout to let slates build in function run first
           setTimeout(() => {
-            dispatch(
-              insertPluginChildAfter({
-                parent: parent.id,
-                sibling: id,
-                document: {
-                  plugin: document.plugin,
-                  state: slicedNodes || emptyDocumentFactory().value,
-                },
-              })
-            )
+            const { path } = selection.focus
+            const node = Node.get(editor, path)
+            if (!Node.string(node).length) {
+              if (path[0] > 0) {
+                const previousNode = Node.get(editor, [path[0] - 1, 0])
+                if (!Node.string(previousNode).length)
+                  editor.deleteBackward('block')
+              } else editor.deleteBackward('block')
+            }
           })
+
+          // no newlines in headings and start new block as paragraph instead of heading again
+          const fragmentChild = editor.getFragment()[0]
+          const isHeading =
+            Object.hasOwn(fragmentChild, 'type') && fragmentChild.type === 'h'
+
+          if (isHeading) {
+            event.preventDefault()
+            Transforms.insertNodes(editor, emptyDocumentFactory().value)
+            return
+          }
         }
 
         // Merge with previous Slate instance on "backspace" key,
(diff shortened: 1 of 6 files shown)


src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
         }
 
         // Create a new Slate instance on "enter" key
```
hejtful: This comment looks outdated, maybe update/remove it?

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
...
-
-          const parent = selectParent(store.getState(), id)
-          if (!parent) return
-
-          event.preventDefault()
-
-          const slicedNodes = sliceNodesAfterSelection(editor)
+        if (isHotkey(['enter', 'shift+enter'], event) && !isListActive) {
+          // prevent two consecutive empty lines
+          // wrapped in setTimeout to let slates build in function run first
```
hejtful:
```suggestion
          // Prevent two consecutive empty lines.
          // Wrapped in `setTimeout` to let Slate's built-in function to run first
```

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
...
-            )
+            const { path } = selection.focus
+            const node = Node.get(editor, path)
+            if (!Node.string(node).length) {
+              if (path[0] > 0) {
+                const previousNode = Node.get(editor, [path[0] - 1, 0])
+                if (!Node.string(previousNode).length)
+                  editor.deleteBackward('block')
+              } else editor.deleteBackward('block')
+            }
```
hejtful:
```suggestion
            // If not an empty line, do nothing
            if (Node.string(node).length) return

            // If node does not exist in editor, delete previous block
            const nodePositionInEditor = path[0]
            if (nodePositionInEditor <= 0) {
              editor.deleteBackward('block')
              return
            }

            // If previous node has no text, delete previous block
            const previousNode = Node.get(editor, [nodePositionInEditor - 1, 0])
            if (!Node.string(previousNode).length) {
              editor.deleteBackward('block')
            }
```

1. As this code is inside `setTimeout`, early exits are possible, so using them here (more performant, lower function complexity, easier to explain with comments).
2. Assigning `path[0]` to a constant. I might have given it a wrong constant name, but that would only hint to me that it needs a name, to help understand what it is.
3. Same with comments, if my comments are wrong, then it's an indicator that comments are needed 😄

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
...
           })
+
+          // no newlines in headings and start new block as paragraph instead of heading again
```
hejtful:
```suggestion
          // Prevent newlines in headings. Instead, add a new paragraph as the next block.
```

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -182,34 +182,31 @@ export function TextEditor(props: TextEditorProps) {
...
+          const fragmentChild = editor.getFragment()[0]
+          const isHeading =
+            Object.hasOwn(fragmentChild, 'type') && fragmentChild.type === 'h'
+
+          if (isHeading) {
+            event.preventDefault()
+            Transforms.insertNodes(editor, emptyDocumentFactory().value)
+            return
```
hejtful: Why is `return` needed here?

hejtful (commented):
Preventing two empty lines only works for new lines (when pressing enter twice). If there are already existing empty lines before the current selection, adding a new line is still possible.

elbotho:
> Preventing two empty lines only works for new lines (when pressing enter twice). If there are already existing empty lines before the current selection, adding a new line is still possible.

yeah… do you have an easy idea so solve that? Otherwise I thought it's okay for now (not the only way to do forbidden things in the editor) and create a backlog issue about adding a nice slate normalizer or something like that.

src/serlo-editor/plugins/text/components/text-editor.tsx
elbotho: it's not :)

hejtful:
Sounds good to me, it's not a critical bug and shouldn't hold us back for now.

hejtful (approved):

PR #2735, opened 2023-08-12
editor(text-plugin): show suggestion on every line

elbotho:


src/serlo-editor/plugins/text/utils/insert-plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface insertPluginArgs {
+  pluginType: string
+  editor: SlateEditor
+  store: RootStore
+  id: string
+  dispatch: ThunkDispatch<unknown, unknown, Action<unknown>>
+  state?: unknown
+}
+
+export function insertPlugin({
```
elbotho, line 23: maybe "split" should be part of the name since it splits the text plugin and then adds a new editor plugin, but I'm not sure if that is needed.

hejtful (approved):

PR #2736, opened 2023-08-12
editor(text-plugin): add button to placeholder to open suggestion menu

elbotho:


src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -54,6 +54,11 @@ export function TextEditor(props: TextEditorProps) {
     [createTextEditor]
   )
 
+  const mayManipulateSiblings = useMemo(
+    () => selectMayManipulateSiblings(store.getState(), id),
+    [id]
+  )
```
hejtful, line 60: This will not update when the state changes, only when the `id` changes (so, I guess, never).

`selectMayManipulateSiblings` check if the plugin is a child of Rows plugin. Can that change? Is it possible to move a plugin from non-Rows to Rows, or from Rows to non-Rows?

elbotho: Thanks for the question, was not 100% sure here either.
As I understand it, drag&dropping plugins is a feature that only works for children of rows plugins so I'm pretty sure this can't change right now.

PR #2742, opened 2023-08-14
refactor(text-plugin): move callbacks into hooks

elbotho:
~actually pretty stupid to do this in parallel to https://github.com/serlo/frontend/pull/2733 now that I think of it 😅~

- this also reintroduces the placeholder add button that got lost in some merge
- one follow up issue here already: https://github.com/serlo/frontend/issues/2743


LarsTheGlidingSquirrel (approved):
Code looks fine and I did not notice any bugs compared to staging.

PR #2744, opened 2023-08-15
Fix bug that visual/latex mode is sometimes not shown in equations

CodingDive:
Any idea why the .toolbar-controls-target is not available on first render? 

I don't like my solution of reaching for the big guns (MutationObserver), and I can refactor this within the next few days to have some proper parent-child communication, or even easier, we don't do any document.querySelector selection and instead, just attach a ref to .toolbar-controls-target and put it into React context, that children can just grab from there.
There are other solutions like adding a small timeout.

Will also add some tests and have already prepared for it, by adding some data-qa attributes.

As it's a holiday today, I'd appreciate if we can get this in quickly, so that Peter and the other authors can do their job without being annoyed.

Bug can be seen here https://www.loom.com/share/10aef34a4d2b4a70a933d671322d730d

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -58,12 +58,31 @@ export function MathEditor(props: MathEditorProps) {
       '.toolbar-controls-target'
     )
 
-    if (!targetRef.current) {
-      // eslint-disable-next-line no-console
-      console.warn(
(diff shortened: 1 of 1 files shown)


elbotho:
alternative solution in https://github.com/serlo/frontend/pull/2745

PR #2745, opened 2023-08-15
editor(equations): quickfix for math toolbar bug

elbotho:
> don't like my solution of reaching for the big guns (MutationObserver)

@CodingDive: Here is a quick solution that works fine for me.
If It works for you as well I'd rather merge the hack with less complexity.
If you want to ship this quickly we should do a hotfix deploy because I'd like to test the new features in staging a bit more.

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -161,8 +161,10 @@ export function MathEditor(props: MathEditorProps) {
   }
 
   function renderControlsPortal(children: JSX.Element) {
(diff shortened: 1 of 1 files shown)


CodingDive:
Sure! That works :tada:

CodingDive (approved):
Thanks! I think this is the way it was before I refactored it into the `useEffect` with unintended consequences :grin: Let's talk about how we can make this nicer down the line, ideally getting rid of the querySelector especially during render.

PR #2748, opened 2023-08-15
refactor(plugin-text): merge code for pasting image, video and geogebra

### LINKED ISSUE #2743: editor(text-plugin): pasting image url works, but leaves url string
at least in https://github.com/serlo/frontend/pull/2742
image

hejtful:


src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
```diff
@@ -27,93 +26,45 @@ export const useEditablePasteHandler = (args: UseEditablePasteHandlerArgs) => {
...
-      )
-      if (!mayManipulateSiblings) return
+      // Exit if unable to select document data or if not allowed to manipulate siblings
+      const document = selectDocument(storeState, id)
+      const mayManipulateSiblings = selectMayManipulateSiblings(storeState, id)
+      if (!document || !mayManipulateSiblings) return
 
+      // Exit if no files or text in clipboard data
       const files = Array.from(event.clipboardData.files)
       const text = event.clipboardData.getData('text')
```
elbotho: premature optimization: should we move this check to the top to not access the store and selectors unnecessarily?

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
```diff
@@ -39,9 +38,8 @@ export const useEditablePasteHandler = (args: UseEditablePasteHandlerArgs) => {
       if (!files.length && !text) return
 
       // Iterate through all plugins and try to process clipboard data
-      const plugins = editorPlugins.getAllWithData()
       let media
-      for (const { plugin, type } of plugins) {
+      for (const { plugin, type } of editorPlugins.getAllWithData()) {
```
elbotho, line 42: It's not more readable, it's probably slower but hey it's functional :D 

suggestion:
``` ts
const media = editorPlugins
  .getAllWithData()
  .map(({ plugin, type }) => {
    const state = (plugin.onFiles?.(files) ?? plugin.onText?.(text))
      ?.state as unknown
    return state ? { state, pluginType: type } : undefined
  })
  .find((object) => object !== undefined)
```

notes:
- `reduce` version is uglier
- it's a Jonas legacy that we use functional version wherever possible in the codebase
- I'm also fine with keeping it at it is :)

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
```diff
@@ -27,93 +26,41 @@ export const useEditablePasteHandler = (args: UseEditablePasteHandlerArgs) => {
...
 
-      if (text) {
-        // Handle pasted video URLs
-        const videoPluginState = editorPlugins
-          .getByType(EditorPluginType.Video)
-          ?.onText?.(text)
-        if (videoPluginState !== undefined) {
-          event.preventDefault()
+      // Exit if no media was processed from clipboard data
+      if (media?.state === undefined) return
```
elbotho: `if (!media) return` should be enough here.

elbotho (approved):

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
hejtful: I wouldn't call this premature at all, it's a great catch!

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
hejtful: I prefer keeping it as it is, because it will exit as soon as it processes the clipboard data, while the suggestion will map all of them, and then only exit once it `find`s.

>it's a Jonas legacy that we use functional version wherever possible in the codebase

I love functional when it makes sense, but using functional, only for the sake of using functional, is one of my least favorite things about functional.

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
hejtful: I have a vague memory of trying that first, and then coming across some bug because of it. But definitely worth trying again, for brevity.

src/serlo-editor/plugins/text/hooks/use-editable-paste-handler.tsx
elbotho: that's a pretty functional view of functional I guess ;)

PR #2749, opened 2023-08-15
Add data-qa attributes

CodingDive:
Sorry for the inaccurate branch name. Just wanted to get the data-qa attributes in for some follow up/regression tests that I'll add once I have the time.

### DIFF
--- a/src/serlo-editor/math/editor.tsx
+++ b/src/serlo-editor/math/editor.tsx
@@ -125,15 +125,8 @@ export function MathEditor(props: MathEditorProps) {
                 props.onEditorChange(e.target.value === 'visual')
               }}
             >
-              <option
-                value="visual"
-                data-qa="editor-plugin-toolbar-math-visual"
-              >
-                {mathStrings.visual}
-              </option>
-              <option value="latex" data-qa="editor-plugin-toolbar-math-latex">
-                {mathStrings.latex}
-              </option>
+              <option value="visual">{mathStrings.visual}</option>
+              <option value="latex">{mathStrings.latex}</option>
             </select>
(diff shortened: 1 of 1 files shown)


src/serlo-editor/math/editor.tsx
```diff
@@ -119,13 +119,21 @@ export function MathEditor(props: MathEditorProps) {
                 hover:bg-editor-primary-200 focus:bg-editor-primary-200 focus:outline-none
                 `}
               value={isVisualMode ? 'visual' : 'latex'}
+              data-qa="editor-plugin-toolbar-math-visual-latex-switch"
```
hejtful: Maybe it can be a bit shorter, like:
```suggestion
              data-qa="plugin-toolbar-math-type-switch"
```

.

\+ I'd use this one + the text of the options, to access the options. Three `data-qa` attributes for a select and its two static options seems like an overkill to me.

Both are just suggestions, of course.

src/serlo-editor/math/editor.tsx
```diff
@@ -119,13 +119,21 @@ export function MathEditor(props: MathEditorProps) {
...
               onChange={(e) => {
                 if (hasError) setHasError(false)
                 props.onEditorChange(e.target.value === 'visual')
               }}
             >
-              <option value="visual">{mathStrings.visual}</option>
-              <option value="latex">{mathStrings.latex}</option>
+              <option
+                value="visual"
+                data-qa="editor-plugin-toolbar-math-visual"
```
elbotho: tiny thing: do we really need attributes for both switch and the options?

elbotho (approved):

PR #2750, opened 2023-08-18
fix(text-plugin): Improve toolbar icons horizontal alignment

LarsTheGlidingSquirrel:
Quick fix for making the toolbar icons more aligned on a horizontal line.


CodingDive (approved):
Thank you! Already looks much better 
image


I'm wondering if the link can be put a little bit up top too, it seems to have more space to the bottom of the toolbar than to the top.

hejtful:
I would always check with Botho or Gregor if they want these changes in the first place 😄 

Edit: It was discussed, ignore my comment please.

PR #2755, opened 2023-08-21
Audio plugin prototype

CodingDive:
For the simple prototype, there are two ways to get audio. You can either record directly in the browser (right now this uses Base64 encoding which is obviously quite big, but I haven't ran into horrible performance issues for short audio files. Tested up until 1 minute of recording), or embed https://vocaroo.com/ URLs. 

There are some leftover tasks, but if the code looks okay, I could address them in follow-up PRs, or tomorrow until you have finished the first review. Mostly disappointed with the UI which looks horrible right now, so I'd need to sit with Gregor or Botho if we think it's worth it to improve upon. 

- [ ] Design/UI, make it look like other Serlo components
- [ ] Change Icon in plugin selection menu
- [ ] Simplify conditional logic by only having a single prop in store (`srcOrBase64`). By having two props, there is a lot of duplicated code right now. We can simply check if it's a url or base64 on the fly when needed (when rendering the iframe or AudioRecorder, and not in 10 different places)
- [ ] Check how I can support audio next to exercises
- [ ] Consider disabling recording of podcasts (>5min audio files) because that would probably lead to performance issues indeed


src/frontend-node-types.ts
```diff
@@ -383,6 +385,14 @@ export type FrontendVideoNode = EditorVideoPlugin & {
   pluginId?: string
 }
 
+export type FrontendAudioNode = EditorAudioPlugin & {
+  type: FrontendNodeType.Audio
+  children?: undefined
+  pluginId?: string
+  src: string
+  base64AudioRecording: string
```
kulla, line 393: You store the audio base64 encoded? -> I would rather want to upload it to our media store instead of storing it into our mysql...

kulla (changes requested):

src/frontend-node-types.ts
CodingDive: I generally agree, but for a prototype? As I wrote, I didn't experience performance issues for small audio files. Blob support is always appreciated though :grin:

Entkenntnis: Would vocaroo support be enough for a prototype?

We have to remember that storing data in a revision means that the data is duplicated on every edit of the content, is also puts more pressure on the redis cache (which is already struggling ..)

elbotho: @CodingDive it's a neat feature to record audio directly on serlo but I'll agree with @Entkenntnis that is the part that is out of scope of a prototype like this.

elbotho:
example editor link: https://frontend-git-feat-audio-plugin-prototype-serlo.vercel.app/entity/repository/add-revision/277232
example frontend link: https://frontend-git-feat-audio-plugin-prototype-serlo.vercel.app/277232
example voca.ro link: https://voca.ro/1gwMMlwAqhRM

src/data/en/index.ts
```diff
@@ -173,6 +173,7 @@ export const instanceData = {
         proof: 'Proof',
       },
       loadingVideoFailed: 'Something went wrong',
+      loadingAudioFailed: 'Something went wrong',
```
elbotho, line 176: I don't mind if you reuse or generalise here.

src/data/en/index.ts
```diff
@@ -887,6 +888,17 @@ export const loggedInData = {
           url: 'URL',
           seoTitle: 'Title for search engines',
         },
+        audio: {
```
elbotho, line 891: some of those are unused (e.g. `audioDescription`)

elbotho (changes requested):
Code overall looks good.
I was not involved in the conception here so forgive me if those questions are not relevant:
- Did you think about [vocaroo privacy](https://datenschutz-schule.info/tag/vocaroo/#:~:text=Speicherdauer,l%C3%A4nger%20nicht%20mehr%20aufgerufen%20werden.)? As things are now we might need an additional consent click before loading :/
- vocaroo will delete audio files after some time or if they are not used from time to time, is that an issue for us?
- Otherwise using just vocaroo (instead of recording on serlo.org) would be a lot simpler and suitable for an POC (see above)
- Could we access the audio file directly or do we have options to style the vocaroo player?

src/serlo-editor/plugins/audio/editor.tsx
```diff
@@ -0,0 +1,84 @@
...
+              onClick={() => setShowRecorder(false)}
+              className="serlo-button absolute right-1 top-1 text-sm font-bold text-gray-600 hover:bg-gray-200 hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-gray-700"
+              aria-label="Close audio recorder"
+              data-qa="plugin-audio-close-recorder"
+            >
+              <FaIcon icon={faXmark} />
+            </button>
+          </div>
+        ) : (
+          <>
```
elbotho, line 60: if there is an audio file or url already I would these buttons expect to hide.

Entkenntnis:
I was part of the conception. We would need a frontend proxy like https://github.com/serlo/frontend/blob/staging/src/pages/api/frontend/lumi/embed/%5Bid%5D.ts to disable ads and hide visitors ip.

another way would be trying to access https://media1.vocaroo.com/mp3/1gwMMlwAqhRM directly (it's protected a bit, but no tokens etc seem to be involved)

CodingDive:
Experimented with the API here https://github.com/serlo/api.serlo.org/pull/1002

PR #2760, opened 2023-08-22
fix(plugin-text): show static placeholder instead of decorate when config.noLinebreaks

### LINKED ISSUE #2756: optional title placeholder only visible after click
If I create a new box and choose "Merke", the optional title is not visible:

grafik

I can see a text cursor, if I click the plugin, I can see the placeholder:

grafik

Is this a bug or just a glitch? It's a regression because one e2e test fails (so it worked before)

hejtful:
Fixes: https://github.com/serlo/frontend/issues/2756

Test PR: https://github.com/serlo/frontend-e2e-tests/pull/66

### DIFF
--- a/src/serlo-editor/plugins/box/editor.tsx
+++ b/src/serlo-editor/plugins/box/editor.tsx
@@ -54,7 +54,7 @@ export function BoxEditor(props: BoxProps) {
           )}
         >
           <b className="block pb-4">{editorStrings.plugins.box.type}</b>
-          <ul className="unstyled-list pb-8">{renderSettingsList()}</ul>
+          <ul className="unstyled-list pb-8">{renderSettingsListItems()}</ul>
         </figure>
       </>
     )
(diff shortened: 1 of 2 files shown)


Entkenntnis (approved):
tested it and it works now

src/serlo-editor/plugins/box/editor.tsx
```diff
@@ -54,7 +54,7 @@ export function BoxEditor(props: BoxProps) {
           )}
         >
           <b className="block pb-4">{editorStrings.plugins.box.type}</b>
-          <ul className="unstyled-list pb-8">{renderSettingsLis()}</ul>
+          <ul className="unstyled-list pb-8">{renderSettingsList()}</ul>
```
elbotho: hmm, it's not rendering the list, but the list items. that's why it was `Lis`. If we change this to be more unambiguous I'd go with `renderSettingsListItems` or move the `<ul>` to the render function as well :D

elbotho (approved):
👍

PR #2766, opened 2023-08-22
refactor(editor): add parent buttons to box+spoiler, use domFocus for multimedia

elbotho:


src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -28,9 +28,9 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
   const document = useAppSelector((state) => selectDocument(state, id))
 
   const focused = useAppSelector((state) => selectIsFocused(state, id))
-  const [domFocus, setDomFocus] = useState<'focus' | 'focusWithin' | false>(
-    focused ? 'focusWithin' : false
-  )
+  const [domFocus, setDomFocus] = useState<
+    'focus' | 'focusWithin' | 'focusWithinInline' | false
+  >(focused ? 'focusWithin' : false)
```
hejtful: How about making this type an enum, and replacing `false` with a string like `'false'` or `'notFocused'`?

```typescript
enum DomFocus {
  focus = 'focus',
  focusWithin = 'focusWithin',
  focusWithinInline = 'focusWithinInline',
  notFocused = 'notFocused',
}
```

src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -146,12 +151,16 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
...
+          isInlineChildEditor ? true : undefined
+        }
       >
         <plugin.Component
           containerRef={containerRef}
           id={id}
           editable
           domFocusWithin={domFocus !== false}
+          domFocusWithinInline={domFocus === 'focusWithinInline'}
           domFocus={domFocus === 'focus'}
```
hejtful: And then here passing only one prop, `domFocus={domFocus}`?

src/serlo-editor/core/sub-document/renderer.tsx
```diff
@@ -35,6 +35,7 @@ export function SubDocumentRenderer({ id, pluginProps }: SubDocumentProps) {
       editable={false}
       domFocus={false}
       domFocusWithin={false}
+      domFocusWithinInline={false}
```
hejtful: It would reduce the amount of props that need to be drilled.

src/serlo-editor/plugins/box/toolbar.tsx
```diff
@@ -8,11 +10,19 @@ import { PluginToolbar } from '@/serlo-editor/editor-ui/plugin-toolbar'
...
+  state,
+  domFocusWithin,
+  domFocusWithinInline,
+  domFocus,
+}: BoxProps) => {
   const boxStrings = useEditorStrings().plugins.box
   const { strings } = useInstanceData()
 
-  return (
+  if (!domFocusWithin) return null
```
hejtful, line 22:
```suggestion
  if (domFocus === DomFocus.notFocused) return null
```

src/serlo-editor/plugins/box/toolbar.tsx
```diff
@@ -8,11 +10,19 @@ import { PluginToolbar } from '@/serlo-editor/editor-ui/plugin-toolbar'
...
+
+  return domFocus || domFocusWithinInline ? (
```
hejtful:
```suggestion
  return[DomFocus.focus, DomFocus.focusWithinInline].includes(domFocus) ? (
```

hejtful (approved):
Works like a charm!

Commented with some proposals for increasing type strictness and reducing prop drilling. But please feel free to ignore those or leave them for later.

src/serlo-editor/plugins/box/toolbar.tsx
```diff
@@ -43,5 +53,16 @@ export const BoxToolbar = ({ id, state }: BoxProps) => {
...
+    <button
+      className={clsx(
+        tw`
+            absolute -top-6 right-14 z-50 block h-6 rounded-t-md bg-gray-100
+            px-2 pt-0.5 text-sm font-bold hover:bg-editor-primary-100
+          `
+      )}
+    >
+      {boxStrings.title}
+    </button>
```
hejtful: This button no longer has a `data-qa` attribute (no idea if you did that on purpose or not), so now there's nothing else to indicate what the purpose of it is. I would propose adding at least a comment

src/serlo-editor/core/sub-document/editor.tsx
elbotho: I thought about this as well.
I'm still not sure.
The current approach increases complexity here but feels a lot cleaner with most consumers (most use just one variable). 
So instead of a nice `domFocus ? …` we would have `domDocus === DomFocus.focus ? : …` for example.
But I agree that three focus props is too much somehow. Especially since `focusWithinInline` is only used once so far.

elbotho: I will merge this but we can continue the discussion here or on slack.
What do you think about having two props?
- `focusWithin: boolean` and
- `focusState: FocusState (enum you proposed)`

Most plugins only need the simple boolean and if you need a more complex feature you can use `focusState`.

hejtful: Sounds good to me.

elbotho: see https://github.com/serlo/frontend/pull/2766/commits/9869771faf04a5a1f631d5a4f3c1c29f86682823

PR #2768, opened 2023-08-23
Support audio plugin prototype with url only

CodingDive:
Replaces https://github.com/serlo/frontend/pull/2755

- [x] Change icon (wait on Yannick)
- [x] Show toolbar (focus plugin) when iframe is clicked (deferred, see https://github.com/serlo/frontend/pull/2768#issuecomment-1690262932)
- [x] Check if vocaroo shows ads within iframe and if we need to do something like https://github.com/serlo/frontend/blob/staging/src/pages/api/frontend/lumi/embed/%5Bid%5D.ts 

Refreshed in Incognito (without ad block) 10 times and didn't see any ads when embedded as iframe


src/serlo-editor/plugins/audio/renderer.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <iframe width="300" height="60" src={vocarooUrl} allow="autoplay" />
+        </PrivacyWrapper>
+      </div>
+    )
+  }
+
+  return <p>Unknown audio type</p>
+}
+
+export function parseAudioUrl(src: string): [string, AudioType | undefined] {
```
elbotho, line 53: personally I'd prefer to remove the AudioType everywhere and reintroduce it if we actually need it. (so return type `string | undefined`
but I guess it's okay like this as well for a prototype.

src/helper/icon-by-entity-type.ts
```diff
@@ -47,6 +48,7 @@ export function getIconByTypename(
 
 export const entityIconMapping = {
   applet: faCubes,
+  audio: faPlay,
```
elbotho, line 51: plugins are not entities, so this is not the place.
also it seems unused, right?

src/data/en/index.ts
```diff
@@ -888,6 +890,13 @@ export const loggedInData = {
           url: 'URL',
           seoTitle: 'Title for search engines',
         },
+        audio: {
+          title: 'Audio',
+          description: 'Link to audio files or record your own',
+          audioUrl: 'Audio URL',
+          url: 'URL',
+          seoTitle: 'Title for search engines'
```
elbotho: usused?

src/data/en/index.ts
```diff
@@ -888,6 +890,13 @@ export const loggedInData = {
           url: 'URL',
           seoTitle: 'Title for search engines',
         },
+        audio: {
+          title: 'Audio',
+          description: 'Link to audio files or record your own',
+          audioUrl: 'Audio URL',
+          url: 'URL',
```
elbotho: unused?

elbotho (changes requested):
small stuff.
the ads would be annoying though, curious what you'll find.

src/helper/icon-by-entity-type.ts
CodingDive: Hmm, I don't think I understand the difference between plugin and entity yet 

It's used by the privacy-wrapper

src/data/en/index.ts
CodingDive: Can you check again? the seoTitle is the only one I don't know about (just copied from other plugin i18n strings)

elbotho: where do you use `url`?

src/data/en/index.ts
elbotho: you don't need it for this plugin.

src/data/en/index.ts
elbotho: Okay, rest looks good. Commented on seoTitle below.

src/serlo-editor-integration/create-plugins.tsx
```diff
@@ -134,12 +138,17 @@ export function createPlugins({
...
-    },
+    ...(isProduction
+      ? []
+      : [
+          {
+            type: EditorPluginType.Audio,
+            plugin: audioPlugin,
+            visibleInSuggestions: true,
+            icon: <IconAudio />,
+          } as PluginWithData,
```
elbotho, line 149: `PluginsWithData[number]` if you are lazy next time ;)

elbotho: nice, I guess that's even better than just hiding it from the suggestions.

src/serlo-editor/plugins/audio/editor.tsx
```diff
@@ -5,12 +5,15 @@ import type { AudioProps } from '.'
...
 import { FaIcon } from '@/components/fa-icon'
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
 
 export const AudioEditor = (props: AudioProps) => {
   const { focused, state } = props
   const [showSettingsModal, setShowSettingsModal] = useState(false)
   const [src, type] = parseAudioUrl(state.src.value)
   const couldBeValid = type !== undefined
+  // i18n string
+  const { audioUrl } = useEditorStrings().plugins.audio
```
elbotho: works, but I suggest to go with `const audioStrings = useEditorStrings().plugins.audio` and `audioStrings.audioUrl` to avoid confusions. Then you don't need a comment.

src/serlo-editor/plugins/audio/editor.tsx
```diff
@@ -27,21 +30,35 @@ export const AudioEditor = (props: AudioProps) => {
...
+      <button
+        className="flex cursor-pointer flex-col items-center border-none bg-transparent p-0"
+        onClick={openSettings}
+      >
+        <FaIcon icon={faLink} className="text-2xl" />
+        <span className="mt-2">{children}</span>
+      </button>
+    </div>
+  )
+}
```
elbotho: Sorry, I know this is a cleaner pattern than the `render…` stuff, but if you abstract it as a component I'd prefer if you would put it into an own file.

src/serlo-editor/plugins/audio/editor.tsx
elbotho: or `const audioUrlTitle = useEditorStrings().plugins.audio.audioUrl`

src/serlo-editor/plugins/audio/toolbar.tsx
```diff
@@ -22,6 +28,25 @@ export const AudioToolbar = ({
...
+    (event) => {
+      event.preventDefault()
+      setShowSettingsModal(false)
+    },
+    {
+      enableOnContentEditable: true,
+      enableOnFormTags: true,
+      scopes: ['global'],
+    }
+  )
```
elbotho, line 42: why is this needed?

elbotho:
> Show toolbar (focus plugin) when iframe is clicked

Imho you can ignore this for now. If we manage to get the new focus approach running this would probably just work.

src/serlo-editor/plugins/audio/toolbar.tsx
CodingDive: The modal doesn't close with ESC

src/serlo-editor-integration/create-plugins.tsx
CodingDive: > PluginsWithData[number] if you are lazy next time ;)

Hehe clever!

src/serlo-editor/plugins/audio/toolbar.tsx
elbotho: hmm, works fine for me with e.g. video plugin.
That would be a bug of our Modal Component.

src/helper/icon-by-entity-type.ts
elbotho: I see. So the usage in the privacy wrapper is not entirely correct but worked since videos and geogebra are also entities. That's also why it's confusing I guess ^^

elbotho: Thanks for checking.

elbotho (approved):

PR #2770, opened 2023-08-23
fix(editor): box and multimedia focus parent button

hejtful:


src/serlo-editor/plugins/box/toolbar.tsx
```diff
@@ -20,6 +21,19 @@ export const BoxToolbar = ({
   const boxStrings = useEditorStrings().plugins.box
   const { strings } = useInstanceData()
 
+  const handleFocusParentButtonClick = useCallback((event: MouseEvent) => {
+    const target = event.target as HTMLButtonElement
```
elbotho: I prefer `MouseEvent<HTMLButtonElement>` in the line above now.

elbotho (commented):
Thanks for solving this!
If it works it works. 
To be honest it feels like a failure of my approach if something like this is necessary.
It's probably way cleaner if we refactor it into a hook like you suggested.

PR #2778, opened 2023-08-24
refactor(focus): remove focus from store (wip)

elbotho:
@hejtful as expected there are some bugs, but it's not as bad as I feared.
If you like you can take this over, I'll be back in the afternoon. 

todos:
- write new functions for `focusNext` and `focusPrevious` (could use the "focus"-Tree or just the dom to do so, performance should not really matter here)
- bug: first focus resets slate-curor to first position instead of where the user clicked
- bug: autofocus does not work as expected any more
- more refactoring (remove unused store, move or rename "focus"-Tree stuff, focus hook, …?)

### DIFF
--- a/src/serlo-editor/core/sub-document/editor.tsx
+++ b/src/serlo-editor/core/sub-document/editor.tsx
@@ -13,9 +13,7 @@ import type { SubDocumentProps } from '.'
 import { useEnableEditorHotkeys } from './use-enable-editor-hotkeys'
 import {
   runChangeDocumentSaga,
-  focus,
   selectDocument,
-  selectIsFocused,
   useAppSelector,
   useAppDispatch,
   insertPluginChildAfter,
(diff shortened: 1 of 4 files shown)


src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -59,20 +54,10 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
 
   useEffect(() => {
     if (domFocusState !== 'focus') return
-    setTimeout(() => autofocusRef.current?.focus())
+    // TODO: does not work any more (with or without timeout)
+    autofocusRef.current?.focus()
```
hejtful, line 58: It seems like this never worked 😄 Try deleting the whole effect on staging, autofocus still works.

This is [the line](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/store/focus/slice.ts#L34) that actually sets the focus. On page load, the plugins are added sequentially, but in reverse order. Each one will be focused in the store, but the first plugin (text plugin of article introduction) will be added last, and therefore focused last. And then this code in `SubDocumentEditor` makes sure store focus and DOM focus are aligned:
```typescript
  const [domFocusState, setDomFocus] = useState<DomFocus>(
    focused ? DomFocus.focusWithin : DomFocus.notFocused
  )
```

It's a hack, but I thought why not try it directly with DOM focus. Doesn't work, breaks the editor completely. So I'll explore the possibility of changing the approach. The first idea that I have is to manually focus the article introduction text plugin from the `ArticleEditor` component.

hejtful: Pull request with the implementation of the mentioned idea: https://github.com/serlo/frontend/pull/2780

PR #2779, opened 2023-08-24
don't let injection confuse dev server

### LINKED ISSUE #2699: Page load leads to ory and api recompilation and full reload when running dev server

Entkenntnis:
fixes #2699 

This one's on me, the way injections load data is great for production and caching, but confusing for the dev server and is the root cause for the second refresh.

Loading the editor on dev should now not lead to page refreshes.

### DIFF
--- a/src/serlo-editor/plugins/injection/renderer.tsx
+++ b/src/serlo-editor/plugins/injection/renderer.tsx
@@ -33,13 +33,17 @@ export function InjectionRenderer({
     }
 
     const encodedHref = encodeURI(href.startsWith('/') ? href : `/${href}`)
-    const fetchUrl = `${window.location.protocol}//${window.location.host}/_next/data/${buildId}/${lang}${encodedHref}.json`
+    const fetchUrl =
+      window.location.hostname === 'localhost'
+        ? `${window.location.protocol}//${window.location.host}/api/frontend/${encodedHref}`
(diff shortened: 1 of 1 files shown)


elbotho (approved):
uh, excited for this one!
No idea how you figured it out, thank you!
Seems to work nicely, let's test it in actual editor PRs.

PR #2780, opened 2023-08-24
refactor(editor): remove selectIsFocused selector

hejtful:
This PR is not to `remove-store-focus` branch, because that branch introduces multiple bugs, and I preferred to introduce and fix bugs one by one.

In `ArticleEditor`, I thought about adding a class instead of a ref, but a class would stay local to the `ArticleEditor`, while a ref can be moved into the focus-dedicated hook and therefore connected to the focus feature.

**Issues:**
- The "table click outside blur" bug does not appear in this branch, but two clicks are needed to focus a table cell. Looking into it.
- Doesn't work for non-article entities, like e.g. [exercise](http://localhost:3000/entity/create/Exercise/60730).
- When creating a new plugin it does not get focused (at all, but also not where it's supposed to get focus via autofocusref).

### DIFF
--- a/src/serlo-editor/plugins/article/editor.tsx
+++ b/src/serlo-editor/plugins/article/editor.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 
 import type { ArticleProps } from '.'
 import { ArticleAddModal } from './add-modal/article-add-modal'
@@ -22,8 +22,18 @@ export function ArticleEditor({ editable, state }: ArticleProps) {
 
   const modalStrings = useEditorStrings().templatePlugins.article.addModal
 
+  // Focus the article introduction explanation on page load
+  // TODO: Move this into a focus-dedicated hook
+  const ref = useRef<HTMLDivElement | null>(null)
+  useEffect(() => {
+    const explanationTextPlugin = ref.current?.querySelector<HTMLDivElement>(
+      '.explanation-wrapper > .plugin-wrapper-container'
+    )
+    explanationTextPlugin?.focus()
+  }, [])
+
   return (
-    <>
+    <div ref={ref}>
       <ArticleRenderer
         introduction={introduction.render()}
         content={content.render()}
@@ -81,7 +91,7 @@ export function ArticleEditor({ editable, state }: ArticleProps) {
           setModalOpen={setModalOpen}
         />
       )}
-    </>
+    </div>
   )
 
   function renderButton(text: string, noIcon?: boolean) {
--- a/src/serlo-editor/core/sub-document/editor.tsx
+++ b/src/serlo-editor/core/sub-document/editor.tsx
@@ -1,21 +1,12 @@
 import clsx from 'clsx'
 import * as R from 'ramda'
-import {
-  useState,
-  useRef,
-  useEffect,
-  useMemo,
-  useCallback,
-  FocusEvent,
-} from 'react'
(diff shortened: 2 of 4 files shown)


src/serlo-editor/plugins/article/editor.tsx
```diff
@@ -22,8 +22,18 @@ export function ArticleEditor({ editable, state }: ArticleProps) {
 
   const modalStrings = useEditorStrings().templatePlugins.article.addModal
 
+  // Focus the article introduction explanation on page load
+  // TODO: Move this into a focus-dedicated hook
+  const ref = useRef<HTMLDivElement | null>(null)
```
elbotho, line 27: slightly prefer something like `wrapperRef`.

src/serlo-editor/plugins/article/editor.tsx
```diff
@@ -22,8 +22,18 @@ export function ArticleEditor({ editable, state }: ArticleProps) {
...
+  useEffect(() => {
+    const explanationTextPlugin = ref.current?.querySelector<HTMLDivElement>(
+      '.explanation-wrapper > .plugin-wrapper-container'
+    )
+    explanationTextPlugin?.focus()
+  }, [])
+
```
elbotho, line 34: what do you think about moving this into `serlo-template-plugins/article.tsx`? 
This way we don't need an additional wrapper div and can also easily change it to focus the title for new articles for examples.

src/serlo-editor/plugins/article/editor.tsx
```diff
@@ -22,8 +22,18 @@ export function ArticleEditor({ editable, state }: ArticleProps) {
...
   const modalStrings = useEditorStrings().templatePlugins.article.addModal
 
+  // Focus the article introduction explanation on page load
+  // TODO: Move this into a focus-dedicated hook
+  const ref = useRef<HTMLDivElement | null>(null)
+  useEffect(() => {
+    const explanationTextPlugin = ref.current?.querySelector<HTMLDivElement>(
+      '.explanation-wrapper > .plugin-wrapper-container'
+    )
+    explanationTextPlugin?.focus()
```
elbotho, line 32: since you have a comment already this would also work:
```suggestion
    ref.current?.querySelector<HTMLDivElement>(
      '.explanation-wrapper > .plugin-wrapper-container'
    )?.focus()
```

elbotho (commented):
Works nicely for articles.

Remaining problems:
- but not for other entities like e.g. [exercise](http://localhost:3000/entity/create/Exercise/60730)
- when creating a new plugin it does not get focused (at all, but also not where it's supposed to get focus via autofocusref)

src/serlo-editor/plugins/article/editor.tsx
hejtful: The name of the ref constant would definitely need to change when moved to a hook, so I didn't put much effort into it.

src/serlo-editor/plugins/article/editor.tsx
hejtful: I'm not familiar with `serlo-template-plugins`, but it makes sense what you wrote. Would you like to add that to the PR?

src/serlo-editor/plugins/article/editor.tsx
hejtful: Did that already, and then moved it back to the current state, because a dangling property/method accessor is hard to detect. This is more code, but makes sure that the `?.focus` part doesn't get overlooked.

elbotho:

PR #2785, opened 2023-08-24
refactor(focus): replace focus next & focus previous

elbotho:
@hejtful this seems to work pretty similar like it did before in my first tests.
One inconsistency: I can use arrow keys to move into equations now but can't jump into table cells. 

We should probably move it out of the focus slice at some point, but for now I'm happy with minimal changes that don't break anything 😸

### DIFF
--- a/src/serlo-editor/store/focus/slice.ts
+++ b/src/serlo-editor/store/focus/slice.ts
@@ -18,13 +18,17 @@ export const focusSlice = createSlice({
       if (!state || !action.payload) return state
       const next = findNextNode(action.payload, state)
       if (!next) return state
+      const nextDomElement = document.getElementById(next) as HTMLDivElement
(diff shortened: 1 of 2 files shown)


src/serlo-editor/store/focus/slice.ts
```diff
@@ -18,13 +18,19 @@ export const focusSlice = createSlice({
...
-      const next = findPreviousNode(action.payload, state)
-      if (!next) return state
-      return next
+      const previous = findPreviousNode(action.payload, state)
+      if (!previous) return state
+      const previousDomElement = document.getElementById(
+        previous
+      ) as HTMLDivElement
+      previousDomElement?.focus()
+      return previous
```
elbotho: can you tell me where the returned id from `focusPrevious` then in turn changes the focus?

elbotho:

PR #2786, opened 2023-08-24
refactor(focus): focus newly created plugins with focusNext

elbotho:
trying simple approach to focus plugins after their creation.
if we could move it into `insertPluginChildAfter` this could be nice, what do you think?

what works:
plugin get's focus

what doesn't:
does not focus the correct part of the plugin (autofocusref or slate)


elbotho:

PR #2789, opened 2023-08-25
Use slate normalization to prevent consecutive empty lines.

LarsTheGlidingSquirrel:
Seems to work great while entering text. However, I found an issue when pasting text. 

TODO: 
- [ ] Test for other issues
- [ ] Fix pasting issue. 

Details: 
Pasting text with blank lines is broken. 
Pasting ...
```
A 

B
```
... will result in ...
```
A
B
```
For some reason the normalization is triggered even though it should not. While debugging I observed that the normalization is called with the pasted text but the slate node that should contain "B" is empty. So it is called with ...
```
A


```
... for some reason and triggers the normalization.


elbotho:
followed up by #2831

PR #2790, opened 2023-08-25
fix(text-plugin): hide decorators on formulas / show placeholder on empty text plugins

elbotho:
some more quick fixes so we can soon deploy changes to production again.

- a line with just a formula was falsely detected as empty
- manual fix for consecutive empty lines removed in favour of [normalization approach](https://github.com/serlo/frontend/pull/2789) (did now work very well anyway)
- currently it's hard to spot empty text-plugins, this is a bit hacky but works fine for now


src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -163,7 +177,7 @@ export function TextEditor(props: TextEditorProps) {
             : decorateEmptyLinesWithPlaceholder
         }
         placeholder={
-          editable && shouldShowStaticPlaceholder
+          (editable && shouldShowStaticPlaceholder) || !editor.selection
```
hejtful: I can't figure out why this was necessary, could you please explain? 😄 

>a line with just a formula was falsely detected as empty

I guessed it was this ^, but couldn't reproduce it on staging.

hejtful (approved):
Looks good!

src/serlo-editor/plugins/text/components/text-editor.tsx
elbotho: For example when you insert a box plugin the newly created text plugin inside is not focused by default.
In this case slate does not have a selection and the text plugin would be basically invisible to the users.

It would be even nicer to have the decorator show in this cases but since somehow this does not work for slate without a selection (could not get it to work).

Should I explain this in a comment or can you maybe think of a better solution?

hejtful: Thanks for explaining! No idea for a better solution, a comment would be nice

elbotho: ah also it should probably be part of `shouldShowStaticPlaceholder` in the first place!

PR #2791, opened 2023-08-25
refactor(rows-plugin): remove first floating plus button, refactor add buttons

elbotho:
- we might remove the floating add buttons soon so it's probably a good start to simplify them :) 
- adding a plugin before the current plugin is an (edge) use case, but can also be archived by adding and dragging
- @hejtful I know I create conflicts with the focus branch, but I will also clean them up if we want to merge the PR at some point :)


src/serlo-editor/plugins/rows/components/add-row-button-floating.tsx
```diff
@@ -0,0 +1,44 @@
...
+import clsx from 'clsx'
+
+import { FaIcon } from '@/components/fa-icon'
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
+import { tw } from '@/helper/tw'
+
+interface AddRowButtonFloatingProps {
+  onClick: React.MouseEventHandler<HTMLButtonElement>
+  focused?: boolean
+  hideAddButton?: boolean
```
hejtful: Minor: I love descriptive names, but this seems like too much 😄 Maybe just `hidden`?

src/serlo-editor/plugins/rows/components/row-editor.tsx
```diff
@@ -38,11 +34,7 @@ export function RowEditor({
     <div
       key={row.id}
       ref={dropContainer}
-      className={clsx(
-        'rows-child relative -ml-12 pl-12',
-        isFirst && 'first',
-        isLast && 'last'
-      )}
+      className={clsx('rows-child relative -ml-12 pl-12')}
```
hejtful:
```suggestion
      className="rows-child relative -ml-12 pl-12"
```

src/serlo-editor/plugins/rows/editor.tsx
```diff
@@ -22,6 +18,9 @@ export function RowsEditor({ state, config, id, editable }: RowsProps) {
     })
   }
 
+  const showLargeAddButton =
+    parentType === EditorPluginType.Article || parentType?.startsWith('type-')
```
hejtful: `parentType?.startsWith('type-')` deserves a comment imo, most people (including me) probably don't understand the logic behind this.

Edit: Assigning it to a variable with a descriptive name is also fine with me.

hejtful (commented):
Two super minor things and one plea for a mini-documentation-comment 😄

src/serlo-editor/plugins/rows/components/add-row-button-floating.tsx
elbotho: how about `hideAddRowButtonFloating`? 😉
good point :) leftover from when I was not only this button

src/serlo-editor/plugins/rows/editor.tsx
elbotho: I hope you don't actually feel like you have to plea for documentation :D
put it into a variable to explain what it does. Okay like this?

hejtful: I think I just misused "plea", I thought it means "ask nicely", my bad 😄 

Looks great to me!

hejtful (approved):

PR #2792, opened 2023-08-26
fix(plugin-image): Show error when image can't be pasted

### LINKED ISSUE #2754: editor(rows/image): no error message when dragging image to upload
when dragging an image-file into the rows plugin:
- upload works: all good, image plugin appears
- upload fails (e.g. filesize error) nothing happens

in the second case we should show an error message as well.

AsterMiha:


src/serlo-editor/plugins/image/index.ts
```diff
@@ -83,5 +86,5 @@ export type ImageProps = EditorPluginProps<ImagePluginState, ImageConfig>
 
 export interface ImagePluginConfig {
   upload: UploadHandler<string>
-  validate: UploadValidator
+  validate: UploadValidator<Array<{ errorCode: number; message: string }>>
```
hejtful: There is this [FileError interface](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/image/image-with-serlo-config.ts#L39) defined already, which you could reuse here. What do you think?

If you do reuse it, please [import it as a type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export), to avoid circular dependency issues.

AsterMiha: Fixed. Thanks for pointing this out!

hejtful (approved):

elbotho (approved):
Thank you, 
interestingly `unauthenticated` error does not show but I think this is unrelated to this PR and not as important.

PR #2794, opened 2023-08-27
refactor(editor): Split richText formatting into 2 options

### LINKED ISSUE #2793: Default italic styling in the image plugin's caption editor
Currently, the image plugin caption editor has a default italic styling.

But, italic is enabled as a formatting option in the caption editor, and it's not obvious when the italic option is active, by just looking at the caption text.

We should consider either:
- removing the default italic styling, so that the intended italic styling is obvious
- removing the italic formatting option from the toolbar

AsterMiha:
Previously, the richText formatting option included both bold and italic.
Since we would like to keep the default italic formatting of image captions, only the bold button should show up in the toolbar.
Now italic and bold formatting can be accessed separately through richTextBold and richTextItalic.

### DIFF
--- a/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
+++ b/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/hooks/use-formatting-options.tsx
@@ -73,12 +73,12 @@ const isRegisteredTextPlugin = (
 const registeredHotkeys = [
   {
     hotkey: 'mod+b',
-    option: TextEditorFormattingOption.richText,
+    option: TextEditorFormattingOption.richTextBold,
     handler: toggleBoldMark,
   },
   {
     hotkey: 'mod+i',
-    option: TextEditorFormattingOption.richText,
+    option: TextEditorFormattingOption.richTextItalic,
     handler: toggleItalicMark,
   },
   {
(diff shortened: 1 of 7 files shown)


hejtful (changes requested):
Smooth and clean changes 👍 

There is one small bug, table cells don't have bold and italic formatting options anymore. This might be [the relevant line](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/serlo-table/editor.tsx#L32).

A weird thing is that [this line](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/image/index.ts#L26) also contains the old `richText` option, but the caption works fine. Maybe that line can be removed completely? If not, then it might be good to update it with the new options.

Edit: A suggestion for improvement would be to replace those strings with the `TextEditorFormattingOption`, if possible. Up to you :)

AsterMiha:
> A weird thing is that [this line](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/image/index.ts#L26) also contains the old `richText` option, but the caption works fine. Maybe that line can be removed completely? If not, then it might be good to update it with the new options.

The line doesn't seem to influence anything so I removed it

hejtful (approved):

AsterMiha:
> > A weird thing is that [this line](https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/image/index.ts#L26) also contains the old `richText` option, but the caption works fine. Maybe that line can be removed completely? If not, then it might be good to update it with the new options.
> 
> The line doesn't seem to influence anything so I removed it

A similar situation seemed to be happening with the box plugin as well

elbotho (approved):
Nice, thank you!

PR #2797, opened 2023-08-28
fix(editor): prevent removing last plugin

### LINKED ISSUE #2773: editor: prevent removing the last plugin in RowsEditor
Currently, we can have a RowsEditor without any rows/plugins, if we remove the last plugin. We should prevent this, so that we always have at least one empty text plugin.

hejtful:
Fixes https://github.com/serlo/frontend/issues/2773 and is a follow-up to https://github.com/serlo/frontend/pull/2791

**Backspace and delete plugin removal:**
We actually already had this check when merging plugins with content, but removing empty plugins happened before this check.
Now, the check is moved up, and refactored a bit.

**Drag'n'drop plugin removal:**
When the only plugin of a rows plugin is dragged and dropped outside its parent rows plugin, replace it with an empty text plugin.
Note: Before, we would first insert the dropped plugin, before removing it from its original position. That meant that, with the new logic, the replacement empty text plugin would be focused, because it was inserted last. Now, we are first removing the dropped plugin from its original position (and replacing it with an empty text plugin, if necessary), and then inserting it in the drop position.

### DIFF
--- a/src/serlo-editor/plugins/rows/editor-renderer.tsx
+++ b/src/serlo-editor/plugins/rows/editor-renderer.tsx
@@ -15,6 +15,7 @@ import {
   selectSerializedDocument,
   store,
 } from '@/serlo-editor/store'
+import { EditorPluginType } from '@/serlo-editor-integration/types/editor-plugin-type'
 
 interface RowDragObject {
   id: string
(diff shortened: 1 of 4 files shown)


elbotho (approved):
Thank you, very elegant fix.
Unfortunately #2773 is only partially done since I can still drag the last plugin out of a rows plugin.
(so we should add an empty text plugin before we move the last child I guess?)

src/serlo-editor/plugins/rows/editor-renderer.tsx
```diff
@@ -54,10 +55,16 @@ export function EditorRowRenderer({
...
+          // Remove the dragged plugin from its original rows plugin
           rows.set((list) => {
             const index = list.findIndex((id) => id === row.id)
             return R.remove(index, 1, list)
           })
+
+          // If the dragged plugin was the only plugin in the current rows plugin,
+          // add an empty text plugin to replace it
+          if (rows.length > 1) return
+          rows.insert(0, { plugin: EditorPluginType.Text })
```
elbotho: opinion: this is one of the cases where the early return does not help me understand.

PR #2806, opened 2023-08-28
chore(fetcher): adapt license fetcher to migrated details

elbotho:
(correct api version should now be deployed in staging and prod!)


Entkenntnis (approved):
thanks

PR #2807, opened 2023-08-29
Add `PluginToolMenuCustomizationContext` for edu-sharing integration

LarsTheGlidingSquirrel:
Small pull request that enables the edu-sharing integration to hide this button from the plugin toolbar menu: 
image

Reason: Anchor links contain an entityId which we do not have in the edu-sharing integration. 

Also: Running `yarn` added a (apparently) missing file to the yarn cache directory

### DIFF
--- a/src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-default-tools.tsx
+++ b/src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-default-tools.tsx
@@ -1,8 +1,9 @@
 import { faClone, faTrashAlt } from '@fortawesome/free-solid-svg-icons'
-import { useCallback } from 'react'
+import { useCallback, useContext } from 'react'
 
 import { AnchorLinkCopyTool } from './anchor-link-copy-tool'
 import { DropdownButton } from './dropdown-button'
+import { PluginToolMenuCustomizationContext } from './plugin-tool-menu-customization-context'
 import { useEditorStrings } from '@/contexts/logged-in-data-context'
 import {
   insertPluginChildAfter,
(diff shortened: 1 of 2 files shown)


elbotho:
Interesting. I was pretty sure [those lines](https://github.com/serlo/frontend/blob/40ff87137cb2088dbdf84cde3743f764036311f8/src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx#L20-L22) would hide it in edusharing anyway. What is `serloEntityId` in your context?

LarsTheGlidingSquirrel:
> Interesting. I was pretty sure [those lines](https://github.com/serlo/frontend/blob/40ff87137cb2088dbdf84cde3743f764036311f8/src/serlo-editor/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx#L20-L22) would hide it in edusharing anyway. What is serloEntityId in your context?

I got this error: 
https://github.com/serlo/frontend/blob/40ff87137cb2088dbdf84cde3743f764036311f8/src/contexts/entity-id-context.ts#L10C16-L10C16

Looking at it now, I see other possible solutions. For example: Provide some entityId in edu-sharing that is interpreted as "not a valid id" and makes the button hidden. What do you prefer @elbotho?

elbotho:
> Looking at it now, I see other possible solutions. For example: Provide some entityId in edu-sharing that is interpreted as "not a valid id" and makes the button hidden. What do you prefer @elbotho?

Hmmm, also a bit hacky… I think we could consume the context directly in the `AnchorLinkCopyTool` and just don't display it if it returns null (instead of throwing). This way we don't have to change any other frontend code. What do you think?

LarsTheGlidingSquirrel:
>Hmmm, also a bit hacky… I think we could consume the context directly in the AnchorLinkCopyTool and just don't display it if it returns null. This way we don't have to change any other frontend code. What do you think?

I prefer handling it in `PluginDefaultTools` and only rendering `AnchorLinkCopyTool`, when it should be rendered. That would be more clear to me. But I am fine with both and there are definitely more important code clarity issues I think. 😅

elbotho:
> and only rendering AnchorLinkCopyTool, when it should be rendered

I think it's mostly a matter of taste. JS is bundled anyway and it does not render in both cases.
If you prefer you could also move the entity check out of the `AnchorLinkCopyTool`.

PR #2812, opened 2023-08-29
feat(text-plugin): also search for pluginType in suggestion search

elbotho:
not functional any more, but I always wanted to try the new js [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) 😉.

why?
mostly so devs can type "injection" and actually find something 😸

### DIFF
--- a/src/serlo-editor/plugins/text/hooks/use-suggestions.tsx
+++ b/src/serlo-editor/plugins/text/hooks/use-suggestions.tsx
@@ -230,13 +230,24 @@ function filterPlugins(
   const search = text.replace('/', '').toLowerCase()
   if (!search.length) return plugins
 
-  const startingWithSearchString = plugins.filter(({ title }) => {
-    return title.toLowerCase()?.startsWith(search)
-  })
-  const containingSearchString = plugins.filter(({ title }) => {
-    const value = title?.toLowerCase()
-    return value?.includes(search) && !value?.startsWith(search)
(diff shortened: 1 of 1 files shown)


src/serlo-editor/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -230,13 +230,24 @@ function filterPlugins(
...
+      filterResults.add(entry)
+    }
   })
-  const containingSearchString = plugins.filter(({ title }) => {
-    const value = title?.toLowerCase()
-    return value?.includes(search) && !value?.startsWith(search)
+
+  // title includes search string
+  plugins.forEach((entry) => {
+    if (entry.title.toLowerCase()?.includes(search)) {
```
hejtful: Why do you need `?` after `toLowerCase` here and on line 238? 😄 It's supposed to always return a string

hejtful (approved):
Fancy 👍

src/serlo-editor/plugins/text/hooks/use-suggestions.tsx
elbotho: Good catch. To be honest I copy pasted your line 😇

hejtful: giphy

PR #2813, opened 2023-08-29
refactor(editor store): clean up `focus` slice

hejtful:
Changes:
- Remove `root` slice, replace its usage with one constant and one `useState`
- Reduce selectors calling each other as much as possible
- Move selectors that only select `documents` slice state to `documents` slice
- Remove some unused arguments from plugin state types
- Refactor selectors that were recursive before, so that they are only called once and use recursive functions instead
- Use more descriptive and accurate names for some selectors

Result: 
- Cleaner `focus` slice, with clearer overview of focus-related code
- Most selectors only select state from their own slice

Follow-up:
- Move `plugin` and `root` slice sagas to `documents` slice, as they are working with `documents` slice state


src/serlo-editor/plugin/object.ts
```diff
@@ -16,17 +16,16 @@ import {
...
-  }
+  types: Ds
 ): ObjectStateType<Ds> {
   type S = StateTypesSerializedType<Ds>
   type T = StateTypesValueType<Ds>
   type U = StateTypesReturnType<Ds>
 
+  const getObjectChildIds = (children: {
+    [K in keyof Ds]: { id: string }[]
+  }) => R.flatten<readonly FocusableChild[][]>(R.values(children))
```
elbotho, line 27: Does the naming of `FocusableChild` still make sense?

elbotho (approved):
Great idea to refactor this and I love how it turned out. 
Also moving the helper function to a different file makes sense 👍
Did no testing after our session so maybe we merge it after the deploy today?

hejtful:
I'm currently investigating to see if `selectChildTreeOfParent` is actually used for anything other than the `id` of the parent. If not, I'll refactor that too, that's why it's Draft again.

src/serlo-editor/plugin/object.ts
hejtful: Any proposals? 😄

PR #2817, opened 2023-08-29
Do not focus next cell in formula plugin (LaTeX editor)

CodingDive:
Issue reported from Peter here https://trello.com/c/hCLS1f2O/263-closing-latex-dialog-causes-focus-forwarding


hejtful (commented):
Because the focus is reset, the left border and the drag'n'drop handle of the Equations plugin will disappear. I'd personally prefer that not to happen, but maybe @elbotho has a different opinion.

Btw, just so I'm up-to-date, did the decision to remove the button completely change? 😄

CodingDive:
> Btw, just so I'm up-to-date, did the decision to remove the button completely change? 😄

I'd like to keep the button as think it's better UX to be able to close modals.

elbotho (approved):
still not a fan of the additional (now reasonable) complexity added here but:
- works okay now
- most of the problem is the original code and we'll probably want to refactor this whole plugin at some point

PR #2825, opened 2023-08-30
app directory migration

### LINKED ISSUE #2823: another attempt at adopting app directory
One big challenge is handling of i18n, I will start another experiment and integrate previous attempts.

Entkenntnis:
This PR will migrate all of `pages` into `app`. It's faster to do it in one go and many things are already working nicely.

list of current to-dos:

- review metadata handling: replace headtags with generateMetaData (it's moved out because head tags should be available before streaming, so it has to be fairly independant of page data, which we are not doing consistently)
- bundle size analyzer not showing app-dir pages - https://github.com/hashicorp/nextjs-bundle-analysis/issues/42
- every page has to pass locale and instanceData to FrontendClientBase -> search for abstraction
- get translations on server is a bit inconvenient because lang is only accessible in page.tsx, maybe use `cache`
- navigation through router.push not working in vercel deployments (localhost is fine) (rewrite is not picked up ...) https://github.com/vercel/next.js/issues/54783
- old router is used here and there
- migrate api routes to route handlers
- restore progress bar (replace Router events)
- fix injections
- access pathname is difficult for meta data https://github.com/vercel/next.js/discussions/50189

preview: https://frontend-git-2823-another-attempt-at-adopting-app-234955-serlo.vercel.app/1553


next.config.js
```diff
@@ -84,4 +84,14 @@ module.exports = withBundleAnalyzer({
       skipDefaultConversion: true,
     },
   },
+  async rewrites() {
+    return process.env.NODE_ENV === 'development'
+      ? [
+          {
+            source: '/app-dir-testing/:path*',
+            destination: '/app-dir-testing/de/:path*', // <- you can set the language of the frontend here
```
Entkenntnis: @elbotho 

Approach here: Make language static. You can still easily change it to another language if you want to develop for it. This way, no middleware is needed (and this makes the mental model a lot easier)

Entkenntnis:
This attempt got quite far, but due to some remaining bugs and challenges (see description), we will postpone this for some more months.

PR #2827, opened 2023-08-30
Fix cutoff content (height) of fractions in math formula renderer

CodingDive:
This bug is somewhat hard to reproduce. Tried it in Firefox and chrome. On article https://de.serlo.org/mathe/261500/261500 I was able to see the box being cutoff, locally it didn't in Chrome and at least the box seemed fine.
In Firefox, everything worked fine for me but broke for Anne (see picture).
Meanwhile, for me the `g` at the bottom of a fraction was cutoff a little bit. 

The fix was to add a tiny bit of padding, and more importantly, change the display type of the span from inline to inline block (so that it autogrows with content).

image

### DIFF
--- a/src/components/content/math-span.tsx
+++ b/src/components/content/math-span.tsx
@@ -32,7 +32,7 @@ export function MathSpan({ formula }: MathSpanProps) {
     <>
       <KaTeXStyles />
       <span
-        className="[page-break-inside:avoid]"
+        className="inline-block py-1 [page-break-inside:avoid]"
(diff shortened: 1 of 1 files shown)


CodingDive:
Please check articles like serlo.org/1545 (localhost:3000/1545) that have an exponent on top of a fraction. It works for me locally in various browsers and with various zoom levels, but I'm not 100% convinced that it works in every case.

hejtful (approved):
The cut-off bug is fixed 👍

Entkenntnis (approved):

PR #2829, opened 2023-08-31
fix(editor): hide anchor link tool in edusharing

elbotho:
trying alternative approach to #2807


LarsTheGlidingSquirrel (approved):
That would work! Thank you! :) 
We could even remove the check `!serloEntityId` in AnchorLinkCopyTool I think.

PR #2831, opened 2023-08-31
fix(plugin-text): restrict amount of empty lines

### LINKED ISSUE #2810: Prevent successive empty lines
> Preventing two empty lines only works for new lines (when pressing enter twice). If there are already existing empty lines before the current selection, adding a new line is still possible.

yeah… do you have an easy idea so solve that? Otherwise I thought it's okay for now (not the only way to do forbidden things in the editor) and create a backlog issue about adding a nice slate normalizer or something like that.

_Originally posted by @elbotho in https://github.com/serlo/frontend/issues/2733#issuecomment-1677079342_

hejtful:
Resolves https://github.com/serlo/frontend/issues/2810

Continuation of https://github.com/serlo/frontend/pull/2789

Test PR: https://github.com/serlo/frontend-e2e-tests/pull/70


elbotho:
Thanks for working on this. Did some testing. Seems to work fine.
Only thing: right now it's quite easy to create persisting double empty lines by going back into an existing empty line and pressing enter again.
image

Entkenntnis (approved):
thank you, solution is robust and works

src/serlo-editor/plugins/text/components/text-editor.tsx
```diff
@@ -121,6 +126,26 @@ export function TextEditor(props: TextEditorProps) {
...
+  // `withEmptyLinesRestriction` plugin.
+  // `useEffect` and event delegation are used because `<Editable`
+  // `onBlur` doesn't work when custom-empty-line-placeholder is
+  // shown before bluring the editor. More info on event delegation:
+  // https://www.quirksmode.org/blog/archives/2008/04/delegating_the.html
+  useEffect(() => {
+    const handleBlur = () => {
+      // @ts-expect-error `operation` can only be `BaseOperation`, but we use
+      //                  this 'blur' hack to do special normalization on blur.
+      editor.normalize({ force: true, operation: 'blur' })
```
elbotho: just a suggestion:

```suggestion
      // @ts-expect-error custom `operation` to do special normalization only on blur.
      editor.normalize({ force: true, operation: { type: 'blur_container' } })
```

elbotho: (would need change in `with-empty-lines-restriction.ts` as well ofc that might make it not worth it…)

elbotho (approved):
Thanks for the thoroughly documented solution. 
It's surprisingly complex 🙈, but it's super tidy with`withEmptyLinesRestriction` 👍

src/serlo-editor/plugins/text/components/text-editor.tsx
hejtful: Looks good to me, please feel free to add it if you want.

src/serlo-editor/plugins/text/plugins/with-empty-lines-restriction.ts
```diff
@@ -0,0 +1,103 @@
+import { Editor, Element, Node, Path, Transforms } from 'slate'
+import { ReactEditor } from 'slate-react'
+
+// Slate normalization docs: https://docs.slatejs.org/concepts/11-normalizing
+// While editing a Text plugin: Two adjacent empty lines are allowed,
+// so that one of them could be transformed into another plugin,
+// using the plus button.
```
LarsTheGlidingSquirrel: Suggestion: "... are allowed for UX reasons."

src/serlo-editor/plugins/text/plugins/with-empty-lines-restriction.ts
```diff
@@ -0,0 +1,103 @@
...
+  const { normalizeNode } = editor
+
+  editor.normalizeNode = (entry, { operation } = { operation: undefined }) => {
+    const [node, path] = entry
+    const isBlur = (operation as unknown as string) === 'blur'
+
+    if (shouldNodeBeRemoved(editor, node, path, isBlur)) {
+      Transforms.removeNodes(editor, { at: path })
+      // TODO: Show warning?
+      if (isBlur) ReactEditor.blur(editor)
```
LarsTheGlidingSquirrel, line 19: Is this line needed? As far as I understand, the blur event already occurred at this point.

src/serlo-editor/plugins/text/plugins/with-empty-lines-restriction.ts
```diff
@@ -0,0 +1,104 @@
...
+  // that is also an empty paragraph, should be removed.
+  const previousSibling = Node.get(editor, previousSiblingPath)
+  const isPreviousSiblingEmptyParagraph = isEmptyParagraph(
+    editor,
+    previousSibling
+  )
+  if (isBlur) return isPreviousSiblingEmptyParagraph
+
+  // Collect the amount of empty paragraph siblings. It should never be more than 1.
+  let amountOfEmptyParagraphSiblings = isPreviousSiblingEmptyParagraph ? 1 : 0
```
LarsTheGlidingSquirrel, line 58: Maybe this line and the following could be simplified by only checking upwards: 
"If the current line has two empty lines above it -> Remove current line"

LarsTheGlidingSquirrel (approved):
Added some ideas for improvements. Awesome that it works now!

src/serlo-editor/plugins/text/plugins/with-empty-lines-restriction.ts
LarsTheGlidingSquirrel: I tried it but it lead to a different behavior and allowed three empty lines in some cases. Abandoned my idea because it works correctly right now.

src/serlo-editor/plugins/text/plugins/with-empty-lines-restriction.ts
LarsTheGlidingSquirrel: Done

PR #2837, opened 2023-09-03
refactor: Refactor focus handling (WIP)

kulla:
This PR refactors the focus handling in the editor (currently in the `equations` plugin) and it does not need the internal grid handler for the focus.

* Link to test: https://frontend-git-refactor-focus-serlo.vercel.app/entity/repository/add-revision/277232 (test the equation plugin with "tab" and "shift + tab")

Todos:

- [x] Fix toolbar in equations plugin
- [x] Fix tab for equations plugin / text plugin
- [ ] Fix bug that click in text plugin doesn't show cursor
- [ ] Focus a correct cell when clicking on an unfocused table plugin
- [ ] Focus a newly added plugin 
- [ ] Option: Add Slate Selection to focus Path


src/serlo-editor/core/editor.tsx
```diff
@@ -132,10 +190,18 @@ function InnerDocument({
   if (!isInitialized) return null
 
   return (
-    <div className="relative" ref={wrapperRef}>
+    <div
+      className="relative"
+      ref={(div) => {
+        wrapperRef.current = div
+        div?.addEventListener('focusin', onFocusinHandler)
+      }}
```
CodingDive:
```suggestion
      ref={wrapperRef}
```

better to handle attaching the eventListener in a useEffect.

```
useEffect(() => {
  if (wrapperRef.current) {
      wrapperRef.current.addEventListener('focusin', onFocusinHandler);

      return () => {
          wrapperRef.current.removeEventListener('focusin', onFocusinHandler);
      };
  }
}, [onFocusinHandler]);
```

kulla: This was my first attempt but it failed for me. In the first rendering wrapperRef.current is null and since it is a reference the component does not get rendered again once the HTML Element is mounted (that's my theory - the bug was that the listener was not added property). Maybe I did something else wrong. I would be happy if someone can fix this.

hejtful: Fixed.

- Conditional rendering of the component means that the component renders `null`, until `initRootSaga` and its recursion is complete. Once this is done, the `isInitialized` state value is switched to `true`. Therefore, we can use `isInitialized` as a dependency for the event attaching `useEffect`.
- Compared to Mikey's example, `wrapperRef.current` was moved into a constant, because `wrapperRef.current` usually no longer exists when the clean-up function is called, so the event listener doesn't get removed.
- The condition for setting the focus path was moved inside the setter, so that the `focusPath` wouldn't be a dependency for `onFocusinHandler` callback. This way, we avoid `onFocusinHandler` being re-declared whenever `focusPath` changes, which in turn prevents the event attaching `useEffect` to fire whenever `focusPath` changes.

src/serlo-editor/editor-ui/focus-helper.ts
```diff
@@ -0,0 +1,34 @@
+import { startsWith } from 'ramda'
+
+import { FocusPath } from '../types'
+
+interface PathAndFocus {
+  pluginId: string
+  path: Array<string | number>
+  targetPath?: Array<string | number>
+  currentFocus: FocusPath
```
hejtful, line 9: This interface reminds me of ZZ Top. In ZZ Top, there are two guys with huge beards, and one other guy. The other guy's last name is Beard.

Same here, we have `path` and `targetPath`, but it's the `currentFocus` whose type is `FocusPath`. 😁

hejtful:
Current issue: When using Tab to switch from a text field to a math field, both text and math controls will be shown in the toolbar.

Screenshot 2023-09-07 at 10 46 30

Edit: This is caused by the current Redux store focus management system working with `onMouseDown` event, instead of `onFocus` or `onBlur`. This seems to me like quite an important issue to tackle early.

2nd edit: fixed.

hejtful:
Another current issue: Click outside of Equations plugin doesn't work, the toolbar will still be shown even after clicking outside.

Edit: This one was fixed in the commit below ([fix(editor core): click outside handler](https://github.com/serlo/frontend/pull/2837/commits/f76a6c6e59af2a3e845f2d29c2cebba4bb2c0fb5))

Entkenntnis:
superseeded by #2878

Entkenntnis:
keep branch for inspiration on some additional ideas to manage focus

PR #2840, opened 2023-09-04
fix(plugin-serlo-table): column nav positioning

hejtful:
https://trello.com/c/9bkKf26X/278-tabellenspalten-lassen-sich-nicht-l%C3%B6schen

Bug: The column nav (add or remove column) used to be above the column header, but it was hidden behind the toolbar.

Quick fix (suggested by @elbotho): Move the column nav below the column.

Note: Static add new row button had to be moved down a bit, to make space for the column nav.

### DIFF
--- a/src/serlo-editor/plugins/serlo-table/editor.tsx
+++ b/src/serlo-editor/plugins/serlo-table/editor.tsx
@@ -208,7 +208,7 @@ export function SerloTableEditor(props: SerloTableProps) {
             </>
           ) : null}
         </nav>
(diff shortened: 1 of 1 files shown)


Entkenntnis (approved):
thank you, can confirm that this works in https://frontend-git-fix-serlo-table-column-nav-positioning-serlo.vercel.app/entity/repository/add-revision/1545

CodingDive (approved):
Wow really simple change! Great job! :raised_hands:

PR #2841, opened 2023-09-04
fix(plugin-text): Slate deleting void nodes when sibling node deleted

### LINKED ISSUE #2842: merging plugin that only contains void nodes will delete void nodes
https://frontend-git-fix-slate-void-node-deletion-serlo.vercel.app/entity/repository/add-revision/1565

grafik

Pressing backspace at the start of the plugin will merge plugins, but also delete both formulas.

Merge is successful if any line contains some text:

grafik

my guess is that isEmptyPlugin does not recognize void nodes.

hejtful:
https://trello.com/c/H3eIoRy2/279-ux-zeilenabst%C3%A4nde-und-formelumgebung

Second commit also resolves https://github.com/serlo/frontend/issues/2842

### DIFF
--- a/src/serlo-editor/plugins/text/components/text-editor.tsx
+++ b/src/serlo-editor/plugins/text/components/text-editor.tsx
@@ -19,6 +19,7 @@ import { useEditorChange } from '../hooks/use-editor-change'
 import { useSlateRenderHandlers } from '../hooks/use-slate-render-handlers'
 import { useSuggestions } from '../hooks/use-suggestions'
 import { useTextConfig } from '../hooks/use-text-config'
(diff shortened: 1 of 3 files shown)


Entkenntnis (approved):
@hejtful Thank you for tackling this, I have tested it and can confirm that I am able to delete the empty lines now without deleting the formulas.

While testing, I discovered another bug that I documented in https://github.com/serlo/frontend/issues/2842

hejtful:
Nice find, thank you!

CodingDive (approved):
Fixes the reported bug, thank you!
Sad that all this extra code is needed, but it's great that you linked to the slate issue. Maybe in future versions this will be already the built-in behavior.

Entkenntnis (approved):
thank you

PR #2852, opened 2023-09-07
set comment status and filter by status

### LINKED ISSUE #2851: allow sorting discussions by status

Entkenntnis:


src/components/comments/comment-area-all-threads-thread.tsx
```diff
@@ -70,32 +62,37 @@ export function CommentAreaAllThreadsThread({
...
+                submitEvent('thread_status_set')
+                onMutate()
+                void mutateStatus({
+                  id: [thread.id],
+                  status: e.target.value as CommentStatus,
+                })
+              }}
+            >
+              <option value="open" className="bg-white">
+                offen
```
hejtful, line 88: I guess no translations because the whole feature is still a work in progress?

hejtful (approved):
Code looks good to me!

I'd be happy to test it as well, if you want. I'd just need a link to an entity with comments.

src/components/comments/comment-area-all-threads-thread.tsx
Entkenntnis: we only use this feature for the german language instance, so all of the UI is not translated

Entkenntnis:
You can test it locally under http://localhost:3000/discussions

As soon as the API is deployed, we can continue the testing in staging.

PR #2857, opened 2023-09-10
Fix merging issue between a block starting with a list and another text block

### LINKED ISSUE #2798: editor: Slate error when merging a text plugin with another text plugin that has a list
In the attached video, the beginning of "image" headline is selected, and "Backspace" key is pressed to merge the two text plugins. A Slate error is thrown.

Relevant files:
- https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/text/hooks/use-editable-key-down-handler.tsx#L136
- https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/text/utils/document.ts#L54

Note: Might be worth adding an e2e test for this case.

https://github.com/serlo/frontend/assets/13198821/e63699cb-96cd-477e-8abb-2e9d39dfb773

AsterMiha:
A new line will be added at the start of the block to avoid the error the list causes (described [here](https://github.com/serlo/frontend/issues/2798)).

This fixes the issue as long as the block containing the list is interacted with before we attempt to merge with the other block.
To completely avoid the issue, we need to make sure the exercise is loaded together with this extra line to start with.

### DIFF
--- a/src/serlo-editor/plugins/text/hooks/use-editor-change.tsx
+++ b/src/serlo-editor/plugins/text/hooks/use-editor-change.tsx
@@ -2,7 +2,6 @@ import { useCallback, useMemo, useRef } from 'react'
 import { Descendant, Editor, Transforms } from 'slate'
 
 import type { TextEditorProps } from '../components/text-editor'
-import { CustomElement, ListElementType, Paragraph } from '../types/text-editor'
 
 interface UseEditorChangeArgs {
   editor: Editor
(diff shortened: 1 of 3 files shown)


hejtful:
> we need to make sure the exercise is loaded together with this extra line to start with

Maybe using normalization would solve this?

Something similar in this PR: https://github.com/serlo/frontend/pull/2831
And we already have a `withLists` plugin file defined, which could be a good place to add the normalization modifications as well: https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/text/plugins/lists.ts

AsterMiha:
Moved the normalization to the list plugin and inserted the extra line at load time too
Thanks for the help, @hejtful!

hejtful:
Looks fine to me! But, @elbotho would be the one to judge the actual behavior: is having an empty line before a list acceptable?

AsterMiha:
> Looks fine to me! But, @elbotho would be the one to judge the actual behavior: is having an empty line before a list acceptable?

An alternative would be to only modify the node right before the merge happens and delete this extra line once the operation is complete. This should leave no visible changes.
I can investigate this a bit to see if this solution would cause other issues.

elbotho:
>An alternative would be to only modify the node right before the merge happens and delete this extra line once the operation is complete. This should leave no visible changes.

If that is not to complex I'd def. prefere this. An empty line appearing when I add a list is a bit unexpected.

AsterMiha:
Closing this PR since [this one](https://github.com/serlo/frontend/pull/2915) offers a more suitable solution to the same issue

PR #2867, opened 2023-09-12
add focus safe-guard for text-plugin

### LINKED ISSUE #2866: ensure 1000% that text plugin is focused
Testing confirms that rerendering of slate can cause `ReactEditor` to lose focus, even if it's still the active element.

This is the reason why calling `focus()` again is not working:

grafik

This bug surfaces if you want to jump into a table, e.g. by arrow down. Because of the complex structure, the focus is set too early, and there are some rerenderers afterwards. Resulting in a completely lost focus:

grafik

My attempt now is to build a safe-guard to refocus the editor after each rerender.

Entkenntnis:
changes are documented in the code

### DIFF
--- a/src/serlo-editor/editor-ui/editor-textarea.tsx
+++ b/src/serlo-editor/editor-ui/editor-textarea.tsx
@@ -24,6 +24,7 @@ export const EditorTextarea = forwardRef<
       {...props}
       ref={ref}
       data-qa={dataQa}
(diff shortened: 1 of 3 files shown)


Entkenntnis:
regressions:

- cursor jumping in latex visual mode: fixed
- e2e tests for suggestion-box fail (manual interaction works fine): fixed

PR #2868, opened 2023-09-12
fix(text-plugin): Prevent slate error when clicking on text plugin suggestion

### LINKED ISSUE #2869: Slate error when clicking option "Text" in plugin suggestion
In `src/serlo-editor/plugins/text/plugins/with-correct-void-behavior.ts` the next node is accessed using a path without checking if the path exists.

LarsTheGlidingSquirrel:
Fixes #2869


hejtful (approved):
Nice find and nice fix!

PR #2871, opened 2023-09-12
chore(fetcher): add sentry error when using legacy format

elbotho:
just in case… 
I'd let this run for a while and then just remove the legacy converter


Entkenntnis (approved):
hopefully no results :)

PR #2876, opened 2023-09-13
z-index hierarchie update

yanncloud:
new z-indexes

### DIFF
--- a/src/serlo-editor/plugins/rows/components/add-row-button-floating.tsx
+++ b/src/serlo-editor/plugins/rows/components/add-row-button-floating.tsx
@@ -25,7 +25,7 @@ export function AddRowButtonFloating({
           <button
             className={clsx(
               tw`
-                add-trigger z-[20] h-7 rounded-full py-2 text-gray-600 transition-all
+                add-trigger z-20 h-7 rounded-full py-2 text-gray-600 transition-all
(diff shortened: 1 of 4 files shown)


elbotho:
@yanncloud nice, that's really not a lot of changes, and the result seems to be pretty much equivalent 🎉

We should find a good way to document the system with a link to the Figma.
Not sure if a comment in the code or a wiki entry here is better, I'll think about it.

Only overlapp I found so far:
image
Do you want to try to solve this one? (Tooltip probably is in the same z-layer as the drag button)

yanncloud:
Ahh. Thanks you. I try top fix this.

src/serlo-editor/plugins/rows/components/add-row-button-floating.tsx
```diff
@@ -25,7 +25,7 @@ export function AddRowButtonFloating({
           <button
             className={clsx(
               tw`
-                add-trigger z-[70] h-7 rounded-full py-2 text-gray-600 transition-all
+                add-trigger z-[20] h-7 rounded-full py-2 text-gray-600 transition-all
```
elbotho: tiny thing: for 20 you can use the default
```suggestion
                add-trigger z-20 h-7 rounded-full py-2 text-gray-600 transition-all
```

elbotho (approved):

PR #2881, opened 2023-09-14
refactor(serloTable-plugin): simplify and improve

### LINKED ISSUE #2887: Confirming column removal removes serlo-table plugin focus
Steps to reproduce:
1. Focus a serlo-table plugin
2. Add a column in order to see the "remove column" button
3. Press the "remove column" button
4. A pop-up appears, where you need to confirm the removal action
5. Confirm the action
6. Plugin has lost focus

Since undo/redo works for the serlo-table plugin, do we really need the native browser confirmation pop-up for this case? Just removing the confirmation step seems like an easy solution here. @elbotho

elbotho:
- always use active table
- remove some plugin hacks
- use new colors
- add tooltips
- …

status:
- tab works perfectly now
- ~unfortunately up/down keys are not working any more~ (described in #2882)
- ~needed to add a timeout to regain focus after removing row (maybe we can get rid of it again)~

https://frontend-git-remove-table-plugin-hacks-serlo.vercel.app/entity/repository/add-revision/277232


src/serlo-editor/plugins/serlo-table/editor.tsx
```diff
@@ -249,18 +221,43 @@ export function SerloTableEditor(props: SerloTableProps) {
       const onRemove = () => {
         const empty = isRow ? isEmptyRow(rowIndex) : isEmptyCol(colIndex)
 
-        if (!empty && !window.confirm(confirmString)) return
+        if (!empty && !window.confirm(confirmString)) {
+          // setUpdateHack((count) => count + 1)
```
elbotho: did not spot a difference without it.

Entkenntnis (approved):
thank you for the nice rework

PR #2885, opened 2023-09-14
chore(migration): remove legacy converter

elbotho:
more steps as follow ups to keep the PRs managable


src/serlo-editor-integration/editor-response-to-state.ts
```diff
@@ -149,23 +150,15 @@ export function editorResponseToState(uuid: MainUuidType): DeserializeResult {
...
-      converted: !isEdtr(convertEditorState(content) || empty),
     }
 
     function getContent() {
-      const convertdContent = convertEditorState(content)
-
-      const convertedContent = toEdtr(convertdContent) as
-        | RowsPlugin
-        | OtherPlugin
+      const convertedContent = toEdtr(convertEditorState(content))
```
Entkenntnis: slightly confused what `convertEditorState` is doing: so it's a wrapper around JSON.parse? I would expect another name, e.g. parseState.

Entkenntnis (approved):
A bit of melancholy to see the legacy converter go :) There was quite a lot of bug fixing involved around this feature, but happy to see that we can finally move on.

src/serlo-editor-integration/editor-response-to-state.ts
elbotho: Good idea, I changed it in https://github.com/serlo/frontend/pull/2891

PR #2886, opened 2023-09-14
refactor(editor): use `onFocus` instead of `onMouseDown`

elbotho:
Works surprisingly well,I could not find problems so far (makes me suspicious :D).

(for https://github.com/serlo/backlog/issues/111)

---

follow up:
- try to remove `onMouseDown={(e) => e.stopPropagation()} // hack to stop edtr from stealing events` to see if it is still needed in those places

### DIFF
--- a/src/serlo-editor/core/sub-document/editor.tsx
+++ b/src/serlo-editor/core/sub-document/editor.tsx
@@ -54,7 +54,7 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
   }, [focused, plugin])
 
   const handleFocus = useCallback(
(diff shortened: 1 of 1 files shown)


hejtful (approved):

Entkenntnis:
I will do some additional testing here, because I noticed some quirks in FF

elbotho:
> I will do some additional testing here, because I noticed some quirks in FF

Can confirm. While this works pretty much flawlessly on chromium and safari it stops slate from activating (no cursor) in FF.

Entkenntnis:
work-in-progress solution in #2889 (it included this change)

PR #2889, opened 2023-09-15
Avoid rerendering hotkeys provider and simplify slate focus management

Entkenntnis:


Entkenntnis:
This PR currently brings more bugs than it solves. It's a difficult place to work with as described here: https://github.com/serlo/frontend/blob/staging/src/serlo-editor/plugins/text/README.md#challenges

PR #2890, opened 2023-09-15
feat(lenabi): reintroduce data wallet integration demo

hugotiburtino:


src/components/pages/data-wallet.tsx
```diff
@@ -0,0 +1,166 @@
...
+            Die Data Wallet ist auf deinem Smartphone, dort sind deine Daten
+            sicher aufgehoben. Um Serlo dein Vertrauen zu schenken, ist eine
+            einmalige Einrichtung erforderlich:
+          </p>
+          <ol className="serlo-ol mb-4 text-lg">
+            <li>
+              Lade dir die{' '}
+              <a
+                className="serlo-link font-bold"
+                href="https://enmeshed.eu/use/basics"
```
elbotho: url does not exist any more (domain sold).
Do we have a current link?

elbotho (changes requested):

src/components/pages/data-wallet.tsx
```diff
@@ -0,0 +1,166 @@
...
+                QR-Code
+              </a>
+              .
+              {qrCode &&
+                ((qrCode === 'loading' && <LoadingSpinner noText />) || (
+                  <img src={qrCode} />
+                ))}
+            </li>
+            <li>
+              Lege in deiner Enmeshed App ein neues Profil an und füge Serlo als
```
elbotho: Also here we should probably rename the app.

elbotho:
How can I test this? Do I need the local api setup or should it work on localhost with staging api already?

src/components/pages/data-wallet.tsx
hugotiburtino: I saw it.
It is strange, since it was working the last days. I think they will fix it soon.
Anyway, I will change it to .de (thanks @hejtful for the suggestion).

hugotiburtino:
Thanks for your review 😊 

> How can I test this? Do I need the local api setup or should it work on localhost with staging api already?

You can use staging api already.
After scanning the QR code you should get an error, we are in contact with the integration team of them to clear this problem

elbotho:
Do you want this to work in production as well? Or only for staging?

hugotiburtino:
For now it is only for staging. The enmeshed endpoint won't work in the api
anyways.

src/pages/wallet.tsx
elbotho, line 1: If you add a check for production here I'm happy to merge this :) 

(I can also do it)

elbotho (approved):

src/pages/wallet.tsx
hugotiburtino: Thanks for doing it ❤️

PR #2892, opened 2023-09-17
chore(editor): remove autofocusRef feature

elbotho:
there is like one case where this was actually helpful. 
I plan to replace this function with a different approach.

First step of new function in https://github.com/serlo/frontend/pull/2894.


hejtful (approved):

Entkenntnis (approved):
thankss for the clean up.

PR #2901, opened 2023-09-19
Prevent opening plugin suggestions in unwanted places

### LINKED ISSUE #2897: `/` opens suggestions in unwanted places (table, box title, ...)

LarsTheGlidingSquirrel:
Quick solution for fixing #2897 

Potential followup: 
Configuring what plugins are allowed where is done multiple ways: 
- `AllowedChildPlugins` context
- Plugin config prop
- Initial plugin state & config

Feels like we might be able to centralize this more.


elbotho:
@LarsTheGlidingSquirrel thanks for tackling this!
I think I'd prefere something like https://github.com/serlo/frontend/pull/2904 since we already have the info almost explicitly set with the `isInlineChildEditor` config. Any downside I forgot? :)

LarsTheGlidingSquirrel:
Closed in favor of #2904

PR #2904, opened 2023-09-19
fix(editor): another approach to prevent opening plugin suggestions in unwanted places

### LINKED ISSUE #2897: `/` opens suggestions in unwanted places (table, box title, ...)

elbotho:
(fixes https://github.com/serlo/frontend/issues/2897)


LarsTheGlidingSquirrel (approved):
Works great! However, I think we can improve the clarity. In both solutions. I'll collect my thoughts and maybe suggest some refactoring. 
For now, I am fine with merging this to get rid of the bug in staging. :+1:

PR #2915, opened 2023-09-21
fix(plugin-text): remove unnecessary double selection change

### LINKED ISSUE #2798: editor: Slate error when merging a text plugin with another text plugin that has a list

AsterMiha:
Fixes the [issue](https://github.com/serlo/frontend/issues/2798) that triggers an error when attempting to merge 2 text plugins with backspace when the first one starts with a list.
It turns out the problem was caused by an invalid selection after the 2 plugins merged.

@hejtful thanks again for the help!

### DIFF
--- a/src/serlo-editor/plugins/text/hooks/use-editable-key-down-handler.tsx
+++ b/src/serlo-editor/plugins/text/hooks/use-editable-key-down-handler.tsx
@@ -118,14 +118,20 @@ export const useEditableKeydownHandler = (
           const direction = isBackspaceAtStart ? 'previous' : 'next'
 
           // Merge plugins within Slate and get the merge value
-          const newValue = mergePlugins(direction, editor, id)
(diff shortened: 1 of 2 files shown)


elbotho (approved):
seems to work fine, thank you!

PR #2920, opened 2023-09-22
refactor(editor): unified renderer in separate routes

elbotho:
For https://github.com/serlo/frontend/issues/2933

### Demos/Tools
- [View with mockup data](https://frontend-git-unified-renderer-draft-serlo.vercel.app/static/show-revision)
- [View with mockup data, and revision info](https://frontend-git-unified-renderer-draft-serlo.vercel.app/static/show-revision?revision)
- [Live editor preview](https://frontend-git-unified-renderer-draft-serlo.vercel.app/___editor_preview)  (split panels)
- [/static/[...slug]](https://frontend-git-unified-renderer-draft-serlo.vercel.app/static/1565) to test actual content

### Bundle Size
**+ 25,1 KB** on `static/[...slug]` compared to `[...slug]`

### Overview
image

Right now using the same approach as we landed on for the plugin registry there is a function to register the static renderers before we run the StaticRender. 
This way integrations can choose to use custom static renderers as well, or wrap the default ones with extra code.

We have a default static renderer for each plugin that is inside the plugin folder but only uses code from the `renderer.tsx` and should otherwise be independent from the editor. 

Overall it's very similar to the current frontend-approach, but the code is moved into the editor and runs on serialized editor state directly with no custom types, conversions (no server-side-only steps at all).


src/components/content/static-entity.tsx
elbotho, line 1: (this is just duplicated for the `/static/…` route)

src/fetcher/static-create-exercises.ts
elbotho, line 1: (old commented code might still be needed to make exercise folders work)

src/fetcher/static-request-page.ts
elbotho, line 1: (this is just duplicated for the /static/… route with minor changes)

src/pages/show-revision.tsx
elbotho, line 1: (for demo purposes only)

src/pages/static/[...slug].tsx
Comment (elbotho) [inline src/pages/static/[...slug].tsx:1]: (this is just duplicated for the /static/… route with minor changes)

Entkenntnis (approved):

hejtful (approved):

PR #2922, opened 2023-09-25
fix(editor): first box fixes/improvements

elbotho:
(for https://trello.com/c/pIaQccPX/326-improve-box-plugin)

### [Preview](https://frontend-git-some-box-improvements-serlo.vercel.app/entity/repository/add-revision/277232)

- does not switch width any more
- less layout jumps (warning pos absolute, type menu similar height as empty state, quote margin fixed)
- low opacity when empty and not focused to indicate it will not be shown to the learners


Entkenntnis (approved):
low opacity is a good indicator

PR #2925, opened 2023-09-25
feat(editor): additional focus handler for tab nav

elbotho:
(for https://github.com/serlo/backlog/issues/111)

another go to fix focus + tab navigation 🤞

tested a bit in chrome, FF and safari and did not find new bugs so far.

Happy to get some help testing 😊

### DIFF
--- a/src/serlo-editor/core/sub-document/editor.tsx
+++ b/src/serlo-editor/core/sub-document/editor.tsx
@@ -52,11 +52,11 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
   // additional focus check to set focus when using tab navigation
   const handleFocus = useCallback(
     (e: React.FocusEvent) => {
-      // if after a short delay dom focus is not set inside focused plugin
-      // we overwrite it here (because it's probably because of tab navigation)
       if (!document) return
       if (['rows', 'exercise'].includes(document?.plugin)) return
 
+      // if after a short delay dom focus is not set inside focused plugin
+      // we overwrite it here (because it's probably because of tab navigation)
       const tabFocusTimeout = setTimeout(() => {
         // fixes a bug in table plugin with disappearing buttons
         if (e.target.nodeName?.toLowerCase() === 'button') return
--- a/src/serlo-editor/core/sub-document/editor.tsx
+++ b/src/serlo-editor/core/sub-document/editor.tsx
@@ -57,7 +57,7 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
 
       // if after a short delay dom focus is not set inside focused plugin
       // we overwrite it here (because it's probably because of tab navigation)
-      const tabFocusTimeout = setTimeout(() => {
+      setTimeout(() => {
         // fixes a bug in table plugin with disappearing buttons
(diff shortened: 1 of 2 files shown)


src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -48,6 +49,31 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
...
+
+        // find closest document
+        const target = (e.target as HTMLDivElement).closest('[data-document]')
+
+        if (!focused && target === containerRef.current) {
+          dispatch(focus(id))
+        }
+      }, 10)
+
+      return () => clearTimeout(tabFocusTimeout)
```
hejtful: I thought this was `useEffect`-specific syntax 😄 Does it work for event handlers in general, or just for `useCallback`?

src/serlo-editor/core/sub-document/editor.tsx
```diff
@@ -48,6 +49,31 @@ export function SubDocumentEditor({ id, pluginProps }: SubDocumentProps) {
     [focused, id, dispatch, document]
   )
 
+  // additional focus check to set focus when using tab navigation
+  const handleFocus = useCallback(
+    (e: React.FocusEvent) => {
+      // if after a short delay dom focus is not set inside focused plugin
+      // we overwrite it here (because it's probably because of tab navigation)
```
hejtful: Maybe this comment should go on line 53 or 60?

hejtful (approved):
Looks good, the tabbing is much nicer now.

It won't work when the selection is inside a list. I guess it's our decision if we want the list indentation or focus change in this case.

There is also this tiny issue with the plus button outline, but it's definitely not a blocker 😄 
Screenshot 2023-09-26 at 10 48 54

elbotho:
>It won't work when the selection is inside a list. I guess it's our decision if we want the list indentation or focus change in this case.

you mean that changing focus via tab does not work within a list? 
That would be okay for the scope of this PR :)

src/serlo-editor/core/sub-document/editor.tsx
elbotho: Good catch, I wrote it more out of habit. I don't think it works here.
I'm always a bit unsure how bad it would be to not clear a timeout like this.

Best and most complex:
Whenever a new focus event is triggered all old timeouts (on other plugins as well) get cleared (because we only want to focus the most current anyway)

In between:
I store the timer in a ref and clear it when unmounting the component.

Lazy:
We don't handle it at all.

Any votes? :)

hejtful: I rarely do `clearTimeout`s, to be honest. They are not recurring like `setEventHandler` or `clearInterval`, they happen once and clear themselves.

The users probably won't be changing the focus twice in 10 milliseconds. Maybe the only problematic behavior would be if the users hold Tab key? But even that works fine in this branch preview. So I vote for lazy, in this case. 😁

elbotho: I'll gladly oblige :D

Entkenntnis (approved):
Congrats! This actually works reasonable well! I haven't found bugs with it.

PR #2926, opened 2023-09-25
editor box plugin code review

elbotho:
For https://trello.com/c/pIaQccPX/326-improve-box-plugin we look at a single plugin a bit more closely and I thought we could do the same for the code / code style.

Imagine this is new code and just give it a proper review, maybe with some bonus questions:
- Is there something that annoys you / that you would like to see refactored?
- Is there something that is hard to understand (e.g. needs simplification / documentation)?
- It's quite a simple plugin, is there something that seems too complicated / that should be moved into the core for example?
- …

In the end the code does not have to be perfect at all, but something that you would not mind working on 😄

I asked all (FE-)devs  from the p_serlo-editor to review, do so only if you feel like it and have the time ✌️

### DIFF
--- a/src/serlo-editor/plugins/box4review/renderer.tsx
+++ b/src/serlo-editor/plugins/box4review/renderer.tsx
@@ -1,0 +1,97 @@
+import {
+  faExclamationTriangle,
+  faHandPointRight,
+  faLightbulb,
+  faMapSigns,
+  faQuoteRight,
+  faScroll,
+  faThumbtack,
+} from '@fortawesome/free-solid-svg-icons'
+import clsx from 'clsx'
+
+import { FaIcon } from '@/components/fa-icon'
+import { useInstanceData } from '@/contexts/instance-context'
+import { tw } from '@/helper/tw'
+
+export const boxTypeStyle = {
+  blank: {},
+  example: {},
+  quote: { icon: faQuoteRight },
+  approach: { icon: faMapSigns },
+  remember: { icon: faScroll },
+  attention: {
+    icon: faExclamationTriangle,
+    borderColorClass: 'border-red-100',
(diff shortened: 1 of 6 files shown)


src/serlo-editor/plugins/box4review/renderer.tsx
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const defaultStyle = {
+  icon: undefined,
+  borderColorClass: 'border-brand-300',
+  colorClass: 'text-brand',
+}
+
+export const types = Object.keys(boxTypeStyle)
+export type BoxType = keyof typeof boxTypeStyle
```
hejtful, line 40: Something like
```typescript
enum BoxType [
  blank,
  ...
]

interface BoxTypeStyle {
  icon: IconType
  borderColorClass: string
  colorClass: string
}

const boxTypeStyles: Record<BoxType, BoxTypeStyle>: { ... }

const types = Object.keys(boxTypeStyles)
```
should make the types more explicit and make the casting redundant in the rest of the PR.

src/serlo-editor/plugins/box4review/renderer.tsx
```diff
@@ -0,0 +1,97 @@
...
+        `,
+        borderColorClass
+      )}
+    >
+      {renderHeader()}
+      {boxType === 'quote' ? <blockquote>{children}</blockquote> : children}
+    </figure>
+  )
+
+  function renderHeader() {
```
hejtful, line 80: I'd personally make all of these `render*` functions either a separate component, or integrate them into the main JSX return.

hejtful (commented):
In general, I'm probably too immersed into this code by now, to be able to spot some obvious improvement areas. This is what came to mind.

PR #2928, opened 2023-09-27
feat(editor): simple focus after undo redo for some cases

elbotho:
low hanging fruits for https://github.com/serlo/backlog/issues/257

- It only sets focus to the plugin (not caring about slate cursor etc. but we handle that inside the text plugin anyway I think)
- Does not work when targeting the rows plugin (deleting/changing order) but most other cases work so it's an improvement

Todo: solve dependency cycle madness


elbotho:
found this introduces bugs, withdrawn…

PR #2929, opened 2023-09-27
feat(box): more icons, limited multimedia types in box

elbotho:
[demo](https://frontend-git-more-box-stuff-serlo.vercel.app/entity/repository/add-revision/277232)

### DIFF
--- a/src/serlo-editor/plugins/multimedia/editor.tsx
+++ b/src/serlo-editor/plugins/multimedia/editor.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
 
 import type { MultimediaProps } from '.'
 import { MultimediaRenderer } from './renderer'
@@ -13,6 +13,7 @@ import {
   selectAncestorPluginTypes,
   selectHasFocusedDescendant,
   selectIsFocused,
+  store,
   useAppSelector,
 } from '@/serlo-editor/store'
 import { EditorPluginType } from '@/serlo-editor-integration/types/editor-plugin-type'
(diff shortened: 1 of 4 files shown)


src/serlo-editor/plugins/multimedia/editor.tsx
```diff
@@ -29,6 +31,20 @@ export function MultimediaEditor(props: MultimediaProps) {
...
+  const hasBoxAnchestor = typesOfAncestors?.includes(EditorPluginType.Box)
+  const filteredPlugins = hasBoxAnchestor
+    ? config.allowedPlugins.filter(
+        (plugin) =>
+          ![EditorPluginType.Video, EditorPluginType.Geogebra].includes(
+            plugin as EditorPluginType
+          )
+      )
+    : config.allowedPlugins
+
```
elbotho: @hejtful what do you think, how terrible is this for performance? 😸 
I would memo it but I'm not sure how to recalculate it reliably if someone moves the plugin to a new parent.

hejtful: I guess it will impact performance, but not sure how much without profiling 😄 

We could reduce the impact by not using `useAppSelector`, and instead checking for `filteredPlugins` value when the modal opens. This way, the selector won't impact re-renders, and I think it's a safe trigger for re-evaluation of this value, because it can't change while the modal is open. I hope I managed to describe what I mean in an understandable way 😄 If not, I can try to rephrase.

elbotho: Hm yes, I think I understood. 
The setting will likely move into to toolbar at some point so I guess than it would not work any more. 
🤔

elbotho: Thanks for the help!

hejtful (approved):

PR #2932, opened 2023-09-29
unified renderer: refactor static h5p to include custom events

elbotho:
first time I used a special serlo component (for the `useEffect`) but because this uses dom events the other components remain unchanged 👌

(includes https://github.com/serlo/frontend/pull/2932)


Entkenntnis (approved):
thanks, code looks cleaner than before

PR #2948, opened 2023-10-06
More info in revision view

LarsTheGlidingSquirrel:
Shows additional info in revision view for plugins `anchor`, `injection` & `image`

Removes `isRevisionView` from `createRenderers`. Consuming context `RevisionViewContext` instead to display additional info. 

Preview revision view on `http://localhost:3000/static/show-revision?revision`


src/serlo-editor-integration/serlo-plugin-wrappers/anchor-serlo-static-renderer.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function AnchorSerloStaticRenderer(state: EditorAnchorPlugin) {
+  const isRevisionView = useContext(RevisionViewContext)
+
+  return (
+    <>
+      <AnchorStaticRenderer {...state} />
+      {isRevisionView ? (
+        <RevisionViewExtraInfo>{state.state}</RevisionViewExtraInfo>
+      ) : null}
```
elbotho: Idea:
If we would keep the logic inside one (ugly) component like it was before we could have one extra comp instead of one for each Renderer that only needs the special revision info.

So inside the create-renderers we could use something like
`(document:…) => <DefaultStaticRenderer {...document}/><RevisionExtra {...document} />` 

What do you think?

elbotho: (`RevisionExtra` would then use the `RevisionViewContext`. Namings are no suggestions ;) )

LarsTheGlidingSquirrel: Yeah, I also noticed we have lots of repetitions. I'll try to change it to something you suggested, hopefully also preventing repetitions in `createRenderer`.

elbotho: repetitions would be fine for me in createRenderer, just to make it less noisy an easy to understand.

LarsTheGlidingSquirrel: Hm, I tried to find a nice solution but `RevisionExtra` gets messy because it needs to handle different input types. Also, `createRenderer` gets messier. I think I prefer to look of having one extra component per plugin even though most of them just use the context and render the additional revision view info. 
@elbotho Maybe you have a better solution in mind?

elbotho: I don't love that `createRender` would be more messy but I would not mind an ugly `RevisionExtra`.
Might actually be nice to have all the revision stuff in one place.

LarsTheGlidingSquirrel: Changed it to new approach. Hope you like it! :relaxed:

elbotho (approved):
👌 I like it

PR #2956, opened 2023-10-09
ignore json-ld error in safari

elbotho:
(flying blind here, probably easiest to test in staging)


elbotho:
merged for testing
@Entkenntnis review still welcome :)

Entkenntnis:
It's not the cleanest solution, but solves the problem at hand ✨

Issue #2916, opened 2023-06-05
improvements such as focus, placeholder and default answers in input-exercise plugin


anbestCL:
Improvements for the input-exercise:

<img width="780" alt="grafik" src="https://github.com/serlo/backlog/assets/43368821/b62c7104-a8cb-4fdf-8aec-c4a6f5cbb4c8">

- [x] Put cursor in answer field instead of feedback field upon adding of new answer
- [x] First answer should/could be correct per default
- [x] #2976

(Moved from edtr-io: https://github.com/edtr-io/edtr-io/issues/293)

Entkenntnis:
last remaining task moved to new issue

PR #2957, opened 2023-10-09
feat(plugin-input-exercise): Mark the first answer as correct by default

AsterMiha:
By default, first answer to an input exercise will be marked as being correct, as suggested in [this issue](https://github.com/serlo/frontend/issues/2916)


elbotho (approved):
👍 thank you!

PR #2964, opened 2023-10-09
chore(layout-plugin): remove old layout plugin

elbotho:


Entkenntnis (approved):
let the clean-up begin

PR #2989, opened 2023-10-13
refactor(static): refactor types, merge requestPage

elbotho:
- uses new functions for generating getMetaDescription already.
- and refactor solution yet again


Entkenntnis (approved):
looks good, didn't notice anything problematic

PR #2990, opened 2023-10-14
refactor(static): use static data for author tools

elbotho:
based on https://github.com/serlo/frontend/pull/2989

some code refactoring as well

(probably the last step before we fully move to static renderer)

e.g. 
- http://localhost:3000/taxonomy/term/sort/entities/234582
- http://localhost:3000/taxonomy/term/move/batch/234582
- http://localhost:3000/entity/link/order/286557/link


Entkenntnis (approved):
works on dev

PR #3035, opened 2023-10-25
refactor(font): use variable karla instead of karmilla

### LINKED ISSUE #3034: chore(font): Karmilla -> Karla
The original Karla font has gotten a lot of updates now including the special characters we need so we can replace the fork (Karmilla) now.

elbotho:
- using karla instead of karmilla (closes #3034)
- using variable fonts for karla ([intro](https://fonts.google.com/knowledge/introducing_type/introducing_variable_fonts)) to save some kbs
- remove woff1 support ([see woff2 support](https://caniuse.com/woff2))
- fallback for older browsers


Entkenntnis (approved):
👍

PR #3038, opened 2023-10-25
refactor(editor): improve selector performance on keystroke

elbotho:
~this experiment just deactivates all selectors that:~
this replaces or refactors the slowest selectors that:
- are triggered on every keystroke
- use `getChildTree`

[long content](https://frontend-git-skipping-selectors-on-change-serlo.vercel.app/entity/repository/add-revision/287109) still is very sluggish (especially loading is not improved) but it's bearable now since typing kind of works.

for [#284](https://github.com/serlo/frontend/issues/3039)

### > [demo](https://frontend-git-skipping-selectors-on-change-serlo.vercel.app/entity/repository/add-revision/287109)


src/serlo-editor/plugins/rows/editor.tsx
```diff
@@ -2,13 +2,12 @@ import type { RowsProps } from '.'
...
 import { RowEditor } from './components/row-editor'
-import { selectParentPluginType, useAppSelector } from '@/serlo-editor/store'
+import { selectParentPluginType, store } from '@/serlo-editor/store'
 import { EditorPluginType } from '@/serlo-editor-integration/types/editor-plugin-type'
 
 export function RowsEditor({ state, config, id, editable }: RowsProps) {
-  const parentType = useAppSelector((state) =>
-    selectParentPluginType(state, id)
-  )
+  // this should not change so we don't need to use useAppSelector
```
hejtful: Can it not change on drag and drop?

elbotho: Thanks for the question.
If I'm not mistaken we only use it to check if the parent of the rows plugin is the template plugin or the article plugin (basically the main root element). And rows plugins can't be moved so we should be good.

Extended the comment and moved some code to make it a bit clearer.

Entkenntnis (approved):
I did some testing, the performance is way better now.

I haven't looked closely for regressions, but I didn't recognize any sofar.

PR #3041, opened 2023-10-26
refactor(text-plugin): replace MathSpan with StaticMath

elbotho:
(preloading math plugin does not seem to necessary any more but we should keep a look out for bugs relating to this)


src/serlo-editor/plugins/text/static-components/static-math.tsx
```diff
@@ -7,7 +7,10 @@ import 'katex/contrib/mhchem'
 import type { MathElement } from '../types/text-editor'
 import { sanitizeLatex } from '@/schema/sanitize-latex'
 
-export function StaticMath({ src, inline }: MathElement) {
+export type StaticMathProps = Omit<MathElement, 'children'>
+
+/** this is quite big 🐘, if you can load it dynamically  */
```
hejtful: Maybe it's my bad English, but at first I thought this means "if you can load this dynamically, then it is quite big". A comma would have probably saved me a couple of minutes, so like: "This is quite big. If you can, load it dynamically."

hejtful (approved):

src/serlo-editor/plugins/text/static-components/static-math.tsx
elbotho: Sorry for the confusion. I'll just add another elephant to make sure ;)

elbotho:
@Entkenntnis check out the bundle sizes in this one 😊
This is just because of loading the Math display dynamically. Let's hope it has no drawbacks.

Entkenntnis (approved):
preloading is still necessary for the fonts to load, otherwise the visuals are wrong:

grafik

we can actually move `import { KaTeXStyles } from '@/../external/katexstyles'` into the editor, this should do the trick, and also `import 'katex/contrib/mhchem'`

Entkenntnis (changes requested):
Didn't intended to approve ..

elbotho:
> preloading is still necessary for the fonts to load, otherwise the visuals are wrong:
>  we can actually move import { KaTeXStyles } from '@/../external/katexstyles' into the editor, this should do the trick, and also import 'katex/contrib/mhchem'

ah, would have been too easy :) 
you mean additionally right? (like in [9a2c48b](https://github.com/serlo/frontend/pull/3041/commits/9a2c48b89ba0a1f10f62ebde7ca6bc472bb99102))

Entkenntnis (approved):
it's working now

PR #3052, opened 2023-10-31
feat(exercise-plugin): only show loaded exercise plugins

elbotho:
This implementations can decide which plugins they want to support by adding or omitting them from `createPlugins`

### DIFF
--- a/src/serlo-editor/plugins/exercise/editor.tsx
+++ b/src/serlo-editor/plugins/exercise/editor.tsx
@@ -36,7 +36,7 @@ export function ExerciseEditor({ editable, state }: ExerciseProps) {
           {interactive.render()}
         </>
       ) : editable ? (
-        <>
+        <div className="mx-side">
           <p className="mb-2 text-gray-400">
             {exStrings.addOptionalInteractiveEx}
           </p>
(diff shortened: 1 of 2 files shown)


src/serlo-editor/plugins/exercise/editor.tsx
```diff
@@ -18,6 +19,11 @@ const interactiveExerciseTypes = [
 export function ExerciseEditor({ editable, state }: ExerciseProps) {
   const { content, interactive } = state
 
+  // only show supported interactive exercise types
+  const interactiveExerciseTypes = allInteractiveExerciseTypes.filter(
+    (type) => !!editorPlugins.getByTypeOrUndefined(type)
+  )
```
LarsTheGlidingSquirrel: Alternative way that came to my mind. Just a suggestion. :) 
```
  const interactiveExerciseTypes = allInteractiveExerciseTypes.filter(
    (type) => editorPlugins.getAllWithData().some(plugin => plugin.type === type)
  )
```

elbotho: I wanted to leave the complexity in  [editor-plugins.tsx](https://github.com/serlo/frontend/pull/3052/commits/54797c53949b220739ddc09e0aad4b4c99d5d03e#diff-9484f1fe9e5182c0953a83cac817f2e0445f62139bfb41e54a03824045a540b4) but this is a lot nicer. Thanks!

LarsTheGlidingSquirrel (approved):
Looks good! :)

PR #3057, opened 2023-11-02
filter trashed grouped exercises

### LINKED ISSUE #3056: Trashed grouped exercise is still visible
Example: https://de.serlo.org/mathe/273210/273210

d) is trashed but still shown.

Entkenntnis:


elbotho (approved):
thank you!

PR #3084, opened 2023-11-09
refactor(editor): remove editable context

elbotho:
since edusharing now uses the static renderer as well we can simplify the editor code a bit


Entkenntnis (approved):
looks like a great simplification of the code

PR #3087, opened 2023-11-10
refactor(licenses): use local license data instead of API

elbotho:


Entkenntnis (approved):
PR looks good, thanks for refactoring this.

We only have to take a look at the size of instanceData and how much it's growing in the future (as of today it has no noticable impact on performance). Maybe some data can be moved to loggedInData (agreements only relevant for editor?)

elbotho:
> We only have to take a look at the size of instanceData and how much it's growing in the future (as of today it has no noticable impact on performance). Maybe some data can be moved to loggedInData (agreements only relevant for editor?)

I took care to separate the content (that is quite a lot of data) from the rest. Content is only used on server now.
The agreements could move as well but it seemed to be too much extra complexity for a few strings.

PR #3088, opened 2023-11-10
feat(pages): move static cf pages to frontend

### LINKED ISSUE #288: Move legal pages to frontend
Currently our legal pages are rendered by the CF-Worker. These are the sites:

* https://de.serlo.org/imprint
* https://de.serlo.org/privacy
* https://de.serlo.org/terms

It would be great when we move those sites to the frontend. The rendering is done by the module https://github.com/serlo/serlo.org-cloudflare-worker/tree/main/src/legal-pages Refactoring of the source code there would be also nice :-)

elbotho:
- [404 for meeting redirect](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/___meeting_not_found)
- Impressum ([DE](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/legal),[EN](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/en/legal))
- Terms ([DE](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/terms),[EN](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/en/terms))
- Privacy ([DE](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/privacy),[EN](https://frontend-git-cf-worker-static-stuff-serlo.vercel.app/en/privacy))

This is a bit quick and dirty, but since we almost never change it ...

I simplified the privacy part a bit.
It now fetches the commit date from the GitHub-API and the archive just links to the markdown files on GitHub(good enough, no?) so we don't have to manually change a list


src/pages/___meeting_not_found.tsx
```diff
@@ -0,0 +1,23 @@
...
+      noHeaderFooter
+      noContainers
+      showNav={false}
+      authorization={{}}
+    >
+      <Head>
+        <meta name="robots" content="noindex" />
+      </Head>
+      <h1 className="serlo-h1 mt-12 p-2 text-center">
+        Das Meeting gibt&#39;s leider nicht. 😶‍🌫️
```
kulla: Side note: A generic `not_found` page would be enough for the CF worker.

src/pages/legal.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const getStaticProps: GetStaticProps<LegalData> = async (context) => {
+  const isGerman = context.locale === 'de'
+  const url = `${legalRepo}/${isGerman ? 'de' : 'en'}/imprint.md`
+
+  try {
+    const response = await fetch(url)
+    const markdown = await response.text()
+    const contentHtml = micromark(markdown, {
+      allowDangerousHtml: true,
```
kulla: Do we actually need this setting?

src/pages/privacy.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <br />
+        {isGerman
+          ? 'Frühere Versionen findest Du im'
+          : 'You can view past versions in the'}{' '}
+        <Link
+          href={`https://github.com/serlo/serlo.org-legal/tree/main/${
+            isGerman ? 'de' : 'en'
+          }/privacy`}
+        >
+          {isGerman ? 'Archiv' : 'archive'}
```
kulla, line 38: Shall we add this to our `i18n`?

src/pages/privacy.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const contentHtml = micromark(markdown, {
+      allowDangerousHtml: true,
+    })
+
+    const lastUpdateResponse = await fetch(
+      `https://api.github.com/repos/serlo/serlo.org-legal/commits?path=%2F${instance}%2Fprivacy%2Fcurrent.md&page=1&per_page=1`
+    )
+    const json = (await lastUpdateResponse.json()) as [
+      { commit: { committer: { date: string } } },
+    ]
```
kulla, line 88: Here you can use `io-ts` to make it more secure...

src/pages/privacy.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const isGerman = context.locale === 'de'
+  const instance = isGerman ? 'de' : 'en'
+  const url = `${legalRepo}/${instance}/privacy/current.md`
+
+  try {
+    const response = await fetch(url)
+    const markdown = await response.text()
+    const contentHtml = micromark(markdown, {
+      allowDangerousHtml: true,
+    })
```
kulla: This seems to be duplicated code which can be stored in a utils function

kulla (approved):
Thank you. Much better and simpler than the current implementation! :heart: I have added some notes. However nothing which blocks merging the PR. Only the `allowDangerousHtml` can be checked again before merging whether this is really needed.

src/pages/privacy.tsx
elbotho: I opted to leave it out for now since we seldomly change it and we'll probably only have `de` and `en` versions. But not a strong opinion here.

src/pages/legal.tsx
elbotho: input markdown example:
```
Serlo Education e. V.<br>
IBAN: DE98 4306 0967 8204 5906 00<br>
BIC: GENODEM1GLS (GLS Bank)

Knr: 8204590600<br>
BLZ: 43060967
```

If we don't use the setting we end up with `<br>` in the output.
Options:
- make sure we never use any html in the markdown files
- add the [tagfilter plugin](https://github.com/micromark/micromark-extension-gfm-tagfilter)
- leave it as it is since the markdown is not user generated and the [source repo](https://github.com/serlo/serlo.org-legal/blob/main/de/imprint.md?plain=1) is as secure (or more so) as this repo

kulla: Or we add a replace all for `&lt;br&gt;` to `<br>`...

src/pages/privacy.tsx
elbotho: secure in what way?
If it's **not** the expected result the frontend will not build, so I'd say it's pretty secure ;)

src/pages/legal.tsx
elbotho: nice idea

src/pages/privacy.tsx
elbotho: now with the manual html conversion I agree it's worth it :)

src/pages/___meeting_not_found.tsx
elbotho: okay. I made it more general and added a status code 404. 
After we de https://github.com/serlo/backlog/issues/287 we could actually just use the default 404 as well.

elbotho:
@kulla thank you for you feedback, implemented most of it and did some other improvements.
Maybe you could give it another quick review.

- https://github.com/serlo/frontend-e2e-tests/pull/79
- https://github.com/serlo/cloudflare-worker/pull/590

kulla (approved):
Thank you! :heart:

src/pages/privacy.tsx
kulla: > If it's not the expected result the frontend will not build, so I'd say it's pretty secure ;)

Secure in a sense that not an `undefined` error is thrown but the can return a good error code (and have a sentry reporting)

elbotho: > and have a sentry reporting

frontend would not build and report an error building (including the path, e.g. `/legal`) so we do not need sentry here imho.

(these are completely static sites, they will not be revalidated after build)

kulla:
Merging it, so that https://github.com/serlo/cloudflare-worker/pull/590 can be merged and tested...

PR #3092, opened 2023-11-13
exercise generation: Get rid of redundant type, fix license and ids

CodingDive:


src/components/content/exercises/exercise-preview-page.tsx
```diff
@@ -39,6 +41,8 @@ export const ExercisePreviewPage: React.FC<ExercisePreviewPageProps> = ({
   const entityId = useEntityId()
   editorRenderers.init(createRenderers())
 
+  const { licenses } = useInstanceData()
+  const license = getLicense(licenses, 1)
```
elbotho: `getDefaultLicense` would be a bit clearer imho.

src/helper/ai-generated-exercises/data-conversion.ts
```diff
@@ -303,13 +305,14 @@ function createInteractive(exercise: ExpectedExerciseTypes): Interactive {
...
-  exercise: ExpectedExerciseTypes
+  exercise: ExpectedExerciseTypes,
+  licenseId: number
 ): EditorSolutionDocument {
   return {
     plugin: EditorPluginType.Solution,
     state: {
       prerequisite: undefined,
-      licenseId: 1,
+      licenseId,
```
elbotho: sorry did not spot this earlier: 
We can further simplify since solution always defaults to the license of the exercise now, so we don't have to provide it at all.

elbotho (changes requested):
looks good already, requested two small changes

src/helper/ai-generated-exercises/data-conversion.ts
CodingDive: Somehow the types forced me to pass `undefined` here instead of omitting it entirely

PR #3095, opened 2023-11-13
Fix close button position in modal

### LINKED ISSUE #3094: regression: close button placed incorrectly
changes [in this PR](https://github.com/serlo/frontend/pull/3000/) broke the other modals with close buttons:

e.g:
image

CodingDive:
I did not check all the modals (there are a lot). But I added back the absolute position.

Fixes #3094

### DIFF
--- a/src/components/modal-with-close-button.tsx
+++ b/src/components/modal-with-close-button.tsx
@@ -78,7 +78,7 @@ export function ModalWithCloseButton({
             onClick={onRequestClose}
             title={strings.share.close}
             dataQa="modal-close-button"
-            className={closeButtonClassName}
+            className={clsx('absolute right-2 top-2', closeButtonClassName)}
           />
         </>
       )}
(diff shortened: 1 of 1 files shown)


src/components/modal-with-close-button.tsx
```diff
@@ -78,7 +78,7 @@ export function ModalWithCloseButton({
             onClick={onRequestClose}
             title={strings.share.close}
             dataQa="modal-close-button"
-            className={closeButtonClassName}
+            className={clsx('absolute right-2 top-2', closeButtonClassName)}
```
elbotho:
```suggestion
            className={clsx('absolute right-3.5 top-3.5', closeButtonClassName)}
```
that was the old value

elbotho (changes requested):

elbotho (approved):
thanks!

PR #3096, opened 2023-11-13
fix(plugin-equations): replace MathRenderer with StaticMath

### LINKED ISSUE #3091: latex styling is wrong in editor
reproduction: https://de.serlo.org/entity/repository/add-revision/291361/291362

Expected: Formula is displayed once

Actual: Formula is duplicated

grafik

Rendered output is working: https://de.serlo.org/mathe/291361/291361

elbotho:
closes #3091

~since this currently removes `data-qa="plugin-math-renderer"` this will probably break some e2e test. 
Had no time to investigate yet.~
could not find `plugin-math-renderer` in in e2e repo so we are probably fine.

This just uses StaticMath as well inside the editor for preview and while the latex-mode is active.
This should make preview more reliable. 
Not sure how much performance impact [this will have](https://github.com/serlo/frontend/pull/3096/files#diff-5cd34eeae532ae77d611bf33ad11c481b4a55b88789bbdaa5015b9b0e6b27a79R50-R52), but I think it's okay for now.


src/serlo-editor/plugins/equations/editor/editor.tsx
```diff
@@ -50,6 +47,10 @@ export function EquationsEditor(props: EquationsProps) {
     ) ||
     focusedElement === state.firstExplanation.id
 
+  const staticDocument = useAppSelector((storeState) =>
+    selectStaticDocument(storeState, id)
+  )
```
hejtful, line 52: `selectStaticDocument` only uses the `documents` slice and is cached by the `id`, so this really shouldn't affect performance at all. 👍

hejtful (approved):
Very nice, I had no idea this was possible.

Entkenntnis (approved):
Thank you, works!

PR #3104, opened 2023-11-15
add event error message (for staging & reviewers)

### LINKED ISSUE #3086: Show warning when event of a notification is null
See https://trello.com/c/zSfG9Twd/372-null-events-shall-not-break-notifications: I changed the api that the event object of a notification can be null. We need to update frontend that it shows a warning when an event is null. 

Blocker: https://github.com/serlo/api.serlo.org/pull/1105 need to be deployed in `staging`

elbotho:
only displaying in staging and for reviewers now so we don't annoy regular users.
image


Entkenntnis (approved):
thank you

PR #3113, opened 2023-11-16
Modify editor state directly & implement feedback of Botho

CodingDive:
* Remove sessionStorage logic
* Remove subject from Wizard and always infer it from taxonomy
* Improve error boundary
* Add hover states
* Solution with 1 number => Input exercise
* Remove Ai Wizard context
* Rename IEditorExerciseData => ExercisePreviewFromAi
* Reintroduce material ui svg (sparkles) - has Apache License


src/components/content/exercises/exercise-generation-wizard/topic.tsx
```diff
@@ -109,3 +109,16 @@ export const Topic: React.FC<TopicProps> = ({
...
+export function extractTopicFromTitle(
+  title: string | undefined
+): string | null {
+  if (!title) {
+    return null
+  }
+
+  const match = title.match(/Aufgaben zu(?:m|r)? (.+)/)
+  return match ? match[1].trim() : null
+}
```
elbotho: Just a suggestion:
Should this actually only return if the string starts with `Aufgaben zu…`?
If yes, maybe simplify the regex to something like `Aufgaben zu. (.+)`

Otherwise this should suffice:
``` ts
export function extractTopicFromTitle(title?: string){
  return title?.replace(/^Aufgaben zu./,'')
}
```

src/helper/error-boundary.tsx
```diff
@@ -14,17 +17,29 @@ export class ErrorBoundary extends Component<Props, State> {
...
+    this.setState({ errorInfo })
   }
 
   render(): ReactNode {
     if (this.state.hasError) {
-      return <h1>Something went wrong.</h1>
+      return (
+        <div>
+          <h1>{this.props.somethingWentWrongString}</h1>
+          {process.env.NODE_ENV === 'development' && (
```
elbotho: I'd prefer `!isProduction` as a check.

src/helper/error-boundary.tsx
```diff
@@ -14,17 +17,29 @@ export class ErrorBoundary extends Component<Props, State> {
...
+            <details style={{ whiteSpace: 'pre-wrap' }}>
```
elbotho:
```suggestion
            <details className="whitespace-pre-wrap">
```

src/pages/entity/create/[type]/[taxonomyId].tsx
```diff
@@ -1,7 +1,8 @@
 import request, { gql } from 'graphql-request'
 import { GetStaticPaths, GetStaticProps } from 'next'
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 
+import { LazyLoadExerciseGenerationWrapperOrNull } from '../../../../components/content/exercises/lazy-load-exercise-generation-wrapper-or-null'
```
Comment (elbotho) [inline src/pages/entity/create/[type]/[taxonomyId].tsx]:
```suggestion
import { LazyLoadExerciseGenerationWrapperOrNull } from '@/components/content/exercises/lazy-load-exercise-generation-wrapper-or-null'
```

src/serlo-editor/plugins/serlo-template-plugins/exercise-group/text-exercise-group.tsx
```diff
@@ -76,6 +78,7 @@ function TextExerciseGroupTypeEditor(
   return (
     <>
       <div className="absolute right-0 -mt-20 mr-side flex">
+        {canUseAiFeaturesOutsideProduction && <AiExerciseGenerationButton />}
```
elbotho: works fine here but I personally stopped using `&&` in jsx in general to avoid bugs.
(reasoning e.g. [this article](https://www.crocoder.dev/blog/react-conditional-rendering/))

elbotho (changes requested):
thanks for implementing the suggestions.
this review is a bit nitpicky, feel free to ignore stuff.

But I found at least one bug.

elbotho:
Suggestion: Now that the modal is bigger overall (I like) you could consider simplifying UI and code by not switching to a full screen version when generating/previewing.

elbotho:
Bug: I can currently insert a `ExerciseGroup` on `/entity/create/Exercise/` and they display fine in the editor.
But after saving only the task displays, not the exercises (which is expected because the saving mutation does not expect another content type). This probably is true for `Exercise` on `/entity/create/ExerciseGroup` but I did not test it.

CodingDive:
> Suggestion: Now that the modal is bigger overall (I like) you could consider simplifying UI and code by not switching to a full screen version when generating/previewing.

Will sync with Gregor & authors about this. I personally like the suggestion. If there was only one modal to be rendered, it could simplify quite a lot of code indeed. Right now, we also get this rather annoying warning `React-Modal: Cannot register modal instance that's already open`. Which I saw in some issues of react-modal seems to occur when conditionally rendering the modal (the unmounting of portals seems a little bit shaky).

src/components/user-tools/foldout-author-menus/author-tools.tsx
```diff
@@ -349,12 +334,36 @@ export function AuthorTools({ tools, entityId, data }: AuthorToolsProps) {
...
+                  loggedInStrings.ai.exerciseGeneration.buttonTitleSingular
+                }
+                href={`/entity/create/${UuidType.Exercise}/${data.id}?showAiWizard=`}
+                icon={faWandSparkles}
+              />
+              <SubItem
+                key="ai-group-exercise"
+                title={loggedInStrings.ai.exerciseGeneration.buttonTitle}
+                href={`/entity/create/${entityType}/${data.id}?showAiWizard=`}
+                icon={faWandMagicSparkles}
```
elbotho, line 362: Other icon on purpose?

src/data/en/index.ts
```diff
@@ -1179,13 +1174,16 @@ export const loggedInData = {
...
+          subtasksTitleExerciseGroup: 'How many subtasks should there be?',
           noSubtasks: 'No',
           yesSubtasks: 'Yes',
+          subtasksLabel: 'Exercises',
           numberOfSubtasksPlaceholder: 'Number of subtasks',
           chooseOption: "Choose an option",
           multipleChoice: "Multiple Choice",
           singleChoice: "Single Choice",
-          solutionWithOneNumber: "Solution with 1 number"
+          inputExercise: "Input exercise",
```
elbotho, line 1185: Those strings already exist in the editor strings so we could avoid a duplication here.

elbotho (approved):

src/components/user-tools/foldout-author-menus/author-tools.tsx
CodingDive: Yeah, I figured to differentiate it but I'm not too sure if it's a good or bad idea :man_shrugging: 
Need to change the i18n strings once it's merged

src/data/en/index.ts
CodingDive: Input exercise exists indeed. "Multiple Choice" and "Single Choice" do not. I could remove the input exercise but for sake of consistency and case sensitivity ("Input Exercise" exists), maybe it's better to keep it here too

elbotho: they do… search for "Single-choice" and "Multiple-choice"

src/components/user-tools/foldout-author-menus/author-tools.tsx
elbotho: Imho they look too similar to make sense to the users.

PR #3114, opened 2023-11-21
refactor(exercise-generation): simplify, move files, minor improvements

elbotho:
It's probably best to go through commit by commit since I moved a bunch of files.


CodingDive (approved):
Thank you a lot! Greatly simplifies some parts of the code. Really happy that the scroll ref is not needed anymore :tada:

PR #3115, opened 2023-11-21
Add option to configure allowedPlugins in solution

LarsTheGlidingSquirrel:
Adds option to configure what plugins are allowed within a solution plugin. Useful for edu-sharing and other future editor integrations. 

TODOs:
- [x] Currently, all plugins are allowed per default. But for example having an exercise in the solution of an exercise makes little sense I think. Should we make the default more restrictive?
- [x] Is this still needed?


elbotho:
Are there other (special) restrictions you are planning?
Because right now I think we probably want the same restrictions for Serlo probably…
We might even think about not allowing the exercises in general and only allowing them in certain places.
They probably don't make sense in the description of an applet for example as well.

LarsTheGlidingSquirrel:
> Are there other (special) restrictions you are planning? Because right now I think we probably want the same restrictions for Serlo probably… We might even think about not allowing the exercises in general and only allowing them in certain places. They probably don't make sense in the description of an applet for example as well.

In the RLP integration I want to allow the edusharing plugin but disallow the exercise plugin for example. So, I feel like we need to be able to set `allowedPlugins`.
I agree, exercises should probably be disallowed in other places as well. Maybe this could be a good starting point to try to simplify the configuration of what plugins are allowed where. I had some ideas in the past concerning that.

LarsTheGlidingSquirrel:
Not needed any more!

PR #3116, opened 2023-11-21
Hide solution license and prerequisite in RLP intergration

LarsTheGlidingSquirrel:
- There are no solution licenses in the RLP integration. -> Hide the button to add a license.
- Solution prerequisite lets you link other serlo.org content by id. This does not work in the RLP integration. -> Hide it there.


elbotho (approved):
👍

PR #3120, opened 2023-11-22
Remove global JSX styles from modal

CodingDive:
Please check the save-modal in the review. I think the responsiveness is a little bit nicer now too. If you like the before better (instead of a fat button, we can just add flex-wrap).

## Before

image

## After

image


src/serlo-editor/plugins/serlo-template-plugins/helpers/content-loaders/revision-history-loader.tsx
```diff
@@ -75,32 +75,17 @@ export function RevisionHistoryLoader<T>({
...
+        className="top-[3rem] max-h-[80vh] w-[900px] max-w-[90vw] translate-y-0 overflow-y-auto pt-0"
       >
-        {/* // a bit untidy, but it's very nice to reuse this component here */}
+        {/* a bit untidy, but it's very nice to reuse this component here */}
         <SerloRevisionHistory
           data={revisionsResponse.data?.uuid}
           hideEdit
           onSelectRevision={onSelectRevision}
           selectedRevisionId={currentRevision}
+          tableClassName="mt-12 !w-auto"
```
CodingDive, line 86: Need the !important here because of the custom value `w-[500px]` inside the ModalWithCloseButton component

elbotho:
weird positioning for me (overlaps bottom of window)
image
image

elbotho:
- can we have a maxwidth for my tv? ;)
- revision modal still broken
image

CodingDive:
I do prefer the simplicity of
https://github.com/serlo/frontend/pull/3121

elbotho (changes requested):
problems with settings modal

CodingDive:
> problems with settings moda

Yeah, as the global styles no longer get applied, a lot of the usages will be broken.

Toolbar settings of 
Multimedia
Image 
Audio
...

Entkenntnis:
@CodingDive Thank you for tackling this! As you have noticed, we are defining the styles in three different places and it's a bit difficult to untangle. It would be very helpful to sort this out properly once and for all and remove the global styles.

Entkenntnis:
@CodingDive can you give an update if the issues botho has noted are solved in this PR?

src/components/modal-with-close-button.tsx
```diff
@@ -19,10 +19,9 @@ interface ModalWithCloseButtonProps {
   title?: string
   onCloseClick: () => void
   children: ReactNode
-  className?: string
+  className?: string | ReactModal.Classes
```
elbotho:
```suggestion
  className?: string
```
Not needed, right?

CodingDive:
The modal of the settings should look good now, however, I just saw that the `lightbox` component is broken 

image

Edit: Should work now, I don't like the hack though

elbotho (approved):

elbotho (approved):

PR #3121, opened 2023-11-22
Experiment with tailwind merge

CodingDive:
Removes quite a few !importants. If we lazy load the modals, the code splitting code would most likely remove the extra 5-6kbs from the bundle. https://github.com/dcastil/tailwind-merge/blob/v2.0.0/docs/when-and-how-to-use-it.md#increases-bundle-size

Do you think we should add it?


src/serlo-editor-integration/components/save-modal.tsx
```diff
@@ -83,7 +82,7 @@ export function SaveModal({
       title={editorStrings.save}
       className={clsx(
         !isOnlyText &&
-          '!top-[2rem] max-h-full w-auto !-translate-x-1/2 translate-y-0 overflow-y-auto pb-20'
+          'top-[2rem] max-h-full w-[900px] -translate-x-1/2 translate-y-0 overflow-y-auto pb-20'
       )}
```
CodingDive: @elbotho How is the modal supposed to look like when isOnlyText is true? As we are now not applying any global styles, I'm sure it'll look odd and will probably be missing some styles.

elbotho: it looks fine. you can see it when editing you profile description for example.

CodingDive:
Closing this in favor of #3125 and #3120 (which did not need tailwind-merge after all)

PR #3122, opened 2023-11-22
Fix: Render ai generate exercise button only for new exercises

CodingDive:
I guess it doesn't make any sense to show this button when editing an existing exercise, does it? We'd have to add a warning that it overwrites the existing exercise, similarly what we do when importing a different entity.

image


elbotho (approved):
since you are inside the editor here you could also check if the state is empty but this should work.

PR #3130, opened 2023-11-24
feat(rows-plugin): restrict where exercises can be used

elbotho:
root plugin:
- serlo: only show in articles and coursePages templates for now
- every else: show everywhere

parent plugins:
- always hide in solutions


Entkenntnis:
exercises within solutions would be ... inception 🤯 good that we prevent this

elbotho:
> exercises within solutions would be ... inception 🤯 good that we prevent this

Lars catched this one :) 

although I personally still like the injection loop best ♾️  :D

LarsTheGlidingSquirrel (approved):
Nice, quick solution! 
I still feel the urge to simplify this but we can keep it like this for now. I'll collect my thoughts about this and will share them with you soon.

PR #3131, opened 2023-11-27
feat(ci): build without lint & tsx

elbotho:
since we run eslint and tsc in other actions this should not make a difference, right?
(aside from speed ⚡)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -87,6 +87,9 @@ const nextConfig = {
   eslint: {
     ignoreDuringBuilds: true,
   },
+  typescript: {
+    ignoreBuildErrors: true,
+  },
 }
 
 const withBundleAnalyzer = bundleAnalyzer({
--- a/next.config.js
+++ b/next.config.js
@@ -84,6 +84,9 @@ const nextConfig = {
       skipDefaultConversion: true,
     },
   },
+  eslint: {
+    ignoreDuringBuilds: true,
+  },
 }
 
 const withBundleAnalyzer = bundleAnalyzer({


Entkenntnis:
Generally you don't want to deploy if there are type errors.

I tried this in another project and the dev experience was terrible - a lot of white screens of death in preview branches, no chance to debug, etc..

Vercel documentation is also quite clear 🙈 

grafik

Entkenntnis (changes requested):
See coment above, I would not recommand this, at least not the `tsc` part.

`eslint` is less critical, but I don't know if the slight speed increase is worth the potential frustration.

elbotho:
> See coment above, I would not recommand this, at least not the tsc part.

> eslint is less critical, but I don't know if the slight speed increase is worth the potential frustration.

As it suggests in the screenshot we are running the checks anyway in the other actions, right?

If I push something that has failing tsc or eslint I get informed by the GitHub actions. The preview might build (incorrectly) but it should be obvious to everybody that something went wrong. 
Most people run `lint` locally anyway, so in most cases there should not be a difference.

So if I'm not missing something (completely possible) the only annoying case I can think of is something like:
- there is a preview for a branch already that a dev sent to other people
- the dev updates the branch and do not run the linter locally and do not check the CI results
- other people try to access the preview and get a potentially broken build

~If we definitely want to avoid that we could probably use something like https://github.com/marketplace/actions/wait-on-check but yeah, I'm not sure if that's worth the effort.~

Entkenntnis:
I would say that our own CI checks are superfluous - I would rather remove the github actions checks than to build without checking. It's not about production or merging, it's about preview branches.

> The preview might build (incorrectly) but it should be obvious to everybody that something went wrong.

This is not obvious to the person who is using the preview branch. Right now, if you push a faulty commit, you still get a working preview branch of the last working state.

The same if you do a faulty quickfix in staging or production. The build fails, and then you have time to fix it.

If you allow errors to be deployed, you have two nasty failure cases:
- Generally, if your types are not working, something is wrong and will likely crash your app, but without nice dev error messages.
- Worse: the previously **working** deployment is replaced! This will put you under immediate time pressure.

I'm sure that Vercel put a lot of thought into there recommandation. I was surprised as well that they are forcing you to lint and typecheck, but I came to the conclusion that this is really helpful best practice. (they went on rewriting the compiler in rust to improve build time! but never disabled type checking 😅 I think we shouldn't be too clever here and over-optimize)

Entkenntnis:
I'm closing this because (this is one of the few places where) I prefer safety over speed and think the double checking is worth it.

PR #3134, opened 2023-11-27
Improve quality of generated multiple choice exercises and add strategy

CodingDive:
Also disable feature for non German locale


src/components/exercise-generation/use-ai-features.tsx
```diff
@@ -1,11 +1,14 @@
 import { Ai } from '@serlo/authorization'
+import { useRouter } from 'next/router'
 
 import { useCanDo } from '@/auth/use-can-do'
 import { isProduction } from '@/helper/is-production'
 
 export function useAiFeatures() {
   const canDo = useCanDo()
-  const canUseAiFeatures = canDo(Ai.executePrompt)
+  const { locale } = useRouter()
```
elbotho: just a note:
if you use this you get types:
```suggestion
  const { lang } = useInstanceData()
```

src/components/exercise-generation/exercise-generation-wizard/difficulty.tsx
```diff
@@ -76,3 +77,16 @@ export function Difficulty({
...
+  difficulty: ExerciseGenerationDifficulty
+): string {
+  const mapping: { [key in ExerciseGenerationDifficulty]: string } = {
+    low: 'einfachen',
+    medium: 'typischen',
+    high: 'komplexen',
+  }
+
+  return mapping[difficulty]
+}
```
elbotho: note:
this seems a bit complicated, maybe something like this is enough (and you could probably remove the type definition above as well)?
``` ts
const difficultyToDescriptor = {
  low: 'einfachen',
  medium: 'typischen',
  high: 'komplexen',
} as const 
```
and accessing it with `difficultyToDescriptor[difficulty]` directly.

elbotho (approved):
just some notes

PR #3142, opened 2023-11-29
User Journey with Data Wallet

hugotiburtino:


src/components/content/entity.tsx
```diff
@@ -59,7 +63,11 @@ export function Entity({ data }: EntityProps) {
       {renderStyledH1()}
       {renderUserTools({ aboveContent: true })}
       <div className="min-h-1/4" key={data.id}>
-        {data.content && renderContent(data.content)}
+        {isMockupCoursePage ? (
```
hugotiburtino: It would be great not having to change this file just for a mock, any ideas?

elbotho: If I remember correctly last time we did not merge it into staging but left it on a branch, right?

kulla: > If I remember correctly last time we did not merge it into staging but left it on a branch, right?

yep, that's fine

hugotiburtino: ok, in that case, please close the PR if you don't have anything to change.

kulla:
@hugotiburtino Can the user journey already be tested?

src/components/content/entity.tsx
elbotho: let's see if that approach works with login and everything. If no please ping me and I'll find a way to at least not change the bundle size in production.

Entkenntnis:
@hugotiburtino branch and preview will stay even if PR is closed. To update branch, you can reopen the PR.

PR #3146, opened 2023-11-30
chore: Small changes for build editor branch

kulla:


package.json
```diff
@@ -29,7 +29,7 @@
     "build": "npm-run-all build:prisma build:next",
     "build:next": "next build",
     "build:prisma": "yarn prisma generate",
-    "build-editor": "tsc && vite build && tailwindcss -o dist/editor-tailwind.css --minify",
+    "build:editor": "tsc && vite build && tailwindcss -o dist/editor-tailwind.css --minify",
```
elbotho: I did not use `build:editor` on purpose. Feel free to rename but building frontend and building editor will (almost) never be in the same pipeline.

kulla: How about editor:build?

hejtful: I'm fine with editor:build

PR #3147, opened 2023-11-30
Improve exercise generation => publish flow through editor 

CodingDive:
- [x] Set changes as [KI generiert] for non subtasks 
- [x] Disable the "open in editor" button if exercise generation is loading
- [x] Allow to save immediately without having to make changes

### DIFF
--- a/src/serlo-editor/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
+++ b/src/serlo-editor/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
@@ -91,8 +91,15 @@ export function ToolbarMain({
       <button
         className="serlo-button-green pointer-events-auto ml-2 md:mr-[-11.5vw] lg:-mr-52 xl:-mr-64"
         onClick={() => {
-          if (isChanged) setSaveModalOpen(true)
-          else showToastNotice('👀 ' + editorStrings.noChangesWarning)
+          if (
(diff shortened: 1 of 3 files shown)


src/serlo-editor/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
```diff
@@ -91,8 +91,15 @@ export function ToolbarMain({
...
-          else showToastNotice('👀 ' + editorStrings.noChangesWarning)
+          if (
+            isChanged ||
+            changes?.value.startsWith('[KI generiert]:') ||
+            false
+          ) {
+            setSaveModalOpen(true)
+          } else {
+            showToastNotice('👀 ' + editorStrings.noChangesWarning)
+          }
```
CodingDive, line 102: I really dislike this hack and would rather modify the history state. 

You can see here how we set the `initialState` via `setEditorState` that is passed down all the way from the taxonomy file https://github.com/serlo/frontend/blob/staging/src/components/exercise-generation/exercise-preview-page.tsx#L223-L235. However, another thing I realized while working on this is that if `isChanged == true`, we get nasty popups upon clicking the "open it in editor" button, asking if we want to leave the page as then the `useLeaveConfirm(isChanged)` kicks in.

Entkenntnis (approved):

PR #3169, opened 2023-12-05
refactor(workspaces): yarn workspaces

elbotho:
Changelog:

- `/src` moved to `apps/frontend/src`
- all yarn commands can still be executed from project root, like `yarn` or `yarn dev`
- editors should be able to pick up correct config (prettier, eslint, tsc)
- CI and vercel now work with new folder structure

Todos:

- [x] avoid two subdirectories with same name frontend, e.g. rename `apps/frontend` to `apps/web`
- [x] check prisma integration - it's a bit all over the place :) at least make sure that it works - env setup is working on vercel
- [x] move readme back to top-level?
- [x] make prettier and eslint work with repo-wide settings
- [x] fix bundle size hack (e.g. https://frontend-git-yarn-workspaces-serlo.vercel.app/biologie)
- [x] update vs code settings for eslint


Entkenntnis:
Thank you very much! Let's get this working 👍

Entkenntnis:
We might need to add a new vercel imtegration with the subpath for the frontend to fix the build error. At least that's what most tutorials include.

elbotho:
BTW: I just aliased all frontend commands so we can work with them as we usually do. Basically just some paths changed.
Since this is still mostly frontend and we work on is a lot more I think that makes sense.

Not yet sure about the eslint errors.
Ah and new vercel intergration sounds smart.

Entkenntnis:
eslint errors related to prisma - I will take a look at it.

Entkenntnis:
@elbotho I would suggest disabling jest. The tests are not maintained, and also quite trivial? (and probably completely outdated atm ...)

Entkenntnis:
ups, we broke bundle size 👀

elbotho:
> @elbotho I would suggest disabling jest. The tests are not maintained, and also quite trivial? (and probably completely outdated atm ...)

I added some 2 months ago, but I'm okay with disabling the failing tests suite for now.


> ups, we broke bundle size 👀

yeah, that's weird. We should make sure the check is correct first though. Not even sure why it runs in this pr.

Entkenntnis:
very probably the cause of the bundle size increase: https://github.com/serlo/frontend/blob/cc00852e7381a995ba845c46e3f7e2e087558ef0/next.config.js#L34

maybe it has to change to apps/frontend/...?

elbotho:
Prettier works again for me, in frontend and on repo-root both with terminal and vscode.

ESlint also works with a new VSCode setting of `eslint.workingDirectories": [{ "mode": "auto" }]`. 
It seems to be a [common](https://stackoverflow.com/questions/67921812/typescript-aliases-not-working-with-vscode-eslint-in-monorepo) [thing](https://medium.com/devityoself/monorepo-eslint-vscode-6f5982c8404d) with monorepos and since it works from the terminal the project config should be correct.
I'm not sure if it actually worked before or I just did not see it, but I did not find another way.

There is still one unexpected ESLint error (only in IDE) in `feature-i18n.ts` at `getInstanceDataByLang`. Did not investigate far but that might be related to ramda.

PR #3176, opened 2023-12-06
Fill in the blanks: Final preparation for release

LarsTheGlidingSquirrel:
Makes the plugin available also outside `staging`.

Drag and drop mode is working on a basic level but there is no way for the user to change to this mode currently. Drag and drop mode will be worked on after this first release.

TODO before merge: 
- [x] Give pasted blanks a unique id 
- [x] Review


hejtful (approved):
Beside the `defaultLearnerFeedback` suggestion that we discussed, everything looks good to me!

src/serlo-editor/plugins/text/types/text-editor.ts
```diff
@@ -12,13 +12,17 @@ export type CustomElement =
   | MathElement
   | Blank
 
+export interface Answer {
+  answer: string
+  learnerFeedback?: string // Currently unused
```
kulla: * Is the learner feedback a formatted string (with math formulas, bold/italic, ...)?
* As long as it is unused it shouldn't end up in the final code in `staging`

src/serlo-editor/plugins/text/types/text-editor.ts
```diff
@@ -12,13 +12,17 @@ export type CustomElement =
...
+
 export interface Blank {
   type: 'blank'
   children: CustomText[]
   blankId: string // Used to uniquely identify a blank
-  correctAnswer: string
-  // Disabled alternative correct solutions for now
-  // alternativeSolutions: string[]
+  correctAnswers: Answer[]
+  incorrectAnswers?: Answer[] // Unused currently
```
kulla: Concerning https://miro.com/app/board/uXjVMgBV4FM=/?moveToWidget=3458764568362219903&cot=14 this is attached to the whole exercise and not to a gap inside the text.

src/serlo-editor/plugins/text/types/text-editor.ts
```diff
@@ -12,13 +12,17 @@ export type CustomElement =
...
+}
+
 export interface Blank {
   type: 'blank'
   children: CustomText[]
   blankId: string // Used to uniquely identify a blank
-  correctAnswer: string
-  // Disabled alternative correct solutions for now
-  // alternativeSolutions: string[]
+  correctAnswers: Answer[]
```
kulla:
```suggestion
  additionalAnswers: Answer[]
```

To avoid duplicate information with `children` only the additional answers should be stored here...

kulla (changes requested):

src/serlo-editor/plugins/text/types/text-editor.ts
kulla: Ah, `children` is always empty... interesting...

src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
```diff
@@ -12,6 +12,10 @@ import { EditorPluginType } from '@/serlo-editor/types/editor-plugin-type'
 
 export type FillInTheBlanksMode = 'typing' | 'drag-and-drop'
 
+export type BlankId = string
+
+export type DraggableId = string
+
 export const fillInTheBlanksExercise: EditorPlugin<FillInTheBlanksExerciseState> =
```
kulla:
```suggestion
export const ExerciseWithBlanks: EditorPlugin<FillInTheBlanksExerciseState> =
```

For me `FillInTheBlanks` already includes the mode `typing` -> What is a good name for `fill-in-the-blanks` and `drag-the-words`? `ExerciseWithBlanks` or `BlankExercise`?

LarsTheGlidingSquirrel: Hm, `ExerciseWithBlanks` and `BlankExercise` could also for example be an image where you have blanks. 
I changed it to `textWithBlanksExercise` for now but only for the database entries. Component renaming can be done later.

src/serlo-editor/plugins/text/types/text-editor.ts
LarsTheGlidingSquirrel: > Is the learner feedback a formatted string (with math formulas, bold/italic, ...)?

Just a plain string currently. I can see how math expressions and other formatting could be useful. 

> As long as it is unused it shouldn't end up in the final code in staging

I am planning to comment out everything that has to do with future features or state. -> Done!

src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
LarsTheGlidingSquirrel: Database entries will show `type: blanksExercise` now. Component renaming still to do, but this does not block the first release.

src/serlo-editor/plugins/text/types/text-editor.ts
kulla: > I can see how math expressions and other formatting could be useful.

That's a question for the PO / user research. A DB migration will be possible.

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
```diff
@@ -12,31 +11,40 @@ import { EditorPluginType } from '@/serlo-editor/types/editor-plugin-type'
...
+          TextEditorFormattingOption.richTextItalic,
           TextEditorFormattingOption.blank,
         ],
+        isInlineChildEditor: true,
       },
     }),
-    mode: string(defaultMode),
+    mode: string(defaultMode), // Currently there is no method for the user to switch to 'drag-and-drop'
+    // This could be used to store additional answers for mode drag-and-drop in the future
+    // additionalDraggableAnswers: optional(list(object({ answer: string() }))),
```
kulla, line 47: Idea: Can / should the `incorrectAnswers` from the blank be used here?

apps/web/src/serlo-editor/plugins/text/types/text-editor.ts
```diff
@@ -12,13 +12,23 @@ export type CustomElement =
...
   children: CustomText[]
   blankId: string // Used to uniquely identify a blank
-  correctAnswer: string
-  // Disabled alternative correct solutions for now
-  // alternativeSolutions: string[]
+  correctAnswers: Answer[]
+  // Here we could specify incorrect answers for this blank with specific learner feedback
+  // incorrectAnswers?: Answer[]
+  // Here we could add a default feedback for the learner
+  // defaultLearnerFeedback: string
```
kulla:
```suggestion
  // defaultIncorrectAnswerFeedback: string
```

apps/web/src/serlo-editor/types/editor-plugin-type.ts
```diff
@@ -27,7 +27,7 @@ export enum EditorPluginType {
   Exercise = 'exercise',
   ScMcExercise = 'scMcExercise',
   InputExercise = 'inputExercise',
-  FillInTheBlanksExercise = 'fillInTheBlanksExercise',
+  FillInTheBlanksExercise = 'blanksExercise',
```
kulla, line 30:
```suggestion
  BlanksExercise = 'blanksExercise',
```

apps/web/src/serlo-editor/plugins/text/types/text-editor.ts
```diff
@@ -12,13 +12,23 @@ export type CustomElement =
...
+  // Here we could specify a method to check the answer
+  // checkAnswerMethod?: 'text' | 'number' | 'math-expression'
```
kulla:
```suggestion
  // checkAnswerMethod?: 'text' | 'number' | 'math-expression'
```

* Maybe this should be merged with `type` to have `blank | blank-number | blank-math-expression`. -> Depends how the blank type should be changed. The current model makes sense when you first add a blank and then you configure the blank to a number blank. A change would be better when you can directely add a number blank in the editor...
* Type of `Answer` will be different depending of the type of the blank. For example:

```typescript
interface BlankBase {
  children: CustomText[]
  blankId: string
}

interface TextBlank extends BlankBase {
  type: "text-blank"
  correctAnswers: TextAnswer[]
}

interface NumberBlank extends BlankBase {
  type: "number-blank"
  answer: number
}
```

Same can be done with `checkAnswerMethod` btw... 

```typescript
interface BlankBase {
  type: "blank"
  children: CustomText[]
  blankId: string
}

interface TextBlank extends BlankBase {
  blankType: "text"
  correctAnswers: TextAnswer[]
}

interface NumberBlank extends BlankBase {
  blankType: "number"
  answer: number
}
```

kulla (approved):

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
LarsTheGlidingSquirrel: Good question! I need to think about this more.

apps/web/src/serlo-editor/types/editor-plugin-type.ts
LarsTheGlidingSquirrel: Done!

apps/web/src/serlo-editor/plugins/text/types/text-editor.ts
LarsTheGlidingSquirrel: I like this idea very much! :heart:  
I propose we remove `checkAnswerMethod` and define the type of the blank and the method for checking the answer through `type: 'textBlank' | 'numberBlank' | 'mathBlank'`
The editor toolbar could show one button to create a new blank (per default 'textBlank') and another UI element to switch the type when the blank is selected. 
-> Removed `checkAnswerMethod` and renamed `blank` to `textBlank`

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
```diff
@@ -12,31 +11,40 @@ import { EditorPluginType } from '@/serlo-editor/types/editor-plugin-type'
...
   return object({
     text: child({
       plugin: EditorPluginType.Text,
       config: {
         formattingOptions: [
-          ...defaultFormattingOptions,
+          TextEditorFormattingOption.code,
+          TextEditorFormattingOption.colors,
+          TextEditorFormattingOption.lists,
+          TextEditorFormattingOption.richTextBold,
```
kulla, line 38: I understand it correctly that `math` formatting is not allowed in the first go?

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -78,121 +61,134 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
   /** Maps blankId to the text that should be displayed in the blank.  */
   const textInBlanks = useMemo(() => {
     const newMap = new Map<BlankId, { text: string }>()
-    blanks.forEach((blankState) =>
+    blanks.forEach((blankState) => {
+      const firstCorrectAnswer = blankState.correctAnswers[0]?.answer ?? ''
```
kulla:
```suggestion
      const firstCorrectAnswer = blankState.correctAnswers.at(0)?.answer ?? ''
```

`.at()` already has the type `undefined | A` for a list `A[]`

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
LarsTheGlidingSquirrel: Yes! I removed it for now because the math toolbar is broken within the blanks exercise: You cannot switch from visual to latex mode.

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/index.tsx
LarsTheGlidingSquirrel: Thought about this but I am unsure! It totally makes sense that incorrect answers also cause draggables to appear for the user to drag. Otherwise, there would be no way (at least in drag-and-drop mode) for the user to answer with those. What I am unsure about is: What if the author adds an incorrect answer that is correct in another blank? Should there be two draggables then or just one?
I think we need to get input about that from UX and author perspective. I will ask them about it.

apps/web/src/serlo-editor/plugins/fill-in-the-blanks-exercise/renderer.tsx
LarsTheGlidingSquirrel: Nice! Did not know about that yet.

LarsTheGlidingSquirrel:
@kulla Rerequested your review. Things that changed: 
- Slate custom element type changed from `blank` to `textBlank` (so that we can have `numberBlank` and so on in the near future) and removed `checkAnswerMethod`
- Rewrote `toggleBlank` function. Tested this a bunch and seems to work in all cases now. 
- Added slate normalization that gives blanks a unique id (needed when copy&past blanks)

apps/web/src/serlo-editor/editor-ui/plugin-toolbar/text-controls/utils/blank.ts
```diff
@@ -1,74 +1,74 @@
...
+    const anchorPath = editor.selection.anchor.path
+    const focusPath = editor.selection.focus.path
+
+    // Node.elements(...) needs the "smaller" path in 'from'. Otherwise it will return nothing. Depending on how the user selects (expanding selection to the left or right) the anchor path might be "smaller" or "bigger" than the focus path. Here we figure out which path is the smaller/bigger one.
+    const isAnchorLeftOfFocus = Path.compare(anchorPath, focusPath) === -1
+    const startPath = isAnchorLeftOfFocus ? anchorPath : focusPath
+    const endPath = isAnchorLeftOfFocus ? focusPath : anchorPath
+    const allElementsInSelection = [
+      ...Node.elements(editor, { from: startPath, to: endPath }),
+    ]
```
kulla:
```suggestion
    const range = isAnchorLeftOfFocus ? { from: anchorPath, to: focusPath } : { from: focusPath, to: anchorPath }
    const allElementsInSelection = [
      ...Node.elements(editor, range),
    ]
```

kulla (approved):

PR #3177, opened 2023-12-06
refactor(workspaces): e2e-test as workspace

elbotho:
https://github.com/serlo/frontend-e2e-tests -> `apps/e2e-tests`


elbotho:
@Entkenntnis do you know why playwright ran without `yarn playwright install` in the old GitHub Action?

Entkenntnis:
it just worked 🤷‍♀️ probably run automatically.

Anyways - I thought about *not* adding it as workspace, but simple into a folder like `./e2e-tests`, reasons: downloading browsers is quite heavy and we don't need it for the frontend + there is no coupling with existing code at all, so I don't see advantages of using it as a yarn workspace (maybe prettier can be shared, but that's only nice to have)

Thoughts?

elbotho:
(on slack)

elbotho:
tests fail right now but the output is exactly the same as on staging.

hejtful (approved):

elbotho:
@Entkenntnis I don't really get why the tests are failing here. Any idea?

Entkenntnis:
maybe login is not possible because it's not staging?

PR #3181, opened 2023-12-07
add prisma commands to workspace

### LINKED ISSUE #3179: update prisma setup for workspaces
Where should we store our env strings, in root or in app?

also update readme

Entkenntnis:


package.json
```diff
@@ -30,10 +30,13 @@
     "lint:eslint": "yarn workspace @serlo/frontend lint:eslint",
     "lint:prettier": "yarn workspace @serlo/frontend lint:prettier",
     "lint:tsc": "yarn workspace @serlo/frontend lint:tsc",
+    "prisma:db-pull": "dotenv -e apps/web/.env.local yarn workspace @serlo/frontend prisma db pull",
```
elbotho, line 33: 🤔 Interesting that this was needed. I would have guessed that it just finds the right .env file…

Entkenntnis: prisma is not that clever and can't handle the nextjs conventions well ..

PR #3136, opened 2023-11-28
tmp: editor package with vite

elbotho:
this is the first step: export editor as react component

### what works: 
- editor and tools to customize can be imported and used in other react repo
- tailwind css styles are exported and can be imported from package
- important types are included

### what does not work / is not included here:
- assets (icons, katex font)
- using workspaces
- limit dependencies

### example usage from other repo

``` ts
import {
  RowsStaticRenderer,
  TextStaticRenderer,
  editorPlugins,
  editorRenderers,
  instanceDataDe,
  loggedInDataDe,
  SerloEditor,
  EditorPluginType,
  createTextPlugin,
  createRowsPlugin,
  MathElement,
  StaticMath,
} from '@serlo/editor'
import {
  InstanceData,
  LoggedInData,
} from '@serlo/editor/dist/src/data-types'
import { LinkRenderer } from '@serlo/editor/dist/src/serlo-editor/plugin/helpers/editor-renderer'
import { ComponentProps } from 'react'

export default function Demo() {
  editorPlugins.init([
    {
      type: EditorPluginType.Text,
      plugin: createTextPlugin({}),
      visibleInSuggestions: true,
      icon: <>text</>,
    },
    { type: EditorPluginType.Rows, plugin: createRowsPlugin() },
  ])

  editorRenderers.init({
    pluginRenderers: [
      { type: EditorPluginType.Rows, renderer: RowsStaticRenderer },
      { type: EditorPluginType.Text, renderer: TextStaticRenderer },
    ],
    mathRenderer: (element: MathElement) => <StaticMath {...element} />,
    linkRenderer: ({ href, children }: ComponentProps<LinkRenderer>) => {
      return <a href={href}>{children}</a>
    },
  })

  return (
    <SerloEditor
      instanceData={instanceDataDe as InstanceData}
      loggedInData={loggedInDataDe as LoggedInData}
    />
  )
}

```

PR #3183, opened 2023-12-08
refactor(workspaces): editor package

elbotho:
## status

- VSCode works as expected now
- You can now run `yarn lint` inside of the editor package as well
- global **prettier** and **eslint** scripts now run in editor and frontend workspace in parallel
- tsc is only checked for the frontend now, the editor code that is used is checked by the frontend `yarn lint:tsc` as well

## Remining errors inside of editor
- **tsc**: some errors because svg import only works with the `fileLoader` and `svgr` correctly set up within the webpack config of nextjs. This will have to be solved when we build an editor package.

--- 

This is the first step of a cleaner separation between editor and frontend again.

what I did:
- moved `src/serlo-editor` to `package/editor/src`
- copied and adapted config and package.json in editor package
- import `@serlo/editor` in frontend and `@serlo/frontend` in editor via `workspace:*`
- adding path `@editor/*` that resolves to `packages/editor/src/*` in both workspaces
- adapt all imports
- had to import/fix some types that broke because of codegen (see TODOs)

[diff view filtered (no ts*)](https://github.com/serlo/frontend/pull/3183/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.svg&show-deleted-files=true&show-viewed-files=true)

follow ups (help welcome):
- [x] run linter on all workspaces
- [x] share config files (eslint mandatory, prettier and tailwind optional) instead of separating
- prepare build for standalone editor build (inspired by https://github.com/serlo/frontend/pull/3136)
- cleanup types / move graphql stuff in frontend
- move frontend/serlo specific stuff (especially nextjs) back to the frontend
- check the > 300 imports from the frontend

Just a note:
I did not use `reference`s in the root tsconfig. I'm not sure if that would be the better solution but since it now works like this I'd probably leave it like it is. 🤷
https://www.typescriptlang.org/docs/handbook/project-references.html#what-is-a-project-reference


hejtful (approved):

kulla (approved):
Thx :heart:

Entkenntnis (approved):
looks good so far - hopefully no surprises 🤷‍♀️

PR #3191, opened 2023-12-11
refactor(blank-exercise): Refactor toggleBlank()

kulla:


LarsTheGlidingSquirrel:
Thank you for this fix! Sadly I noticed that selecting blank->text->blank still did not give the correct result. However, it gave me an idea how to solve this. I changed it and it seems to work in all cases now.

PR #3193, opened 2023-12-12
refactor(e2e): add e2e-test as standalone package

elbotho:
(replacement of https://github.com/serlo/frontend/pull/3177)
- this adds e2e tests as a standalone package instead of a workspace (like you suggested @Entkenntnis) 
- somehow that means we don't have to install playwright in the CI (still don't know why but it's nice)
- it uses the latest test code now (sorry @hejtful  …)
- seems to work even with workers again, but let's see how reliably (@hejtful your improvements are still welcome here)


hejtful (approved):
Added some fixes and improvements, looks good to me for now.

Entkenntnis (approved):
thank you! looks like we can retire the repo `frontend-e2e-tests` from now on

PR #3194, opened 2023-12-12
chore(e2e): update playwright, update readmes

elbotho:
- e2e tests now also run on branches starting with `e2e-` or `e2e/`
- added e2e section in main readme
- updated playwright package
- added instructions on how to install playwright browsers (since that now has to be done as an extra step)
- installs playwright browsers in ci (installs ff only for ff tests and chromium for main test run)

Unfortunately tests break again if I update `codeceptjs` to the most current version (3.5.5 still works) so this will be a follow up.
https://github.com/codeceptjs/CodeceptJS/releases


.github/workflows/e2e_tests.yaml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - staging
-      - feat-e2e-tests
+      - e2e-**
```
Entkenntnis, line 7: @elbotho I would suggest added a comment like this:

`# even if tests run on a different branch, they are hard coded to the staging deployment`

Entkenntnis (approved):
looks good, added one comment that could clearify the scope of the tests more

PR #3203, opened 2023-12-13
simplify env vars and better debug output

### LINKED ISSUE #3199: simplify env-system
- [ ] move public constants out of .env - applying changes to them will need a redeployment anyway, so we can just store them as code
- [ ] identify necessary env vars, confirm there value on build for better debugging
- [ ] fix VERCEL_GITHUB_COMMIT_SHA (renamed)
- [ ] deprecate `.env.production` - at this point it contains no important things anyways.

Entkenntnis:
- to be sure that everything works I need a deployment - being careful
- remove branch protection rules for env checks
- maybe reduce output on dev


apps/web/.env
```diff
@@ -1,18 +1,3 @@
...
-HYDRA_CLIENT_ID_STAGING=frontend
-HYDRA_CLIENT_SECRET_STAGING=YjgwZThhZDk0MTYxMTA4ODY1NzYxZmJk
-HYDRA_HOST_STAGING=https://hydra.${NEXT_PUBLIC_SERLO_DOMAIN_STAGING}
-HYDRA_CLIENT_ID_LOCAL=frontend.serlo.org
-HYDRA_CLIENT_SECRET_LOCAL=frontend.serlo.org
-HYDRA_HOST_LOCAL=http://localhost:4444
-KRATOS_HOST_STAGING=https://kratos.${NEXT_PUBLIC_SERLO_DOMAIN_STAGING}
-KRATOS_HOST_LOCAL=http://localhost:4433
\ No newline at end of file
+NEXT_PUBLIC_ENV=staging
```
elbotho, line 3: just an idea: If we would add a script like this we could remove `.env` as well and make it even easier to find this functionality.
`"dev-local": "cross-env NEXT_PUBLIC_ENV=local next",`

elbotho: I guess the downside is that we either have to fall back to staging sometimes (like we already do in the code below) or set `cross-env NEXT_PUBLIC_ENV=staging ` for the `dev` script as well.
```
// if env is not set, it's a production build running on localhost, use staging as default
const KRATOS_HOST =
  KRATOS_HOSTS[
    process.env.NEXT_PUBLIC_ENV ? process.env.NEXT_PUBLIC_ENV : 'staging'
  ]
```

elbotho:
We should probably do a test run with local api and kratos.
I don't have it set up right now, do you?

Entkenntnis:
Highest prio right now is the production deployment - I think your ideas are good and would like to persue them in follow-up issues.

PR #3210, opened 2023-12-14
refactor(gap-exercise): Shorten code for normalizeNode

kulla:


LarsTheGlidingSquirrel (approved):
I like that `normalizeNode` is called only at the bottom now. Thank you!

PR #3218, opened 2023-12-14
feat(e2e): add basic editor focus tests

### LINKED ISSUE #3219: Extract focus management tests in a separate suite
An action item from the focus management retro. In order to test focus management feature more easily, all focus management related tests should be in their own suite.

Best started after https://github.com/serlo/frontend-e2e-tests/pull/81 is merged, to avoid conflicts.

elbotho:


Entkenntnis (approved):
nice idea with `:focus`!

PR #3229, opened 2023-12-17
Add text area exercise to staging

kulla:


kulla:
Since it does not change the implementation of the editor in serlo (it is not added to the list of plugins) – I am going to merge it since I need it in https://github.com/serlo/frontend/pull/3220

PR #3257, opened 2024-01-08
feat(plugin-blanks): autogrow input blanks

elbotho:


LarsTheGlidingSquirrel (approved):
I have no clue how this works! But if it works, I'm fine with it. :)

elbotho:
> I have no clue how this works! But if it works, I'm fine with it. :)

Yeah it's a bit magic but it's the easiest approach I could find. The other version would be a not very well maintained library so I'm relatively happy with this one 😄 

It's basically inserting an element with css with the same content and text style as the input and that determines the size of the input field.

PR #3258, opened 2024-01-08
feat(blanks-plugin): show only the blank controls when a blank is selected

elbotho:


LarsTheGlidingSquirrel (approved):
Very nice solution! Thank you! :)

PR #3260, opened 2024-01-08
feat(plugin-blanks): prominent blank button alternative

elbotho:


packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
```diff
@@ -36,21 +38,32 @@ export function PluginToolbarTextControls({
         {controls.map((control, index) => {
           if (mathActive && !isMath(control)) return null
           if (blankActive && !isBlank(control)) return null
+
+          const next = controls[index + 1]
```
LarsTheGlidingSquirrel: Maybe use `controls.at(index + 1)` here instead. Returns `ControlButton | undefined` instead of `ControlButton` with implicit undefined if outside array limits.

LarsTheGlidingSquirrel (approved):
I like it! Also the separators in the toolbar make it easier to process visually.

packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
elbotho: interesting difference, thank you.

PR #3262, opened 2024-01-10
WIP: Fix/exercise number heading alignment

CodingDive:
Gregor and I noticed in production that the numbers for exercises are quite a bit offset https://de.serlo.org/mathe/23869/aufgaben-zur-prozentrechnung. To my surprise, they are positioned absolutely, so I changed it to flexbox.

Before
image

By doing so, I realized that we have a lot of styles applied at a component instead of a container level. This leads to a lot of style duplication and makes it harder for things to align nicely.

E.g here a screenshot of the devtools, one can see that the breadcrumbs, h1, paragraph and every other plugin has a margin-left of some sorts instead of handling it once. Sometimes they even have different margins for smaller devices which makes it even harder to coordinate all components being aligned.
image

After (when removing the extra margin-top and padding-top from the header, I'm playing with this locally but it's not part of the PR yet as it'll probably mess up other things :see_no_evil:)

image

My suggestion would be to remove all component-based margin-tops/margin-lefts and handle it on the container level. Please give some quick feedback here or in Slack if you'd like me to pursue this further, or if there is a good reason for why the styles are the way they are now.


Entkenntnis:
alignment in [preview ](https://frontend-git-fix-exercise-number-heading-alignment-serlo.vercel.app/23869) is worse now:

grafik

Entkenntnis:
for some historic context: we never planned for headings in exercises (the positioning was done under the assumption that all content will start without margin/padding-top). Some clever authors started to use this method.

Removing the padding to on the h2-heading could do the trick as well (without touching too much code) - maybe some css can be used here. Not sure if the padding-top is necessary at all and can be removed (because we always use margin bottoms to vertically space)

CodingDive:
> alignment in [preview ](https://frontend-git-fix-exercise-number-heading-alignment-serlo.vercel.app/23869) is worse now:
> 
> grafik

Yes, only aligns perfectly when removing the mt-[30px] and pt of the heading. Just wanted to check in if I'm on the right track.

CodingDive:
I think there may still a lot of issues and things I need to validate but as we have said on Monday, I'll try to keep it as small as possible.

Below are some screenshots of things that look better, the same or slightly different. 


## Better
**This PR**  https://frontend-8x0o9aptc-serlo.vercel.app//23869
image

**Production** https://de.serlo.org/23869
image

**This PR**
image


**Production**
image


**This PR** (Anchor is aligned perfectly)
image

**Production**
image




# Slightly different 
Let me know which of these are deal breakers for you and I'll try to find a better solution.

**This PR** (drag & drop indicator for multimedia plugin)
image

**Production**
image

**This PR** Active row indicator in nested plugins
image

**Production**
image

**This PR** (toolbar of box content starts at the offset of the content)
image

**Production**
image



# Same 
Hopefully most things!

**This PR**
image

**Production**
image

**This PR**
image


**Production**
image

Entkenntnis:
I have marked this PR as keep-for-preview.

I think the timebox we set for this has expired. Maybe we can find a good way of improving thinks in smaller steps without having to change to many styles at once.

At least a separation between exercise-numbering and general spacing could be helpfu.

PR #3263, opened 2024-01-10
WIP: feat(plugin-blanks): Toggle preview/edit mode from toolbar

CodingDive:
Added WIP/draft label as there are a lot of open PRs. Once they get merged, I'll be happy to fix the merge conflicts and update the PR (night Germany time). I'm also running in a few circular dependencies and couldn't figure out how to make the types nicer. 

Will investigate further tomorrow.

### DIFF
--- a/packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
+++ b/packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
@@ -1,22 +1,33 @@
 import { PluginToolbar } from '@editor/editor-ui/plugin-toolbar'
 import { selectDocument, selectIsFocused, useAppSelector } from '@editor/store'
 import { EditorPluginType } from '@editor/types/editor-plugin-type'
+import { faEye, faPen } from '@fortawesome/free-solid-svg-icons'
+import { useState } from 'react'
 
-import type { FillInTheBlanksExerciseProps, FillInTheBlanksMode } from '.'
+import type {
+  FillInTheBlanksExerciseProps,
+  FillInTheBlanksMode,
+  fillInTheBlanksExercise,
+} from '.'
 import { FillInTheBlanksRenderer } from './renderer'
+import { FillInTheBlanksStaticRenderer } from './static'
 import { InteractiveToolbarTools } from '../exercise/toolbar/interactive-toolbar-tools'
+import { FaIcon } from '@/components/fa-icon'
 import { useEditorStrings } from '@/contexts/logged-in-data-context'
 
 export function FillInTheBlanksExerciseEditor(
   props: FillInTheBlanksExerciseProps
 ) {
   const { focused } = props
+  const [isEditMode, setIsEditMode] = useState(true)
+  const toggleRenderEditMode = () =>
+    setIsEditMode((currentEditMode) => !currentEditMode)
+  const editorStrings = useEditorStrings()
+  const blanksExerciseStrings = editorStrings.plugins.blanksExercise
 
   const isRendererTextPluginFocused = useAppSelector((storeState) => {
     return selectIsFocused(storeState, props.state.text.id)
   })
-
-  const editorStrings = useEditorStrings()
 
   const hasFocus = focused || isRendererTextPluginFocused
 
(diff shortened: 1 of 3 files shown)


packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -26,27 +37,66 @@ export function FillInTheBlanksExerciseEditor(
...
+        <FillInTheBlanksStaticRenderer
+          state={{
+            text: {
+              ...textPluginState,
+              state:
+                (
+                textPluginState?.state as (typeof fillInTheBlanksExercise)['state']
+                // prettier-ignore
+                // @ts-ignore
+              )?.value,
```
CodingDive, line 93: :vomiting_face: 
I'm not sure why, but the state has a random `value` property. I think this has to do with the `createState` function of index.tsx. I was hoping I could use getStaticDocument or another helper method but haven't managed to get the types right.

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -1,23 +1,34 @@
...
+import { FaIcon } from '@/components/fa-icon'
 import { useEditorStrings } from '@/contexts/logged-in-data-context'
 
 export function FillInTheBlanksExerciseEditor(
   props: FillInTheBlanksExerciseProps
 ) {
   const { focused } = props
+  const [isEditMode, setIsEditMode] = useState(true)
+  const toggleRenderEditMode = () =>
+    setIsEditMode((currentEditMode) => !currentEditMode)
```
elbotho, line 24:
```suggestion
  const toggleRenderEditMode = () => setIsEditMode(!isEditMode)
```

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -1,23 +1,34 @@
...
+  const editorStrings = useEditorStrings()
+  const blanksExerciseStrings = editorStrings.plugins.blanksExercise
```
elbotho, line 26:
```suggestion
  const blanksExerciseStrings = useEditorStrings().plugins.blanksExercise
```

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -26,27 +37,66 @@ export function FillInTheBlanksExerciseEditor(
...
 
   if (!textPluginState) return null
+  console.log('Editor: ', { isEditMode, textPluginState, props })
 
   return (
     <div className="mb-12 mt-10 pt-4">
-      {hasFocus ? (
+      {hasFocus || !isEditMode ? (
         // TODO: Add button to toggle between fill-in-the-blanks and drag-and-drop
         <PluginToolbar
```
elbotho, line 46: move toolbar to own file probably makes sense now (see other plugins)

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -26,27 +37,66 @@ export function FillInTheBlanksExerciseEditor(
...
+                  <>
+                    <FaIcon icon={faEye} /> {blanksExerciseStrings.preview}
+                  </>
+                ) : (
+                  <>
+                    <FaIcon icon={faPen} /> {blanksExerciseStrings.edit}
+                  </>
+                )}
+              </button>
+            </>
```
elbotho, line 67: all fragments here can be removed

elbotho: while this is probably a bit clearer I stil prefer the approach [here](https://github.com/serlo/frontend/blob/91c884a791814cf9329176d99c2c235d92ae79bd/packages/editor/src/plugins/input-exercise/toolbar.tsx#L31-L45) since it does not cause the button to suddenly change length.

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
elbotho: have a peek into input-exercise where I added a preview move already:
https://github.com/serlo/frontend/blob/staging/packages/editor/src/plugins/input-exercise/editor.tsx#L33-L36
should also work here I hope.

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -26,27 +37,66 @@ export function FillInTheBlanksExerciseEditor(
   })
```
elbotho, line 37: not your change, but I'd prefer a different name for `textPluginState` since it uses selectDocument. Maybe just `textDocument`

elbotho (changes requested):
not sure why you requested a review on a WIP draft so I left a lot of comments and hope that's what you wanted ^^

elbotho:
the dependency errors can be solved by adding a `type` [here](https://github.com/serlo/frontend/blob/43319d51393b4f05d7b63efcec812b5bc6aef7c0/packages/editor/src/plugins/fill-in-the-blanks-exercise/static.tsx#L2) and [here](https://github.com/serlo/frontend/blob/43319d51393b4f05d7b63efcec812b5bc6aef7c0/packages/editor/src/types/editor-plugins.ts#L8).

packages/editor/src/plugins/fill-in-the-blanks-exercise/editor.tsx
```diff
@@ -26,27 +37,66 @@ export function FillInTheBlanksExerciseEditor(
...
 
   if (!textPluginState) return null
+  console.log('Editor: ', { isEditMode, textPluginState, props })
 
   return (
     <div className="mb-12 mt-10 pt-4">
-      {hasFocus ? (
+      {hasFocus || !isEditMode ? (
```
elbotho, line 44: what is this supposed to do?

CodingDive: Thank you for all the feedback! Will integrate it later.
Here my idea was that in preview mode you always see the toolbar. Just to indicate that we are in preview, not edit mode. Maybe there is a better way to visually indicate this than to potentially render 2 toolbars 😵‍💫

CodingDive:
Closing in favor of https://github.com/serlo/frontend/pull/3265 and subsequent PRs that I'm basing against the blanks-add-type-switch branch.

PR #3265, opened 2024-01-10
feat(plugin-blanks): toolbar tools (preview, mode switcher)

elbotho:
also:
- Delete gap when del/backspace is pressed within an empty gap (by @CodingDive)
- refactor: new toolbar switch component with fixed padding


packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
```diff
@@ -30,7 +30,8 @@ export function BlankRenderer({ element }: BlankRendererProps) {
 
     // Focus editor when the blank is removed
     return () => {
-      ReactEditor.focus(editor)
+      // TODO: uncommented this breaks the preview mode somehow, investigate
+      // ReactEditor.focus(editor)
```
elbotho: @hejtful do you have an idea why this could happen? 
when I activate the preview this line breaks the editor (you could test it in this branch).

@CodingDive this might have got you as well over in https://github.com/serlo/frontend/pull/3263

CodingDive: Yeah! Also ran into it. I'll close my PR and send PRs against yours. 
For this exception, I tried a lot of things. Slate is confusing, focus management can be confusing but nothing is quite as confusing as focus management within Slate :grinning: This PR seems to solve it for me, but I hate to introduce the timeout... https://github.com/serlo/frontend/pull/3266

elbotho: @CodingDive I merged [your fix](https://github.com/serlo/frontend/pull/3266).

packages/editor/src/plugins/fill-in-the-blanks-exercise/toolbar.tsx
```diff
@@ -21,6 +21,9 @@ export const FillInTheBlanksToolbar = ({
 }) => {
   const blanksExerciseStrings = useEditorStrings().plugins.blanksExercise
 
+  // TODO: only for testing while drag & drop mode is not actually available
+  const allowModeSwitch = true
```
elbotho: flip before merge

packages/editor/src/editor-ui/plugin-toolbar/components/toolbar-select.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function ToolbarSelect({
+  tooltipText,
+  value,
+  changeValue,
+  options,
+  dataQa,
+}: ToolbarSelectProps) {
+  return (
+    <label className="serlo-tooltip-trigger relative mx-1.5 text-sm">
```
CodingDive: I like your changes, it looks good. Although there is no guarantee that the buttons and select in the toolbar will have the same size with different zoom levels. I just found out yesterday that the caret of the select (at least in Chrome) changes the height of the select compared with a button with the exact same padding, margin, line height, font size, etc. :grimacing: One can only hope for https://open-ui.org/components/selectlist/ to ship in all browsers!

I believe the "relative" is not needed here.

elbotho: > I believe the "relative" is not needed here.

Thank you, removed it.

Now that we have all toolbar selects using one component we can easily change the styling later (or use a component from some library) if we run into problems.

packages/editor/src/editor-ui/plugin-toolbar/components/toolbar-select.tsx
```diff
@@ -0,0 +1,41 @@
+import { EditorTooltip } from '@editor/editor-ui/editor-tooltip'
+
+import { cn } from '@/helper/cn'
+
+interface ToolbarSelectProps {
+  value: string
+  dataQa?: string
+  changeValue: (value: string) => void
```
CodingDive, line 8: Prefixing this with `on` would make it more clear that it's the event handler we're passing. Either way, the types here help a lot.

packages/editor/src/plugins/highlight/toolbar.tsx
```diff
@@ -19,33 +19,15 @@ export const HighlightToolbar = ({ id, state }: HighlightProps) => {
...
-              })}
-            </select>
-          </div>
+          <ToolbarSelect
+            tooltipText={highlightStrings.languageTooltip}
+            value={state.language.value}
+            changeValue={(value) => state.language.set(value)}
+            options={languages.map((language) => ({
+              value: language,
+              text: language.charAt(0).toUpperCase() + language.slice(1),
```
CodingDive, line 28: Technically it's Java**S**cript, not `Javascript` which will get rendered here but I don't think it's a big deal or most people care. In fact, I usually prefer to write Javascript too.

CodingDive (approved):
It's looking great! 🙌 
Reminder to change this boolean back before merging as you already mentioned! https://github.com/serlo/frontend/pull/3265/files#diff-bb54506b2496e27bf92840ab92108c911f92c0e28c6d21030cfe687863714808R25

packages/editor/src/editor-ui/plugin-toolbar/components/toolbar-select.tsx
elbotho: But it's not the event handler we are passing…?

CodingDive: I see what you mean, it's a function that reacts to the event value with the `event.target` stripped.
The naming is not a big deal which is why I approved the PR. I personally prefer to prefix booleans with "is" and fns we are passing to inputs with "on".

CodingDive (changes requested):
Sorry for not finding it yesterday already. I found an overflow issue in the toolbar that I outlined and proposed to fix here https://github.com/serlo/frontend/pull/3275

hejtful (commented):
Can confirm the styling issue that @CodingDive reported.

Screenshot 2024-01-15 at 11 20 14

But not sure if removing the toggle code button is the right solution...

elbotho:
My solution is just shortening the german string, but I can only do that on staging.
En version should already work.

hejtful (approved):
@elbotho In that case, I'd say let's merge

CodingDive (approved):
Let's shorten the strings with Crowdin then :crossed_fingers:

CodingDive:
Could you review & merge https://github.com/serlo/frontend/pull/3268 in here before merging to staging? :pray:

PR #3266, opened 2024-01-11
fix(blanks): Focus editor after gap is removed

CodingDive:
Would be so happy to see a better solution for this but after a lot of trial and error, this seems to be one of the only options with good UX + no exception.

### DIFF
--- a/packages/editor/src/editor-ui/plugin-toolbar/text-controls/utils/blank.ts
+++ b/packages/editor/src/editor-ui/plugin-toolbar/text-controls/utils/blank.ts
@@ -54,7 +54,7 @@ function removeBlanks(editor: SlateEditor) {
   // After the blank is removed, we must refocus the editor. Without the timeout
   // it blows up and moving this code to the unmounting of the blank-renderer
   // also blows up when changing from preview to edit mode.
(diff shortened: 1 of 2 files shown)


packages/editor/src/editor-ui/plugin-toolbar/text-controls/utils/blank.ts
```diff
@@ -50,6 +51,10 @@ function removeBlanks(editor: SlateEditor) {
       },
     })
   })
+  // After the blank is removed, we must refocus the editor. Without the timeout
+  // it blows up and moving this code to the unmounting of the blank-renderer
+  // also blows up when changing from preview to edit mode.
+  setTimeout(() => ReactEditor.focus(editor), 1)
```
elbotho: this would probably also work, right?
```suggestion
  setTimeout(() => ReactEditor.focus(editor))
```
I prefer to use `0` or no value to make it clear that it does not have to be exactly `1ms` but just not in the current thread (if that is the case)

CodingDive: I do think so and will add some e2e tests to confirm!

PR #3269, opened 2024-01-11
fix(gaps): Delete gap when del/backspace is pressed within an empty gap

CodingDive:
Let me know what you think of this behavior. This is how I'd expect it to be, but I'm curious if someone prefers not having the gap deleted. 

Edit: It just dawned on me, that  I need to check tomorrow if this will delete the gap in render mode for the students solving the exercise :see_no_evil: Obviously, it should only delete the gap when the author is editing the gaps :sweat_smile:

### DIFF
--- a/apps/web/src/assets-webkit/styles/serlo-tailwind.css
+++ b/apps/web/src/assets-webkit/styles/serlo-tailwind.css
@@ -355,6 +355,17 @@
     }
   }
 
+  /* When using appearance-none, we need to add back the caret */
+  .serlo-custom-select::after {
+    /* caret symbol */
+    content: '▼';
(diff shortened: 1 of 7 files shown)


elbotho:
> Edit: It just dawned on me, that I need to check tomorrow if this will delete the gap in render mode for the students solving the exercise 🙈 Obviously, it should only delete the gap when the author is editing the gaps 😅

Since that view uses the static renderer this should not happen.

elbotho (approved):
nice behaviour, simple implementation 👍

elbotho:
merged into https://github.com/serlo/frontend/pull/3265 via cherry-pick

PR #3273, opened 2024-01-12
feat(plugin): Add new type of root plugin `type-generic-content`

LarsTheGlidingSquirrel:
Draft for a new root plugin that wraps around the content on a page. 

Motivation: On serlo.org content is wrapped inside plugins like 'article'. In the RLP integration there is no wrapper, its just a 'rows' plugin directly. This wrapper is intended to be used in the RLP integration but can be used in other integrations as well. 

Idea: A plugin with a mandatory 'content' property (a rows plugin) that can be extended in the future by additional optional properties (title, id, ...) without having to do database migrations.


packages/editor/src/plugins/root-with-optionals/renderer.tsx
```diff
@@ -0,0 +1,5 @@
+import { ReactNode } from 'react'
+
+export function RootWithOptionalsRenderer({ content }: { content: ReactNode }) {
+  return <>{content}</>
+}
```
elbotho: def. a bit verbose, but I don't mind for consistency / possible future features.

elbotho (approved):
Nice, very clean.
If you resolve the naming discussion with kulla ~and remove the frontend imports (as you noted)~ this can be merged.

elbotho (changes requested):
ah one more thing. For serlo we call this a template-plugin and they are all in this folder:
`plugins/serlo-template-plugins` so for me it would be clearer if we had a similar folder or have the template in the name somewhere.

elbotho:
(resolved conflicts and removed frontend imports)

elbotho (approved):
Thanks!

PR #3275, opened 2024-01-13
fix(blanks): Overflow of text toolbar

CodingDive:
While working on the tests, I saw in the PR https://github.com/serlo/frontend/pull/3265 that the toolbar is overflowing with too many controls. 

Screenshot from 2024-01-13 10-05-54

There are multiple ways we could fix it. Either disallow some of the other toolbar controls like I did in this PR by removing the code editor within fill-in-the-gap. Alternatives I could see to removing the code toggle are the two list buttons. 
image

If however, that is undesired and one should be able to create code blocks or lists within the fill-in-the-gaps, we could think of changing the title `Lückentext Aufgabe` to just `Lückentext` (10 chars) which also makes it work. However, with the internationalized English version, "Fill in the blanks" (18 chars) I haven't tried if it'd still overflow. 

@LarsTheGlidingSquirrel which text controls should be available? To me it'd make sense to only render those that are allowed within a fill-in-the-blanks exercise.

### DIFF
--- a/packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
+++ b/packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
@@ -3,7 +3,11 @@ import { Fragment, useState } from 'react'
 import { Editor as SlateEditor } from 'slate'
 
 import { PluginToolbarTextControlButton } from './plugin-toolbar-text-control-button'
-import type { NestedControlButton, ControlButton } from './types'
+import {
+  type NestedControlButton,
+  type ControlButton,
+  TextEditorFormattingOption,
+} from './types'
 import { FaIcon } from '@/components/fa-icon'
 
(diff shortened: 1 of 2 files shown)


elbotho:
> However, with the internationalized English version, "Fill in the blanks" (18 chars) I haven't tried if it'd still overflow.

I did change the string in the toolbar PR you reviewed. It works in both `en` and `de` for me.

CodingDive:
Removing the code option here https://github.com/serlo/frontend/blob/staging/packages/editor/src/plugins/fill-in-the-blanks-exercise/index.tsx#L36 is much better.

CodingDive:
> I did change the string in the toolbar PR you reviewed. It works in both en and de for me.

Great! I just checked if I was zoomed in and I don't think I was. Maybe trying it with different viewport sizes leads to the hiccup?

CodingDive:
Closing in favor of trying to make the strings shorter in staging, see comment in  #3265

PR #3276, opened 2024-01-13
E2E tests for fill in the gaps

CodingDive:
Please note the failing test in https://github.com/serlo/frontend/pull/3276/files#diff-104cb85978425664605bb4849f059a3fe1a58e7a10414a9210cd00e4d2048427R115. It's a minor bug in the focus management.

Also, it'd be great if when the fill-in-the-gap plugin is created, we autofocus the `text-editor` instead of having to click it before being able to type. I couldn't figure out how to do so but maybe one of the editor-focus wizards know :mage_man:  cc @hejtful @elbotho


e2e-tests/tests/450-blank-exercise.ts
```diff
@@ -0,0 +1,199 @@
+import assert from 'assert'
+import { popupWarningFix } from './helpers/popup-warning-fix'
+
+Feature('Serlo Editor - Blank exercise')
+
+Before(popupWarningFix)
+
+// TODO ideally import from
+// editor/src/types/editor-plugin-type.ts/EditorPluginType.FillInTheBlanksExercise
```
hejtful: I wouldn't be doing that in e2e tests, tbh.

e2e-tests/tests/450-blank-exercise.ts
```diff
@@ -0,0 +1,199 @@
...
+
+Scenario('Create and remove fill in the gap exercise', async ({ I }) => {
+  I.amOnPage('/entity/create/Exercise/23869')
+
+  I.seeNumberOfElements('$plugin-text-editor', initialTextPluginCount)
+  I.click(`$add-exercise-${FillInTheBlanksExercise}`)
+  I.seeNumberOfElements('$plugin-text-editor', initialTextPluginCount + 1)
+
+  // I.click('$plugin-text-leaf-element')
+  // I.click('Schreibe einen Text und füge Lücken ein')
```
hejtful: Leftover?

e2e-tests/tests/450-blank-exercise.ts
```diff
@@ -0,0 +1,199 @@
+import assert from 'assert'
+import { popupWarningFix } from './helpers/popup-warning-fix'
+
+Feature('Serlo Editor - Blank exercise')
+
+Before(popupWarningFix)
+
+// TODO ideally import from
+// editor/src/types/editor-plugin-type.ts/EditorPluginType.FillInTheBlanksExercise
+const FillInTheBlanksExercise = 'blanksExercise'
```
hejtful: What is the benefit of having this constant?

packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
```diff
@@ -39,6 +46,14 @@ export function PluginToolbarTextControls({
...
           if (mathActive && !isMath(control)) return null
           if (blankActive && !isBlank(control)) return null
+          // To save space in the toolbar, we don't render the code editor
+          // within a fill-in-the-gap exercise
+          if (
+            hasFillInTheBlanks &&
+            control.name === TextEditorFormattingOption.code
+          ) {
+            return null
+          }
```
hejtful: Leftover?

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -138,20 +137,18 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
...
+              data-qa="plugin-exercise-check-answer-button"
+              className="serlo-button-blue mr-3 h-8"
+              onClick={() => {
+                checkAnswers()
+                setShowFeedback(true)
+              }}
+            >
+              {exStrings.check}
+            </button>
+          ) : null}
```
hejtful: Could you please elaborate on this change? I guess an invisible element was a problem for the tests, but why change the condition itself as well (leaving out `blanks.length > 0`)?

packages/editor/src/editor-ui/plugin-toolbar/text-controls/plugin-toolbar-text-controls.tsx
CodingDive: Nice catch! I think this slipped in from the other PR

e2e-tests/tests/450-blank-exercise.ts
CodingDive: ```I.click(`$add-exercise-${FillInTheBlanksExercise}`)``` is used throughout. That's why I wanted to import it, the plugin name is unlikely to change but I still prefer having a constant

e2e-tests/tests/450-blank-exercise.ts
CodingDive: Kind of! I wanted to see if there is a better way to select the text field than `I.click(locate('$plugin-text-editor').last())`

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
CodingDive: Nice catch! I mistakenly thought that `allBlanksHaveText` will evaluate to true when empty. As this is not the case, you prevented a bug here! :100:

e2e-tests/tests/450-blank-exercise.ts
hejtful: In that case, why not create a constant for the button and just have `I.click(addBlanksExerciseButton)` instead?

CodingDive:
@hejtful even with the fix in https://github.com/serlo/frontend/pull/3284/files this one annoying test is still failing. Changed `createRef` to `useRef` but with or without timeout, it doesn't reliably pass (and cursor doesn't appear in the first gap in the browser). Could you please take another look? :pray: 

`<div role="textbox" aria-multiline="true" class="outline-none [&amp;>[data-slate-node]]:mx-side" data-qa="plugin-text-editor" data-slate-editor="true" data-slate-node="value" contenteditable="true" ... ` is still the element that receives focus

Once the PR and the tests look good, feel free to merge it before we run into more merge conflicts :grinning:

hejtful:
@CodingDive I'd just leave that scenario as `Scenario.todo` for now then.

hejtful (approved):
Looks good, thanks for adding the tests!

packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
```diff
@@ -23,17 +23,11 @@ export function BlankRenderer({ element }: BlankRendererProps) {
...
     // Focus input when the blank is added
     setTimeout(() => inputRef.current?.focus())
-
-    // Editor gets refocused when the blank is removed from within
-    // text-controls/utils/blank.ts as it leads to slate errors on unmount.
-
-    // Only run on mount
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [inputRef])
```
hejtful, line 30: Sorry, just noticed this. Will it still only run on mount, even though it has this ref in dep array now?

CodingDive: This is needed because the ref previously was not defined on the first render as it's being passed to a child component where it attaches to the DOM element. Does this have negative consequences on unmount?

CodingDive: So it does run twice but doesn't do anything the first time around.

hejtful: >This is needed because the ref previously was not defined on the first render as it's being passed to a child component where it attaches to the DOM element. Does this have negative consequences on unmount?

No idea.

>So it does run twice but doesn't do anything the first time around.

Alright then, sounds good.

PR #3277, opened 2024-01-15
feat(editor): mvp for copy&pasting plugins

elbotho:
## [Demo](https://frontend-git-feat-editor-plugin-copy-tool-serlo.vercel.app/entity/repository/add-revision/277232)

- copy state to clipboard from any plugin toolbar
- paste state into any text plugin (if the copied plugin is valid there)

### DIFF
--- a/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
+++ b/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
@@ -38,7 +38,7 @@ export function PluginCopyTool({ pluginId }: PluginCopyToolProps) {
   return (
     <DropdownButton
       onClick={handleOnClick}
-      label="Copy plugin to clipboard"
+      label={editorStrings.edtrIo.pluginCopyButtonLabel}
       icon={faCopy}
       className="mt-2.5 border-t pt-2.5"
       dataQa="plugin-copy-tool-button"
--- a/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
+++ b/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
@@ -2,7 +2,7 @@ import { selectStaticDocumentWithoutIds, store } from '@editor/store'
 import { faCopy } from '@fortawesome/free-solid-svg-icons'
 import { shouldUseFeature } from '@serlo/frontend/src/components/user/profile-experimental'
 import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
-// import { useEditorStrings } from '@serlo/frontend/src/contexts/logged-in-data-context'
+import { useEditorStrings } from '@serlo/frontend/src/contexts/logged-in-data-context'
 import { showToastNotice } from '@serlo/frontend/src/helper/show-toast-notice'
 import { useCallback } from 'react'
 
@@ -15,7 +15,7 @@ interface PluginCopyToolProps {
  * experimental plugin to copy plugin's editor state to the clipboard
  */
 export function PluginCopyTool({ pluginId }: PluginCopyToolProps) {
-  // const editorStrings = useEditorStrings()
+  const editorStrings = useEditorStrings()
   const { strings } = useInstanceData()
 
   const handleOnClick = useCallback(() => {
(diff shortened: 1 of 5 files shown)


packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const document = selectStaticDocumentWithoutIds(store.getState(), pluginId)
+    const rowsDocument = { plugin: 'rows', state: [document] }
+
+    void navigator.clipboard.writeText(JSON.stringify(rowsDocument))
+    showToastNotice(strings.share.copySuccess, 'success', 2000)
+    showToastNotice(
+      '👉 ' + editorStrings.edtrIo.pluginCopyInfo,
+      undefined,
+      4000
+    )
```
hejtful, line 31: Interesting approach with consequent toasts, is this something that generally works well UX-wise?

packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }, [pluginId, strings, editorStrings])
+
+  if (!navigator.clipboard || !shouldUseFeature('editorPluginCopyTool')) {
+    return null
+  }
+
+  return (
+    <DropdownButton
+      onClick={handleOnClick}
+      label="Copy plugin to clipboard"
```
hejtful: Since you added the toast strings to data, maybe this one as well?

packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-default-tools.tsx
```diff
@@ -82,6 +83,7 @@ export function PluginDefaultTools({ pluginId }: PluginDefaultToolsProps) {
         icon={faTrashAlt}
         dataQa="remove-plugin-button"
       />
+      <PluginCopyTool pluginId={pluginId} />
```
hejtful, line 86: So clean 😁

packages/editor/src/plugins/text/hooks/use-editable-paste-handler.tsx
```diff
@@ -47,6 +51,29 @@ export const useEditablePasteHandler = (args: UseEditablePasteHandlerArgs) => {
         }
       }
 
+      // pasting editor document string and insert as plugins
+      if (
+        shouldUseFeature('editorPluginCopyTool') &&
+        !media &&
```
hejtful, line 48: This line alone makes me think that it's time to split this hook into smaller functions, I'll try to think of a clean way to do it.

hejtful (approved):
Wow, so many reused things, makes you think we're not that bad at thinking ahead! Great job and amazing how quickly this went from a reminder in the weekly to a smoothly working MVP 🤩 

Didn't detect any bugs while pasting in newly created Text plugin, already existing empty Text plugins, Text plugins with content (split + append). Copying and pasting nested plugins also works, like a Spoiler with a Multimedia plugin inside it, or a Table plugin with an Image plugin inside a cell.

packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
elbotho: well… tbh. it's not great. We only fall back to it for power-user features so far.

packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/plugin-copy-tool.tsx
elbotho: thanks! forgot about that one.

packages/editor/src/plugins/text/hooks/use-editable-paste-handler.tsx
elbotho: this is my main concern with this PR right now (but also with other code): Where does stuff like this belong?
It's part of the text plugin, but also related to the rows plugin, could be a standalone plugin, could be a helper …

elbotho: custom hook is probably a great start (and probably end) though ;)

hejtful: I'd keep using this hook, but just extract the actual handling of different things into separate handler functions (which can still be in this same file). Then the root function (namely, the hook) only serves as a kind of switch statement, determining which handler needs to be called.

PR #3283, opened 2024-01-16
feat(blank): check drag and drop answers

hejtful:
The list of commits shows what's done, but to expand on it:
- Uncommented drag-and-drop code
- Adjusted and completed drag-and-drop solution-checking code from [Lars' branch](https://github.com/serlo/frontend/tree/blanks-exercise-drag-and-drop)
- Fixed styling of drag-and-drop solution that has no text + some other minor styling adjustments
- Extracted `SolutionFeedback` from `ScMcExercise`, as it's not used just there
- Unified `BlanksExercise` types, as they were also defined in Text plugin
- Some renaming to match component names to their respective file names

Questions: 
- @LarsTheGlidingSquirrel @elbotho Any ideas on how to further split up the `FillInTheBlanksRenderer` component?

### DIFF
--- a/packages/editor/src/editor-ui/exercises/solution-feedback.tsx
+++ b/packages/editor/src/editor-ui/exercises/solution-feedback.tsx
@@ -1,13 +1,15 @@
 import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
 import type { ReactNode } from 'react'
 
-export interface FeedbackProps {
+export interface SolutionFeedbackProps {
   correct: boolean
   children?: ReactNode
   missedSome?: boolean
 }
 
-export function Feedback({ children, correct, missedSome }: FeedbackProps) {
+export function SolutionFeedback(props: SolutionFeedbackProps) {
+  const { children, correct, missedSome } = props
+
   const exStrings = useInstanceData().strings.content.exercises
   const fallbackString =
     exStrings[correct ? 'correct' : missedSome ? 'missedSome' : 'wrong']
--- a/packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-check-button.tsx
+++ b/packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-check-button.tsx
@@ -1,0 +1,43 @@
+import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
+import { useState } from 'react'
+
+// TODO: Shouldn't import this from another plugin
+import { Feedback } from '../../sc-mc-exercise/renderer/feedback'
+import { cn } from '@/helper/cn'
+
+interface BlankCheckButtonProps {
+  isVisible: boolean
+  feedback: Map<string, { isCorrect?: boolean | undefined }>
+  onClick: () => void
+}
+
+export function BlankCheckButton(props: BlankCheckButtonProps) {
+  const { isVisible, feedback, onClick } = props
+
+  const exStrings = useInstanceData().strings.content.exercises
+
+  // Used to show feedback when user clicked "Stimmts?" button
+  const [showFeedback, setShowFeedback] = useState<boolean>(false)
+
+  return (
+    <div className="mt-2 flex">
+      <button
+        className={cn(
+          'serlo-button-blue mr-3 h-8',
+          isVisible ? '' : 'pointer-events-none opacity-0'
+        )}
+        onClick={() => {
+          onClick()
+          setShowFeedback(true)
+        }}
+      >
+        {exStrings.check}
+      </button>
+      {showFeedback && (
+        <Feedback
+          correct={[...feedback].every((entry) => entry[1].isCorrect)}
+        />
+      )}
+    </div>
+  )
+}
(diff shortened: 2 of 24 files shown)


packages/editor/src/editor-ui/exercises/solution-feedback.tsx
```diff
@@ -0,0 +1,33 @@
+import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
+import type { ReactNode } from 'react'
+
+export interface SolutionFeedbackProps {
+  correct: boolean
+  children?: ReactNode
+  missedSome?: boolean
+}
+
+export function SolutionFeedback(props: SolutionFeedbackProps) {
```
elbotho: Solution is usually what we call the `sample solution` authors provide. So to avoid ambiguity, I'd like to avoid it here. Maybe `Feedback` or `ExerciseFeedback`?

hejtful: `ExerciseFeedback` sounds great, way better than `SolutionFeedback` even without the `sample solution` context 😁

elbotho:
> Any ideas on how to further split up the FillInTheBlanksRenderer component?

Nothing super obvious you did not think about ^^
We could move some utils / functions out, remove the debug code…? The only big thing would be to separate the drag and drop code more, but I'm not really sure how much cleaner the result would be.

elbotho (approved):
thanks for making it work 🎉

LarsTheGlidingSquirrel:
I noticed a small issue: The red/green outline persists when moving the draggable. 
Steps to reproduce: 
1. Check answers
image
2. Move one of them out -> red outline stays
image
3. Move to the correct spot -> red outline stays
image

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-check-button.tsx
```diff
@@ -0,0 +1,33 @@
+import { ExerciseFeedback } from '@editor/editor-ui/exercises/exercise-feedback'
+import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
+
+import { cn } from '@/helper/cn'
+
+interface BlankCheckButtonProps {
+  isVisible: boolean
+  feedback: Map<string, { isCorrect?: boolean | undefined }>
```
LarsTheGlidingSquirrel, line 8: As discussed, for consistency, it would be better to have either `Map<string, ...` or `Map<BlankId, ...` everywhere. I noticed that even if using `Map<BlankId, ...` VS code will still show me `Map<string, ...` as a type. So I think I favor `Map<string, ...` now.

elbotho:
> I noticed a small issue: The red/green outline persists when moving the draggable.

(same for typing mode I think)

packages/editor/src/plugins/fill-in-the-blanks-exercise/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  blankId: t.string,
+  correctAnswers: t.array(Answer),
+  // Here we could specify incorrect answers for this blank with specific learner feedback
+  // incorrectAnswers?: Answer[]
+  // Here we could add a default feedback for the learner
+  // defaultIncorrectAnswerFeedback: string
+})
+// BlankType is used internally within the BlanksExercise, in order to satisfy io-ts.
+// BlankInterface is used within Slate, as io-ts can't convert CustomText interface.
+// A long term solution would be to switch Text plugin's types/interfaces to io-ts completely.
```
LarsTheGlidingSquirrel, line 19: Let's see how we can clean this up later. Thanks for trying to remove this redundancy.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -78,156 +57,171 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
   }, [blanks, textUserTypedIntoBlanks, initialTextInBlank])
 
   const draggables = useMemo(() => {
-    return blanks.map(({ blankId, correctAnswers }) => ({
+    const sorted = blanks.map(({ blankId, correctAnswers }) => ({
       draggableId: `solution-${blankId}`,
       text: correctAnswers[0].answer,
     }))
-  }, [blanks])
+    if (isEditing) return sorted
```
LarsTheGlidingSquirrel, line 64: Nice that you kept it sorted in edit mode.

LarsTheGlidingSquirrel (approved):
Looks great! Nice that the "check answer" button is now a shared component between all exercise types. Just added some minor comments.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Tbh, just asked Gregor what the right way is 😁

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-check-button.tsx
hejtful: I just copy/pasted this one around, it was actually like this when I started working on it. But yeah, we can try to make it consistent all around in some following PR. I'd even try switching maps to objects, just to see if the code gets noticeably smaller.

PR #3284, opened 2024-01-16
fix(blank): focus input when added to empty text plugin

hejtful:
The fix for a failing test in https://github.com/serlo/frontend/pull/3276

### DIFF
--- a/packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
+++ b/packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
@@ -27,7 +27,10 @@ export function BlankRenderer({ element }: BlankRendererProps) {
   useEffect(() => {
     // Focus input when the blank is added
     const input = inputRef.current
-    if (input) input.focus()
+    if (!input) return
+    setTimeout(() => {
+      input.focus()
+    })
 
     // Editor gets refocused when the blank is removed from within
     // text-controls/utils/blank.ts as it leads to slate errors on unmount.
--- a/packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
+++ b/packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
@@ -26,7 +26,6 @@ export function BlankRenderer({ element }: BlankRendererProps) {
   const inputRef = createRef<HTMLInputElement>()
   useEffect(() => {
     // Focus input when the blank is added
-    const input = inputRef.current
     setTimeout(() => inputRef.current?.focus())
 
     // Editor gets refocused when the blank is removed from within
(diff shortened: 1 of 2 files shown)


CodingDive:
Very interesting! The typing works immediately, but weirdly enough, when the first gap is created the cursor is not visible. Only when I create a gap, delete it then create another one (see the end of the gif), the cursor gets visible.   


focus_fill_in_the_gap


You can see this by typing `document.activeElement` in the console.

image


Steps to reproduce: 

Go to http://localhost:3000/entity/create/Exercise/23869 => Fill in the blanks => Create gap without text 

Following element is focused (cursor is invisible, yet typing immediately makes the focus jump to the gap) `<div role="textbox" aria-multiline="true" class="outline-none [&amp;>[data-slate-node]]:mx-side" data-qa="plugin-text-editor" data-slate-editor="true" data-slate-node="value" ... `

Then you delete the gap (still no text), create a new one and you'll see that the correct gap input element gets focused.
`<input class="h-[25px] rounded-full border border-brand bg-brand-50" data-qa="blank-input" size="4" spellcheck="false" autocorrect="off" placeholder="" type="text" value="">` 


The UX is definitely better now but not seeing a cursor at all is kind of strange and makes you think that the fill in the gap plugin has lost focus entirely. I noticed it by running the test because it actually checks whether the blank input has focus. https://github.com/serlo/frontend/pull/3276/files#diff-104cb85978425664605bb4849f059a3fe1a58e7a10414a9210cd00e4d2048427R125 If it's too tricky to place the focus in the gap instead of the slate textbox, we can simply change the test implementation.

hejtful:
Thank you for the detailed report. I can't reproduce this locally, but luckily I could reproduce it on the deployed preview. Pushed a timeout fix, if that doesn't work, I'll push some console logs to try and figure out what the difference is between my local and your local / deployed preview.

hejtful:
@CodingDive Timeout seems to have fixed it (sadly 😞), could you please check again?

packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
```diff
@@ -27,7 +27,10 @@ export function BlankRenderer({ element }: BlankRendererProps) {
   useEffect(() => {
     // Focus input when the blank is added
     const input = inputRef.current
-    if (input) input.focus()
+    if (!input) return
+    setTimeout(() => {
+      input.focus()
+    })
```
elbotho:
```suggestion
    setTimeout(() => inputRef.current?.focus())
```
I love a good one liner, but that's just preference 😌

elbotho (approved):
Thank you!

packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
hejtful: Nice and clean! The only difference is that previously, setTimeout was not run if the `input` doesn't exist. It's a micro-optimization, though, can't see that slowing the app down. Do you prefer a micro-optimization or a one-liner in this case? 😄

CodingDive (approved):
Works great with the timeout 🎉🥲

packages/editor/src/plugins/fill-in-the-blanks-exercise/blank-renderer.tsx
elbotho: personally definitely one-liner but both are fine.

elbotho: if you can measure the performance difference you can keep it 😂

PR #3287, opened 2024-01-17
feat(plugin-image): handle pasting images in caption

elbotho:
main idea here is to improve the workflow for power authors.
now when pasting an image from clipboard or a valid image url in the caption the src of the parent image plugin gets updated.


elbotho:
@hejtful not a super clean separation but at least it's in the right plugin and has it's own file. What do you think?

hejtful (approved):
Felt motivated for refactoring after doing it with blanks, so opened a PR to your PR 😄 https://github.com/serlo/frontend/pull/3288 Hope you don't mind! Please feel free to close it and merge this one, if you don't like my suggestions!

PR #3288, opened 2024-01-17
refactor(plugin-image): clean captionPasteHandler

hejtful:
Changes:
- Shorten the name, both `on-` prefix and `-Handler` suffix are a bit much
- Skip check for document being Text plugin, it should always be when this is called
- Use early exits to reduce code complexity
- The returned value was not being used in any way, so omit last return and don't return any specific value
- More verbose comment
- Shorter code in general


elbotho (approved):
love it, thank you!

PR #3291, opened 2024-01-17
feat(plugin-solution): add suggestions to prerequisite link

elbotho:
## [Demo](https://frontend-git-feat-plugin-solution-add-link-suggestions-serlo.vercel.app/entity/create/Exercise/60730)

image
image
image

This is also to speed up the creation of the sample solutions.
Not the nicest code, but I could reuse a lot of existing components.

Caveats:
- Titles don't have to be entered manually any more (but this also means they can't be entered manually any more 😅)
- If a content is not in the quickbar data yet it can't be selected

Long term:
There are more plans to refactor / change the structure of exercises and solutions. If we have a migration anyway it probably makes sense to simplify this to only accept one serlo uuid (instead of id/title/alias).


packages/editor/src/plugins/solution/renderer.tsx
```diff
@@ -27,57 +25,72 @@ export function SolutionRenderer({
...
         {elementAfterToggle}
-      </div>
-      {solutionVisible && renderSolutionContent()}
+      </nav>
+      {visible ? (
+        <div className="serlo-solution-box">
+          {visible === 'solution' ? renderSolutionContent() : null}
+          {visible === 'strategy' && strategy ? strategy : null}
+        </div>
+      ) : null}
```
hejtful: If `visible === 'strategy' && strategy === null`, this `div` will be rendered empty. Is that the desired outcome?

packages/editor/src/plugins/solution/editor.tsx
```diff
@@ -37,58 +67,77 @@ export function SolutionEditor({ state, focused }: SolutionProps) {
...
+                value={prerequisite.defined ? `/${prerequisite.id.value}` : ''}
+                removeLink={() => {
+                  prerequisite.remove()
+                  setShowPrerequisiteLinkTool(false)
+                }}
+                quickbarData={quickbarData}
+              />
+            ) : (
+              <LinkOverlayEditMode
+                isSerloLinkSearchActive //
```
hejtful:
```suggestion
                isSerloLinkSearchActive
```

packages/editor/src/plugins/solution/editor.tsx
```diff
@@ -37,58 +67,77 @@ export function SolutionEditor({ state, focused }: SolutionProps) {
...
-              } else {
-                prerequisite.create({ id: '', title: value, alias: undefined })
-              }
-            }}
-            placeholder={solutionStrings.linkTitle}
-          />
-        </a>
-      </>
+                }}
+                removeLink={() => {}} //
```
hejtful: Should this really be just `() => {}`?

packages/editor/src/plugins/solution/editor.tsx
```diff
@@ -1,20 +1,50 @@
...
+        .then((fetchedData) => fetchedData && setQuickbarData(fetchedData))
+        // eslint-disable-next-line no-console
+        .catch(console.error)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [quickbarData])
+
+  useEffect(() => {
+    if (!focused) setShowPrerequisiteLinkTool(false)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
hejtful: Would be cool to explain why

packages/editor/src/plugins/solution/editor.tsx
```diff
@@ -1,20 +1,50 @@
...
+
+  useEffect(() => {
+    if (!isSerloLinkSearchActive) return
+    if (!quickbarData) {
+      fetchQuickbarData()
+        .then((fetchedData) => fetchedData && setQuickbarData(fetchedData))
+        // eslint-disable-next-line no-console
+        .catch(console.error)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
hejtful: Would be cool to explain why

hejtful (approved):
Some uncertainties, but nothing critical, can be merged imo.

Entkenntnis:
@elbotho merging staging into this branch will make the PR cleaner (currently changes from solution/strategy are also shown) :)

elbotho:
> @elbotho merging staging into this branch will make the PR cleaner (currently changes from solution/strategy are also shown) :)

done!

packages/editor/src/plugins/solution/renderer.tsx
elbotho: You are correct, but if `strategy === null` the toggle is not shown so this can't happen.
Do you have a simple idea to make this cleaner?
I can't think of something right now where I think it's worth the extra complexity.

I will simplify `{visible === 'strategy' && strategy ? strategy : null}` to `{visible === 'strategy' ? strategy : null}` because it does the same.

packages/editor/src/plugins/solution/editor.tsx
elbotho: Jup, because there is no link defined at this point so I can't delete it.
I'll make `removeLink` optional to get rid of this weird prop.

packages/editor/src/plugins/solution/editor.tsx
elbotho: thanks, actually not needed.

elbotho:
@Entkenntnis ah… I think I merged the solution/strategy toggles into this branch to avoid conflicts, they are not actually in staging yet.

packages/editor/src/plugins/solution/editor.tsx
elbotho: thanks, updated the array instead.

elbotho:
thank you @hejtful for the feedback!

packages/editor/src/plugins/solution/renderer.tsx
hejtful: Nah, sounds good, I didn't get that context when commenting, thanks for explaining.

Entkenntnis:
@elbotho Check, I've updated staging to reflect the current status

PR #3293, opened 2024-01-18
experiment(plugin-image): experiment with direct upload

elbotho:


packages/editor/src/plugins/rows/editor-renderer.tsx
```diff
@@ -149,11 +149,11 @@ export function EditorRowRenderer({
         case NativeTypes.URL: {
           const urls: string[] = monitor.getItem<{ urls: string[] }>().urls
           const text = urls[0]
-          plugins.find(({ type, plugin }) => {
+          plugins.find(async ({ type, plugin }) => {
```
Entkenntnis: not 100% if `plugins.find` can properly handle the aync and return the awaited value (most likely not - but we don't actually use the return, so it's fine here)

I would suggest adding a comment here.

Entkenntnis (approved):

packages/editor/src/plugins/rows/editor-renderer.tsx
elbotho: good catch!

elbotho: what do you think of this refactor?
https://github.com/serlo/frontend/pull/3293/commits/c1bb107a0d4b713c4c6f8fffe0faeadb3d062845

Entkenntnis: thank you, the refactor makes the change of behaviour clearer

PR #3305, opened 2024-01-24
feat(plugin-video): allow timestamp in youtube embeds

elbotho:


Entkenntnis (approved):
interesting, that query parameters differ between url and embed - otherwise nice improvement

PR #3307, opened 2024-01-24
feat(plugin-solution): split solution and strategy, add extra toggle

elbotho:
### [Editor-Demo](https://frontend-git-feat-plugin-solution-toggle-for-strategy-serlo.vercel.app/entity/repository/add-revision/286542)

Wish by U_Content and feedback is positive.


Entkenntnis (approved):
Looks good and nice that we don't need to change editor state.

you can merge into staging and update crowdin

PR #3309, opened 2024-01-27
feat(blank): drag-and-drop dummy answers

hejtful:


apps/web/src/data/en/index.ts
```diff
@@ -925,6 +925,8 @@ export const loggedInData = {
           previewMode: 'Preview',
           previewIsActiveHint: 'Edit the exercise again',
           previewIsDeactiveHint: 'Preview the exercise',
+          dummyAnswers: 'Dummy answers',
+          addDummyAnswer: 'Add a dummy answer',
```
hejtful: Strings from the design are "Falsche Antwortmöglichkeiten" and "falsche Antwort hinzufügen" respectively. If this English translation is not optimal, please suggest a better one 🙏

elbotho: I like `Incorrect Answers` a bit better, but dummy also works

apps/web/shared/tailwind-base.js
```diff
@@ -63,6 +63,10 @@ export default {
         red: {
           DEFAULT: colors.red[900],
           100: colors.red[100],
+          // This color is present in the design for fill-in-the-blanks exercise,
+          // dummy drag-and-drop answers feature. For now, it's just a temporary
+          // variable used in staging testing.
+          temp: '#ff8686',
```
elbotho: please use `red-400` and see if someone notices ;)

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-preview.tsx
```diff
@@ -0,0 +1,65 @@
...
+        onClick={onClick}
+      >
+        <span className={cn(isInHoverMode && 'opacity-20')}>{text}</span>
+        <span
+          className={cn(
+            'absolute bottom-0 left-0 right-0 top-0 text-center',
+            !isInHoverMode && 'opacity-0'
+          )}
+        >
+          <FaIcon icon={faPencil} className="text-sm" />
```
elbotho: Sorry this is probably design feedback (but also code simplification).
I don't like pencil icon on hover. 
It's more confusing than helpful for me from an UX perspective.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
```diff
@@ -0,0 +1,58 @@
...
+            text={answer}
+            onChange={(event) => {
+              handleDummyAnswerChange(event.target.value, index)
+            }}
+            onRemove={() => {
+              handleRemoveDummyAnswer(index)
+            }}
+          />
+        ))}
+      </div>
```
elbotho: I'd like to only see the button (not the title and extra space) if there are no additional answers inserted yet.

apps/web/src/data/en/index.ts
elbotho: Or maybe even `Extra incorrect answers` for the title

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-preview.tsx
elbotho: (I suggest not having any extra styling on the input on hover)

elbotho: And instead of the trash button that is popping out I'd prefer to just have a permanent `x` icon on the right side of the input (but inside) as a delete button. (less layout shift, less unexpectedness)

(rough draft just for illustration :)
image

packages/editor/src/plugins/fill-in-the-blanks-exercise/static.tsx
```diff
@@ -14,6 +14,7 @@ export function FillInTheBlanksStaticRenderer(
       textPluginState={text}
       mode={state.state.mode as FillInTheBlanksMode}
       initialTextInBlank="empty"
+      additionalDraggableAnswers={state.state.additionalDraggableAnswers}
```
elbotho: Not directly related but the `state.state` would problably nicer with destucuring in L8.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -137,6 +153,14 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
           </BlankDraggableArea>
         ) : null}
 
+        {mode === 'drag-and-drop' && isEditing ? (
+          <BlankDraggableDummyArea
+            additionalDraggableAnswers={
+              additionalDraggableAnswers as FillInTheBlanksExerciseProps['state']['additionalDraggableAnswers']
```
elbotho: Interesting, why is this type assertion needed?
Maybe at least a local type in the top for `FillInTheBlanksExerciseProps['state']['additionalDraggableAnswers']` would be nice, since we need it two times.

packages/editor/src/plugins/fill-in-the-blanks-exercise/index.tsx
```diff
@@ -45,8 +46,7 @@ function createState() {
       },
     }),
     mode: string(defaultMode),
-    // This could be used to store additional answers for mode drag-and-drop in the future
-    // additionalDraggableAnswers: optional(list(object({ answer: string() }))),
+    additionalDraggableAnswers: list(object({ answer: string() })),
```
elbotho:
```suggestion
    extraDraggableAnswers: list(object({ answer: string() })),
```

I'd like this naming better, since I read additional as "more of the same" and extra as "more of something special".

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface BlankDraggableDummyAreaProps {
+  additionalDraggableAnswers: FillInTheBlanksExerciseProps['state']['additionalDraggableAnswers']
+}
+
+export function BlankDraggableDummyArea(props: BlankDraggableDummyAreaProps) {
+  const { additionalDraggableAnswers } = props
+
+  const blanksExerciseStrings = useEditorStrings().plugins.blanksExercise
+
+  const dummyDraggables = additionalDraggableAnswers.map(
```
elbotho: maybe `dummyValues` would be clearer here? I was expecting something more complex when reading `dummyDraggables`.

elbotho (changes requested):

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Because of the difference between editor and static plugin props.

I'd stick to the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), but if you insist, I can add the local type.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-preview.tsx
hejtful: Do you mind giving this feedback to Gregor? I'm happy to make the change, but it wouldn't be nice to change his design without asking.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
hejtful: Extra space removed, thanks for pointing that out. The title is there in the design, even without any extra incorrect answers, so it would be great if you could pass that feedback to Gregor as well.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Btw, if you have any ideas on how to avoid the type assertion, I'm all ears! 😁

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-preview.tsx
hejtful: Implemented this suggestion and some other minor fixes in https://github.com/serlo/frontend/pull/3310

Will share it with Gregor as well and see what he thinks.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
hejtful: Implemented this suggestion and some other minor fixes in https://github.com/serlo/frontend/pull/3310

Will share it with Gregor as well and see what he thinks.

elbotho (approved):

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { cn } from '@/helper/cn'
+
+interface BlankDraggableDummyAnswerProps {
+  text: string
+  onChange: ChangeEventHandler<HTMLInputElement>
+  onKeyDown: KeyboardEventHandler
+  onRemoveClick: MouseEventHandler
+}
+
+export function BlankDraggableDummyAnswer(
```
LarsTheGlidingSquirrel, line 20: Maybe rename to `ExtraDragAndDropAnswer`? By the name I would have expected it to be a draggable element mixed with the other draggable elements.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
```diff
@@ -0,0 +1,53 @@
...
+  props: BlankDraggableDummyAnswerProps
+) {
+  const { text, onChange, onKeyDown, onRemoveClick } = props
+  const blanksExerciseStrings = useEditorStrings().plugins.blanksExercise
+
+  return (
+    <div className="relative">
+      <span className="serlo-autogrow-input mb-1 mr-2" data-value={text + '_ '}>
```
LarsTheGlidingSquirrel, line 28: Why is the `+ '_' ` needed?

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -137,6 +153,14 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
...
         ) : null}
 
+        {mode === 'drag-and-drop' && isEditing ? (
+          <BlankDraggableDummyArea
+            extraDraggableAnswers={
+              extraDraggableAnswers as FillInTheBlanksExerciseProps['state']['extraDraggableAnswers']
+            }
+          />
+        ) : null}
+
```
LarsTheGlidingSquirrel, line 167: Hm, here the renderer component is doing stuff that feels like it should instead be done by the editor component. Can we move this into the `editor.tsx`?

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
```diff
@@ -0,0 +1,62 @@
...
+                  if (e.key === 'Enter') {
+                    extraDraggableAnswers.insert()
+                  }
+                }}
+                onRemoveClick={() => {
+                  extraDraggableAnswers.remove(index)
+                  // focus new last input to make sure we don't loose focus
+                  const allInputs = areaRef.current?.querySelectorAll('input')
+                  if (allInputs) {
+                    void [...allInputs]?.at(-2)?.focus()
```
LarsTheGlidingSquirrel, line 44: How does this `at(-2)` work?

LarsTheGlidingSquirrel (approved):
Looks great and seems to work fine! Thank you! 
A few comments about the code. Also, I think the editor should show the extra draggable answers immediately in the list of answers. Not only in preview. But that is also a question for UX.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Could you please elaborate?

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
hejtful: Because without it the autogrow input becomes too small for its content

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
hejtful: Tbh, it's a part @elbotho added, maybe he can chip in 😁

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
hejtful: Good point, we can improve this in a next PR

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
LarsTheGlidingSquirrel: As far as I understood it, the idea behind the split into `editor.tsx`, `renderer.tsx` and `static.tsx` is that ...
- `editor.tsx` renders and provides all editor functionality
- `static.tsx` handles the static rendering view
- `renderer.tsx` renders the plugin based on its state and is used by both `editor.tsx` and `static.tsx`.

So, using that logic anything that should only be available in the editor should live in `editor.tsx`.

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
LarsTheGlidingSquirrel: Ah, I get it! Is this the case everywhere we use the `serlo-autogrow-input`? Or only here?

hejtful: Everywhere, I think

elbotho: Everywhere. Only we use `'_ '` here to have space for the delete button 🙈

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Ah, good point!

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
elbotho: It's the nicest way I know to get the second to last entry in an array. I'm not getting the last one, because that's the one about to be deleted.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

elbotho: I considered writing a complex css query, but decided this was easier to understand. Maybe I was wrong :)

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-answer.tsx
elbotho: _(Also `'_'` is a cool guy, obviously.)_

LarsTheGlidingSquirrel: All right! Thanks! :)

packages/editor/src/plugins/fill-in-the-blanks-exercise/components/blank-draggable-dummy-area.tsx
LarsTheGlidingSquirrel: Makes sense now! Thanks! :)

PR #3331, opened 2024-02-01
fix(plugin-exercise): hide empty content in static view to avoid unexpected margins

elbotho:
### [Demo](https://frontend-git-hide-empty-content-in-exercises-serlo.vercel.app/294720)


hejtful (approved):
Neat

PR #3332, opened 2024-02-01
feat(blank): replace filled blank on drop

hejtful:
- replace a filled blank with a draggable from the area
- swap a filled blank with another filled blank
- reset feedback for replace/swap
- some styling changes to fit replace/swap functionality


elbotho (approved):
love it, thank you!

PR #3336, opened 2024-02-02
refactor(editor): SlateOverlay reusable component

hejtful:
Merge overlays from Text plugin link controls, Text plugin suggestions and BlankExercise plugin blank controls into one reusable SlateOverlay component.

Note: The positioning of Text plugin suggestions is slightly different and has a bit lighter shadow.


packages/editor/src/editor-ui/slate-hover-overlay.tsx
```diff
@@ -1,15 +0,0 @@
-import { useSlate } from 'slate-react'
-
-import { HoverOverlay, type HoverOverlayProps } from './hover-overlay'
```
elbotho, line 3: pretty sure `HoverOverlay` can then also be removed

packages/editor/src/editor-ui/slate-overlay.tsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        {children}
+      </div>
+    </div>
+  )
+}
+
+// If provided an anchor element, returns its size and position (DOMRect).
+// Otherwise, checks for native DOM selection, and provides a DOMRect based on it.
+function getAnchorRect(editor: ReactEditor, anchor: CustomElement | undefined) {
```
elbotho, line 56: could be
```suggestion
function getAnchorRect(editor: ReactEditor, anchor?: CustomElement) {
```

elbotho (approved):
thank you!

packages/editor/src/editor-ui/slate-overlay.tsx
hejtful: I did write it like that the first time, but then it looks like that argument can be omitted entirely. In this case, I'd like the function signature to point out that this argument is expected to be passed, but the passed variable can be of value `undefined`. Hope that makes sense?

elbotho: > Hope that makes sense?

With your explanation: yes! 
In the wild I'd probably still read it as a more verbose version of `anchor?:…` 
If you want to make it very clear I'd probably go with `CustomElement | null` or something like that.

In this case: I don't mind either way :)

elbotho:
@hejtful merged this one only half intentionally.
Okay for you? :)

hejtful:
Fine with me :)

PR #3340, opened 2024-02-03
fix(blank): don't close blank controls overlay on triple click

hejtful:
Slate seems to have some specific handler that changes `editor.selection` on triple click. Removing the `isCollapsed` condition in overlay anchor-setting logic fixes this bug.


elbotho (commented):
Nice, thanks for the investigation.
While I love the "less code less bugs" approach here the `isCollapsed` check prevents the popover to open when e.g. more than one gap is selected or the gap and some other text.
I like that as a way to prevent unexpected behaviours.

hejtful:
Good point, I'll open a new PR with the "stop propagation on triple click" fix.

PR #3342, opened 2024-02-05
fix(blank): prevent blank controls close on triple click

hejtful:
Slate seems to have some specific handler that changes `editor.selection` on triple click. Here, we stop the click event being propagated to Slate, in case of triple clicks.


elbotho (approved):
❤️

PR #3343, opened 2024-02-05
fix(plugin-article): fix deleting sources and restore order function

elbotho:
for https://trello.com/c/x5XqNNGd/544-deleting-sources


Entkenntnis (approved):
thank you for the fix, approach is fine, basically rerendering on every change.

with some more thoughts, maybe a more "react" solution can be found, but not worth the effort right now.

PR #3347, opened 2024-02-05
refactor(plugin-article): rework sources, remove old input components

elbotho:
no more slate, no more hacks need 🤔

Overall the slate-input had some weird behaviour. Let's hope this version behaves better.


hejtful:
A really nice thing about this is that now the `HoverOverlay` component can be removed as well, as its only remaining use case was in `InlineSettings`.

elbotho:
> A really nice thing about this is that now the HoverOverlay component can be removed as well, as its only remaining use case was in InlineSettings.

Thank you for the reminder, that's why I started this PR in the first place 😅

hejtful (approved):
Nice!

PR #3348, opened 2024-02-06
refactor(editor): replace algebra.js with mathjs

hejtful:
Reasons:
- `algebra.js` throws `ReferenceError: assignment to undeclared variable Term` when the editor is used as an NPM package in edusharing
- `algebra.js` is in public archive and not maintained for several years now


kulla (approved):
Thx :heart:

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -237,10 +236,9 @@ export function FillInTheBlanksRenderer(props: FillInTheBlanksRendererProps) {
         if (!solution || !submission) return false
 
         // Both submission and solution are valid mathematical expressions.
-        // Using algebra.js, subtract the submission from the solution, and
-        // if the result of the subtraction is 0, submission and solution are
-        // mathematical equivalents, therefore the submission is correct.
-        return solution.subtract(submission).toString() === '0'
+        // Subtract the submission from the solution. If the result is 0,
+        // submission and solution are mathematical equivalents.
+        return solution - submission === 0
```
LarsTheGlidingSquirrel, line 241: Does this behave different to https://mathjs.org/docs/reference/functions/symbolicEqual.html and why did you choose to use this?

LarsTheGlidingSquirrel (approved):
Looks good! :)

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
Entkenntnis: This was the behaviour from before. Symbolic equal is even stronger, we may want to use it in future (but not for now)

PR #3350, opened 2024-02-08
feat(packages/editor): Expose `SerloRenderer` for editor integrations

LarsTheGlidingSquirrel:
Problem: In the RLP editor integration `instanceData` and `loggedInData` are not provided in static renderer view leading to an error. 

For the editor view the editor package exposes `SerloEditor` that internally provides these. 

Solution: Now, the editor package exposes `SerloRenderer` that pretty much does the exact same thing for the static renderer view.

For the future: Maybe change the editor API so that things like language can be configured. For now there is only German.

### DIFF
--- a/packages/editor/src/package/serlo-renderer.tsx
+++ b/packages/editor/src/package/serlo-renderer.tsx
@@ -1,0 +1,29 @@
+import { StaticRenderer } from '@editor/static-renderer/static-renderer'
+import type { AnyEditorDocument } from '@editor/types/editor-plugins'
+
+import { InstanceDataProvider } from '@/contexts/instance-context'
+import { LoggedInDataProvider } from '@/contexts/logged-in-data-context'
+import type { InstanceData, LoggedInData } from '@/data-types'
+
+export interface SerloRendererProps {
+  instanceData: InstanceData
+  loggedInData: LoggedInData
+  documentState: AnyEditorDocument | AnyEditorDocument[]
+}
+
+export function SerloRenderer({
+  instanceData,
+  loggedInData,
+  documentState,
+  ...props
+}: SerloRendererProps) {
+  return (
+    <InstanceDataProvider value={instanceData}>
+      <LoggedInDataProvider value={loggedInData}>
+        <div className="serlo-editor-hacks">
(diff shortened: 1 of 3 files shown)


packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { AnyEditorDocument } from '@editor/types/editor-plugins'
+
+import { InstanceDataProvider } from '@/contexts/instance-context'
+import { LoggedInDataProvider } from '@/contexts/logged-in-data-context'
+import type { InstanceData, LoggedInData } from '@/data-types'
+
+export interface SerloRendererProps {
+  instanceData: InstanceData
+  loggedInData: LoggedInData
+  documentState: AnyEditorDocument | AnyEditorDocument[]
```
elbotho: calling it `documentState` complicates things below. 
I'd just go with `document` (since state can mean so many things it's basically meaningless :).
But you can also merge it as it is.

packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function SerloRenderer({
+  instanceData,
+  loggedInData,
+  documentState,
```
elbotho:
```suggestion
```

packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -0,0 +1,29 @@
...
+  ...props
+}: SerloRendererProps) {
+  return (
+    <InstanceDataProvider value={instanceData}>
+      <LoggedInDataProvider value={loggedInData}>
+        <div className="serlo-editor-hacks">
+          <StaticRenderer document={documentState} {...props} />
```
elbotho:
```suggestion
          <StaticRenderer {...props} />
```

elbotho (approved):

packages/editor/src/package/serlo-renderer.tsx
LarsTheGlidingSquirrel: Good point! I like `document` more as well. Changed!

PR #3353, opened 2024-02-09
feat(plugin-injection): select single exercise from group via hash

elbotho:


Entkenntnis (approved):
Thank you, logic looks good and this should keep analytics compatible 👍 

One question: the code currently don't handle the case of new exercise groups, right? E.g. there is no logic to provide new grouped exercises with uuids - we need some other solution.

For the analytics side: entityId is stored as number, so no luck with appending a hash. But currently revisionId is unused, we can probably reuse this column.

PR #3366, opened 2024-02-12
fix(plugin-text): suggestions - fix timing of deleting previous block

hejtful:
This fixes the [disappearing list bug](https://trello.com/c/3fQ4mANW/563-spoiler-after-bulletpoint-list), but the focus issue is way deeper and harder to solve ([this is the problematic line](https://github.com/serlo/frontend/blob/staging/packages/editor/src/core/sub-document/editor.tsx#L67)), so it's out of scope of this bug.

In addition, opening suggestions with "/" while in a list is disabled in this PR.

E2E tests for both cases added.

### DIFF
--- a/packages/editor/src/plugins/text/hooks/use-suggestions.tsx
+++ b/packages/editor/src/plugins/text/hooks/use-suggestions.tsx
@@ -165,9 +165,11 @@ export const useSuggestions = (args: useSuggestionsArgs) => {
       return
     }
 
-    // split the text-plugin and insert selected new plugin
+    // Split the text-plugin and insert selected new plugin,
+    // then remove the leftover empty line.
     setTimeout(() => {
       insertPlugin({ pluginType, editor, id, dispatch })
+      editor.deleteBackward('block')
     })
   }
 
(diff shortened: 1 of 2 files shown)


elbotho:
Without the line it results in an unnecessary empty line above the inserted content (tested in chrome.
That's probably why I added it.

image

hejtful:
Alright, then I will try to fix the bug by wrapping this line in a "is the previous line a list?" condition.

Edit: changing the timing of the deletion of the empty line seems to work.

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -165,11 +165,11 @@ export const useSuggestions = (args: useSuggestionsArgs) => {
...
 
-    editor.deleteBackward('block')
-
-    // split the text-plugin and insert selected new plugin
+    // Split the text-plugin and insert selected new plugin,
+    // then remove the leftover empty line.
     setTimeout(() => {
       insertPlugin({ pluginType, editor, id, dispatch })
+      editor.deleteBackward('block')
     })
```
elbotho, line 173: just checking: the setTimeout still improves something, right?

elbotho (approved):

PR #3370, opened 2024-02-12
fix(dnd): Add touch support to react-dnd

### LINKED ISSUE #3369: Mobile / touch support for fill in the gap with drag & drop
As reported by Quinn in Slack

CodingDive:
Fixes https://github.com/serlo/frontend/issues/3369, #3379


packages/editor/src/core/components/dnd-wrapper.tsx
```diff
@@ -1,14 +1,29 @@
...
+import { ReactNode, useEffect, useState } from 'react'
 import { DndProvider } from 'react-dnd'
 import { HTML5Backend } from 'react-dnd-html5-backend'
+import { TouchBackend } from 'react-dnd-touch-backend'
 
 export const DndWrapper = ({ children }: { children: ReactNode }) => {
+  const [hasTouchScreen, setHasTouchScreen] = useState(false)
+
+  // ensure it runs on client (could the same be accomplished with 'use client'
+  // directive instead of using the effect?)
```
hejtful: Could this help? https://nextjs.org/docs/app/api-reference/functions/userAgent

Entkenntnis:
Opening https://frontend-git-fix-drag-drop-touch-support-serlo.vercel.app/mathe/298184/einhorn-der-mathematik-episode-1 on my android/chrome is not working yet

hejtful:
I'll also work on fixing this now.

hejtful:
Current version seems to work. Changes:
1. Lazy loaded the DndWrapper component without SSR, to avoid the hydration errors. Dragging on mobile still didn't work, so:
2. Refactored a bit to simplify logic and reduce room for errors. Not as neat as Mikey's code, but now it works.

CodingDive:
https://frontend-chwwduwfk-serlo.vercel.app/mathe/298184/einhorn-der-mathematik-episode-1 works great now on Android and latest chrome. Only problem is that one doesn't see the items as it gets dragged. Only when it's dropped. Could someone check on an iPhone / Safari?

hejtful:
Yeah, styling is not optimal, we could use [this package](https://louisbrunner.github.io/dnd-multi-backend/packages/react-dnd-preview/) to improve it. Edit: Just tried it, works only globally, so not really suitable for our use-case.

Sadly, I don't have an Apple mobile device to try it out.

Entkenntnis:
DnD works for me now, too. The missing feedback (visible while dragging) is kind of a bummer, though 😢 

Another problem is that the touch backend has no mouse support. My device has touch enabled. The mouse is not working on my desktop.

CodingDive:
> The mouse is not working on my desktop.

Yikes! Could you try again in a moment? I set `enableMouseEvents` to true; however there is a scary warning of the docs as I also wrote in the source code https://react-dnd.github.io/react-dnd/docs/backends/touch.

> NOTE: This is buggy due to the difference in touchstart/touchend event propagation compared to mousedown/mouseup/click.


If that is too buggy indeed, we could try https://github.com/LouisBrunner/dnd-multi-backend/tree/main as Vito suggested or another package to enable touch + mouse events.

Entkenntnis:
@CodingDive This is how it looks like right now. It is still not working 😢 

firefox_k4vhhbaC5y

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
```diff
@@ -1,4 +1,4 @@
-import { DndWrapper } from '@editor/core/components/dnd-wrapper'
+import dynamic from 'next/dynamic'
```
elbotho: using nextjs features in the editor breaks external integrations (e.g. edusharing)

elbotho (changes requested):

elbotho:
just noted: while dragging my cursor defaults to default cursor instead of dragging hand (mac, chrome and safari)

CodingDive:
> while dragging my cursor defaults to default cursor instead of dragging hand (mac, chrome and safari)

Hmm some inconsistency from the OS. Works fine for me on Ubuntu with Firefox and Chrome. I added an extra line now, but if it still doesn't work, you could try and see where to add it.

packages/editor/src/plugins/fill-in-the-blanks-exercise/renderer.tsx
hejtful: Thanks, that was my bad...

elbotho:
test results:
works on mac: chrome, safari, firefox and all also with emulated touch
also in android chrome
iOS: will test soon

Entkenntnis (approved):
Thank, the latest version is also working on my touch/mouse-device.

I think we can go again and merge this PR.

There are more usability considerations (e.g. see slack), but these can be handled in follow ups.

PR #3372, opened 2024-02-12
fix(sorting): retrieve preview string from multimedia

### LINKED ISSUE #3364: sorting exercises: show string preview
https://de.serlo.org/taxonomy/term/sort/entities/298184

grafik

The string preview feature is quite useful - and would be nice if it works with multimedia plugins.

Entkenntnis:
example: http://localhost:3000/taxonomy/term/sort/entities/298184

### DIFF
--- a/packages/editor/src/plugins/text/utils/static-extract-text.ts
+++ b/packages/editor/src/plugins/text/utils/static-extract-text.ts
@@ -1,9 +1,5 @@
 import { AnyEditorDocument } from '@editor/types/editor-plugins'
-import {
-  isMultimediaDocument,
-  isRowsDocument,
-  isTextDocument,
-} from '@editor/types/plugin-type-guards'
+import { isTextDocument } from '@editor/types/plugin-type-guards'
 import type { Descendant } from 'slate'
 
 export function extractDescendant(node: Descendant): string {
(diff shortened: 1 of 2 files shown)


hejtful (approved):

packages/editor/src/plugins/text/utils/static-extract-text.ts
```diff
@@ -9,7 +13,19 @@ export function extractDescendant(node: Descendant): string {
...
+  if (
+    document &&
+    isMultimediaDocument(document) &&
+    isRowsDocument(document.state.explanation)
+  ) {
+    return document.state.explanation.state
+      .map(extractStringFromTextDocument)
+      .join(' ')
+      .trim()
+  }
```
elbotho: Since this is a helper in a file inside the text plugin and is named `extractStringFromTextDocument` I'd prefer this code to be somewhere else.
Imho. this should be a generic helper in e.g. `/plugin/helpers/static-extract-text-from-document` or something like that.

elbotho (changes requested):

packages/editor/src/plugins/text/utils/static-extract-text.ts
elbotho: (or inside multimedia plugin)

elbotho:
See comment above, but code seems to work fine, thank you!

packages/editor/src/plugins/text/utils/static-extract-text.ts
Entkenntnis: I have move the code to the one place it's used.

elbotho (commented):
thank you!

PR #3374, opened 2024-02-12
style(breadcrumbs): Increase margin to text on bigger screens

### LINKED ISSUE #3362: single exercise view: missing margin top
https://de.serlo.org/mathe/298524/298524

No margin between breadcrumbs and first line of text. Looks a bit stuffed.

grafik

CodingDive:
Fixes https://github.com/serlo/frontend/issues/3362

The real problem here is not that there is not enough margin-bottom, but there is a div with a negative margin :scream: This and a lot of other issues will be fixed once and for all in #3262 once I can fix the remaining problems.
 
image

### DIFF
--- a/apps/web/src/components/navigation/breadcrumbs.tsx
+++ b/apps/web/src/components/navigation/breadcrumbs.tsx
@@ -23,7 +23,10 @@ export function Breadcrumbs({
   noIcon,
 }: BreadcrumbsProps) {
   return (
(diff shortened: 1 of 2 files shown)


Entkenntnis:
@CodingDive I have talked to Botho and we agreed that the negative margin can be removed:

grafik

Can you adapto the PR to this change instead? This way, we are not adding additional margins that are superflous in this PR.

Entkenntnis (approved):
Thank you!

PR #3393, opened 2024-02-15
fix(rows): harden static renderer

### LINKED ISSUE #3391: user profile not working - 503
https://de.serlo.org/user/131532/ga87vof

https://de.serlo.org/user/148805/julianbabl

Entkenntnis:
In #3391 we encouter malformed states liek `{plugin: "rows"}` (state is missing).

Maybe this is happening somewhere else too. Harden the static renderer to not throw an error.


elbotho (approved):
ty

PR #3395, opened 2024-02-15
fix(geogebra): aovid using conditional rendering for modal

### LINKED ISSUE #3356: Cannot scroll page after closing modal
Steps to reproduce: 
- Create a geogebra element
- Click `Applet auswählen`
- Error is thrown in console `React-Modal: Cannot register modal instance that's already open`. After closing the modal you cannot scroll the page. 

It seems like this is only an issue using `yarn dev`, on staging it works fine.

Entkenntnis:
Caused by conditional rendering, which is not necessary because the modal is handling this by itself (see https://github.com/reactjs/react-modal/issues/808#issuecomment-1155423058)

@elbotho Probably justs an oversight and the conditional rendering was not intended or was for fixing some other bug?

### DIFF
--- a/packages/editor/src/plugins/audio/toolbar.tsx
+++ b/packages/editor/src/plugins/audio/toolbar.tsx
@@ -57,32 +57,29 @@ export const AudioToolbar = ({
           >
             {audioStrings.audioUrl} <FaIcon icon={faPencilAlt} />
           </button>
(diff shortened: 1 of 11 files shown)


elbotho:
@Entkenntnis pretty sure it's fine without the conditional. In my commit I adapted all other conditionally rendered modals I could find.

elbotho (approved):

PR #3396, opened 2024-02-15
fix(math): always use static math

### LINKED ISSUE #3394: consequently use static math in editor
We use two completely seperate code paths for rendering.

A consolidation could improve the situation.

Entkenntnis:
Not directly solving the issue with newlines, but will cause less confusion for editors.

Again, probably an oversight, as StaticMath is already used for edit mode.


elbotho (approved):
thank you 🧹

PR #3406, opened 2024-02-16
refactor(graphql): use `licenseId` instead of `license`

elbotho:
~waiting for https://github.com/serlo/api.serlo.org/pull/1262~


Entkenntnis (approved):
Thanks, change looks simple and straighforward

PR #3415, opened 2024-02-22
feat: construct entity id for exercises-group children

elbotho:


Entkenntnis (approved):
Nice, this is a clever hack! Can you add a filter to limit the length to 9 digits to avoid overflows? Otherwise this will work.

PR #3420, opened 2024-02-23
direct-deploy: show licenses for exercise groups

elbotho:
@Entkenntnis cherry-picked two commits from staging to get them online without having to think about implications of shipping injections stuff today ;) 
Any objections?


Entkenntnis (approved):
looks fine

PR #3426, opened 2024-02-25
draft(math-skills): personalized welcome screen

elbotho:
I started out with the modal approach and then tried this instead.

<table><tr><td>image</td><td>image</td></tr></table>

What I like about it:
- less intrusive, learners can also skip the step (we might ask again somewhere else)
- we avoid a lot more content on one page so there is no or less scrolling

It needed a bit more code to get it working as I wanted it to, but now it should work and we can also use it for points/stars/history etc.
(e.g. provider on each page, but that is mostly to not put stuff in `FrontendClientBase` for now)

Not yet done:
- stars
- persist to localStorage after asking
- avoid layout jump on landing page after entering name
- handle wrong user / delete data more cleanly


Entkenntnis:
Really cool alternative, it feels very satisfying to enter your name and take ownership of the page :)

Entkenntnis:
Provider-approach is suitable here, esp. for later stuff

elbotho:
too much?
image

Entkenntnis:
I like the name in the header, only on the landing page it's a bit duplicated - I would keep it as "Meine Mathe-Skill" there, but otherwise it's quite nice

PR #3430, opened 2024-02-27
fix(donation): Hide button for non German locale

CodingDive:
On the donation page, I'm also changing Wolfgang => Lena but am still waiting for her picture and will do it in a follow-up PR.


elbotho (approved):
nice, thank you!

PR #3434, opened 2024-02-27
feat(plugin-injection): temporarily make injections work with the migrated injection state

elbotho:
quickfix so we can migrate the injection migrations first without breaking something in production.

this basically parses the new format (`/{exercise-group-uuid}#{grouped-exercise-id}`) and loads the groupedExercise entity.

### DIFF
--- a/packages/editor/src/plugins/injection/static.tsx
+++ b/packages/editor/src/plugins/injection/static.tsx
@@ -102,10 +102,14 @@ export function InjectionStaticRenderer({
   >('loading')
   const { strings } = useInstanceData()
   const [base, hash] = href.split('#')
-  const cleanedHref = base
-    ? base.startsWith('/')
-      ? base
-      : `/${base}`
+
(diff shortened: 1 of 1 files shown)


Entkenntnis (approved):
looks good

packages/editor/src/plugins/injection/static.tsx
```diff
@@ -102,10 +102,14 @@ export function InjectionStaticRenderer({
   >('loading')
   const { strings } = useInstanceData()
   const [base, hash] = href.split('#')
-  const cleanedHref = base
-    ? base.startsWith('/')
-      ? base
-      : `/${base}`
+
+  // TODO: Temporary until migration is done
+  const baseRelacedWithHash = hash ? hash.replace(/\D/g, '') : base
```
Entkenntnis: -> baseRe**p**lacedWithHash

PR #3436, opened 2024-02-27
feat(math-skills): add skill points 🦓

elbotho:
image


Entkenntnis (approved):
cute

Entkenntnis:
The learning loop starts to take a nice shape! Clever rework of the overview. I started the app and then just couldn't stop for a while :)

PR #3446, opened 2024-02-28
math skill history swipe

elbotho:


Entkenntnis:
The dragger is very close the the edge, if I tap just some pixel to far to the left, the history swipe gesture activates, also on this PR

PR #3448, opened 2024-02-28
fix(article): Check for empty introductions in articles

CodingDive:
Fixes bug reported by Anne in Slack, showing the following warning on articles that have an introduction.

image

### DIFF
--- a/packages/editor/src/plugins/rows/utils/static-is-empty.ts
+++ b/packages/editor/src/plugins/rows/utils/static-is-empty.ts
@@ -1,10 +1,33 @@
 // eslint-disable-next-line import/no-cycle
-import type { AnyEditorDocument } from '@editor/types/editor-plugins'
+import type {
+  AnyEditorDocument,
+  EditorArticleDocument,
+  EditorArticleIntroductionDocument,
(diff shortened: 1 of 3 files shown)


packages/editor/src/plugins/rows/utils/static-is-empty.ts
```diff
@@ -1,10 +1,33 @@
...
+  AnyEditorDocument,
+  EditorArticleDocument,
+  EditorArticleIntroductionDocument,
+} from '@editor/types/editor-plugins'
 import { isRowsDocument } from '@editor/types/plugin-type-guards'
 
 import { isEmptyTextDocument } from '../../text/utils/static-is-empty'
 
-export function isEmptyRowsDocument(rows: AnyEditorDocument) {
+export function isEmptyArticle(article: EditorArticleDocument): boolean {
```
Entkenntnis: Can you move this function into a separate util file or into the article plugin? This file is within the rows-plugin.

CodingDive: Nice catch!

Entkenntnis (approved):
👍

PR #3449, opened 2024-02-28
fix(geogebra): Make geogebra sizing work regardless of being wrapped in `PrivacyWrapper` / `EmbedWrapper` or not

LarsTheGlidingSquirrel:
## Changes: 
- Before, geogebra renderer rendered differently if not in PrivacyWrapper or EmbedWrapper. Now, Geogebra renderer always renders in 16/9 aspect ratio regardless of wrapping component.
- Before, PrivacyWrapper and EmbedWrapper rendered placeholder all the time and rendered children on top of the placeholder. Now, PrivacyWrapper and EmbedWrapper stop rendering the placeholder after the button is clicked and replace it with `children`. This has one downside: If `children` render slowly it is not a seamless transition. 
- Before, some geogebra applets were not positioned correctly when first interacted with. Jumped up and were cropped. Now, this seem to be fixed. 
- Before, the placeholder appeared zoomed in compared to the loaded content. Now, they look the same. 

## Possible followup: 
- Merge PrivacyWrapper and EmbedWrapper code since they are doing very similar things.

### DIFF
--- a/apps/web/src/components/content/privacy-wrapper.tsx
+++ b/apps/web/src/components/content/privacy-wrapper.tsx
@@ -1,4 +1,4 @@
-import { faHeart, faSpinner } from '@fortawesome/free-solid-svg-icons'
+import { faHeart } from '@fortawesome/free-solid-svg-icons'
 import Image from 'next/image'
 import { useState, KeyboardEvent, useEffect } from 'react'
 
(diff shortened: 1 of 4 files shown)


apps/web/src/components/content/privacy-wrapper.tsx
```diff
@@ -95,31 +94,50 @@ export function PrivacyWrapper({
       provider: provider,
     })
 
-    if (showIframe && provider === ExternalProvider.Vocaroo) return null
-    if (isTwingle && showIframe) return null
```
hejtful, line 99: Did you figure out what the reason was behind these lines? Is it 100% safe to remove them?

hejtful (approved):
Works nice! Just one concern about the removed conditions.

apps/web/src/components/content/privacy-wrapper.tsx
LarsTheGlidingSquirrel: Probably, the placeholder had to be disabled in these cases because if would show through when the embedded content was rendered on top using `position: absolute`. But I did not test it. 
I think it is save to remove because `renderPlaceholder` in this PR is only called when `showContent == false`

PR #3453, opened 2024-02-29
chore(docs): Update readme and remove edtr-io type comments

CodingDive:
FYI: One of the files (packages/editor/src/plugin/migration.ts) is unused. I wonder if we should delete it entirely? 

In a follow-up PR, we could also change the `edtrIo` i18n string but I don't know if Crowdin lets us change the key easily https://github.com/serlo/frontend/blob/bba7d428b3fdc41d9629d69f178aa2a6f69377b7/apps/web/src/data/en/index.ts#L1056

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,21 +53,25 @@ Add the environment variable `DATABASE_URL` to `apps/web/.env.local` (next.js). 
 
 ## Repository Overview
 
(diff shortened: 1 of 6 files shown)


Entkenntnis (approved):
thanks for the chores

PR #3460, opened 2024-03-02
layout: improve mobile, make more compact, some ux ideas

elbotho:
I mostly worked on the mobile layout but threw in some ux ideas from the first user test.

changes:
- better use of screen space on mobile
- layout is more compact, but footer always stays at the bottom
- points always show (so it's easier to know that there will be three)
- "zur Übersicht" is now "☰ Aufgaben"
- points now replace the title in the header on mobile (more space for exercise)

topic overview:
- reuse welcome section for topic overview (makes it more personal) (could also be a state if we like it)
- cards use `<details>` now and are folded by default on smaller screens
- new feature: exercise of the day for an easy start

@Entkenntnis I did note merge it yet because it contains so many changes and I was not sure you'd like all of them :)

### DIFF
--- a/apps/web/src/components/math-skills/exercise-implementations/memory/memory-game.tsx
+++ b/apps/web/src/components/math-skills/exercise-implementations/memory/memory-game.tsx
@@ -119,7 +119,7 @@ export function MemoryGame({
         </h2>
         <nav
           className={cn(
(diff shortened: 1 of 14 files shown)


Entkenntnis (approved):
contains a lot of things a wanted to give as feedback anyways (more compact on mobile, etc) and feels more "rounded" overall.

You can go ahead and merge it

PR #3461, opened 2024-03-03
draft for diagram component (as task and exercise input)

elbotho:
evaluated chart libs first, but since we only need it for some exercises and since dragging bars is usually not supported out of the box anyway ^^ I chose to use a prebuild svg.

### [demo](https://frontend-git-math-skills-diagram-serlo.vercel.app/meine-mathe-skills/klasse5/diagramm-wip)

image


Entkenntnis:
postponed for now

PR #3470, opened 2024-03-05
fix(blanks): UI fixes

hejtful:
Please check commit messages (short and long) for more info.

### DIFF
--- a/apps/web/src/data/en/index.ts
+++ b/apps/web/src/data/en/index.ts
@@ -902,6 +902,7 @@ export const loggedInData = {
         exercise: {
           title: 'Exercise',
           description: 'Interactive or text based exercise',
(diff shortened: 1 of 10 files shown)


apps/web/src/data/en/index.ts
```diff
@@ -902,6 +902,7 @@ export const loggedInData = {
         exercise: {
           title: 'Exercise',
           description: 'Interactive or text based exercise',
+          placeholder: 'Type the assignment here (Optional)',
```
hejtful, line 905: @elbotho The original German text I got from Gregor was "Füge hier den Arbeitsauftrag ein. (Optional)". Would you mind adding the German version to Crowdin?

elbotho (approved):
lovely!

PR #3473, opened 2024-03-05
math-skills: quick exeriment using JSXGraph

elbotho:
~currently only works using page router, if you access page directly it breaks for some reason.
Also builds break 😬.
Will investigate if we like the approach.~

demo: https://frontend-git-math-skills-jsx-graph-text-serlo.vercel.app/meine-mathe-skills/klasse5/jsxgraph-example


elbotho:
closing in favour of https://github.com/serlo/frontend/pull/3490

PR #3475, opened 2024-03-06
math-skills: euros

elbotho:
just to get a feel, but I think I got lucky with a pretty easy section.


Entkenntnis:
Looks great! Some minor notes:

- https://realmath.de/Neues/Klasse5/geld/euro3.php is missing?
- https://frontend-git-math-skills-sizes-serlo.vercel.app/meine-mathe-skills/klasse5/plusminus-vergleichen-1: the cent values are multiples of 5 in the original to ease calculating - I think we can keep it, at least for level 1. And maybe add `euro` to the url to make distinction clearer.

PR #3489, opened 2024-03-08
math-skills: trigonometry exercise draft

elbotho:


Entkenntnis:
Love it, especially the simplicity of the definition

Entkenntnis:
If you can also add some labels to the line parts, this would be perfect

PR #3495, opened 2024-03-12
Add SpoilerOpened tracking event, for saving in PlanetScale

shn-srl:
Sends a SpoilerOpened tracking event to Planetscale

To confirm:

[ ] Is all the required information being saved ?

- Important: PlanetScale migration also needs to run

https://app.planetscale.com/serlo/serlo_planetscale/test-spoiler-tracking/

image


packages/editor/src/plugins/spoiler/static.tsx
```diff
@@ -1,18 +1,40 @@
...
   const { title, content } = state
+  const [hasSentSpoilerTrackingEvent, setHasSentSpoilerTrackingEvent] = useState(false)
+
+  const ab = useAB()
+
+  const trackSpoilerOpened = () => {
+    if(!ab || hasSentSpoilerTrackingEvent) return
+    // send tracking event
+    abSubmission({
+      entityId: -1, // should this be a real id?
```
shn-srl: This needs to be double checked.

If entity id is needed, what id is it? I see string Ids but the abSubmission expects an int

Entkenntnis (changes requested):
Two minor things:

- Can you revert the db change?
- Check if entity id is one of the three we use this event for

shn-srl:
> Two minor things:
> 
>     * Can you revert the db change?
> 
>     * Check if entity id is one of the three we use this event for

DB change was on a development branch, which is now deleted :heavy_check_mark: 

Added EntityId check :heavy_check_mark:

Entkenntnis (approved):

Entkenntnis (changes requested):
move code to integration and out of editor repo

Entkenntnis (approved):
Thank you

PR #3496, opened 2024-03-12
feat(editor): package simpler language strings

hejtful:
Changes:
- No need to export language data, it is now provided via the `SerloEditor` component
- An optional `language` property (currently, only `de` and `en` supported and packed)

Implementation example in: https://github.com/serlo/serlo-editor-for-edusharing/pull/350

Note: Instructions for simple local editor development with integrations added.


packages/editor/src/core/editor.tsx
```diff
@@ -135,8 +140,8 @@ function InnerDocument({ children, onChange, ...props }: EditorProps) {
   function renderChildren(id: string) {
     const document = <SubDocument id={id} />
 
-    if (typeof children === 'function') {
-      return children(document)
+    if (typeof children === 'function' && loggedInData) {
```
hejtful, line 143: This seemed like a possible bug trigger, but I didn't notice any problems while testing.

elbotho (approved):
thank you!

CodingDive (approved):
Leads to small bundle size increase, but looks good to me! Thank you

PR #3530, opened 2024-03-20
feat(editor): editor object in place of store exports

hejtful:
**Change in API design**: Instead of providing only the rendered `editor` element, provide an `editor` object containing the element, language data, Redux store data, etc.

**Localized change**: To reduce the amount of Editor exports, provide the Redux store data (`history` and `selectRootDocument`) as properties of the `editor` object. These properties only expose the data from the store needed for using the Editor package.

**Side effect changes**: Split the `packages/editor/src/core/editor.tsx` file, that held two components, into three files that hold one component each. Extracted types into a separate file. Rename `languageData` property to `i18n`.

Matching edusharing PR: https://github.com/serlo/serlo-editor-for-edusharing/pull/361


packages/editor/src/core/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  initialState: {
+    plugin: string
+    state?: unknown
+  }
+  onChange?: OnEditorChange
+}
+
+export type EditorChildren = ReactNode | ((editor: EditorData) => ReactNode)
+
+export interface EditorData {
```
hejtful: A better name, maybe? 😬 Any suggestions?

packages/editor/src/core/editor-children.tsx
```diff
@@ -0,0 +1,75 @@
...
+        pendingChanges,
+        dispatchUndo,
+        dispatchRedo,
+        dispatchPersistHistory,
+      },
+      selectRootDocument,
+    })
+  }
+
+  return (
```
shn-srl: nitpick: Returning this above the if statement would be better (return early)

packages/editor/src/core/types.ts
shn-srl: EditorConfigData ? :see_no_evil:

shn-srl (approved):

packages/editor/src/core/editor-children.tsx
LarsTheGlidingSquirrel: Could we drop this case altogether? I think all editor integrations will want to pass a function (render prop) as `children` to get access to `selectRootDocument`, which is needed to save the document.

packages/editor/src/core/types.ts
LarsTheGlidingSquirrel: SerloEditorChildrenParameters ? (or abbreviated)

packages/editor/src/core/editor-children.tsx
```diff
@@ -0,0 +1,75 @@
...
+    void dispatch(persistHistory(documents))
+  }, [dispatch])
+
+  const selectRootDocument = useCallback(() => {
+    return selectStaticDocument(store.getState(), ROOT)
+  }, [])
+
+  const editor = <SubDocument id={ROOT} />
+
+  if (typeof children === 'function' && loggedInData) {
```
LarsTheGlidingSquirrel: As far as I understand `loggedInData` can never be falsy here. Maybe clarify that in the code?

LarsTheGlidingSquirrel (approved):
Thank you! Added some comments but no blockers from my point of view.

packages/editor/src/core/editor-children.tsx
hejtful: @shn-srl good point!

@LarsTheGlidingSquirrel The frontend does use this case though.

packages/editor/src/core/editor-children.tsx
hejtful: I'm not sure what you're saying, sorry. Its type at this point is `LoggedInData | null`, therefore this check. Maybe you could provide like a code block suggestion of the changes you're proposing?

LarsTheGlidingSquirrel: Something like 
```
if(!loggedInData) { 
  // Handle error, maybe return null, not sure what would be best here 
}
if (typeof children === 'function') {
  // ...
}
```
Right now, it reads like also `loggedInData` determines if we call `children` or render it directly. 
Hope that makes it more clear.

packages/editor/src/core/editor-children.tsx
LarsTheGlidingSquirrel: > @LarsTheGlidingSquirrel The frontend does use this case though.

Oh, you are right. Nevermind.

packages/editor/src/core/editor-children.tsx
hejtful: Thanks for explaining, makes sense :)

PR #3540, opened 2024-03-21
fix(exams-info-box): Fix layout

LarsTheGlidingSquirrel:
Quick fix for two layout issues with the exams-info-box, see: 
image
image

### DIFF
--- a/apps/web/shared/tailwind-base.js
+++ b/apps/web/shared/tailwind-base.js
@@ -163,6 +163,7 @@ export default {
       md: '1024px',
       lg: '1216px',
       xl: '1300px',
+      '2xl': '1436px',
       print: { raw: 'print' },
     },
   },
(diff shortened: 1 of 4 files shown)


elbotho:
thank you. this is probably the most complexity for a responsive feature we have 🙈. 
https://frontend-git-fix-layout-exams-info-box-serlo.vercel.app/mathe/302439/pflichtteil-1-wtr is still broken on 1024+ screens. 

We should probably move up the breakpoints again so it always has enough space.

elbotho (approved):
Thank you very much!
I made it break even later, so the exercise numbers don't get clipped.

LarsTheGlidingSquirrel:
> Thank you very much! I made it break even later, so the exercise numbers don't get clipped.

Even better! Thanks!

PR #3543, opened 2024-03-22
feat(plugin-course): improve and force course editing

elbotho:
- CoursePages are now opened in Editor inside the whole Course 
- Courses only show one CoursePage at a time in the editor

Downside: Authors will not be warned when there are previous unrevised revisions of course pages.

This is in preparation of moving CoursePages to the Courses editor state.


apps/web/src/components/user-tools/edit-or-invite/edit-or-invite.tsx
```diff
@@ -49,7 +49,16 @@ export function EditOrInvite({
   const hasUnrevised =
     unrevisedRevisions !== undefined && unrevisedRevisions > 0
 
-  const href = hasUnrevised ? getHistoryUrl(data.id) : getEditHref()
+  const isCourse =
+    data.type === UuidType.Course || data.type === UuidType.CoursePage
+
+  const href = isCourse
```
shn-srl, line 55: Nit/Suggestion:

The ternary is quite pretty, however I would be inclined to do something like:

``` tsx
function getHref({ isCourse, data, hasUnrevised }) {
  if (!isCourse) return hasUnrevised ? getHistoryUrl(data.id) : getEditHref();
    
    return data.unrevisedCourseRevisions && data.unrevisedCourseRevisions > 0
      ? getHistoryUrl(data.courseId ?? data.id)
      : getEditHref();
}

// Usage
const href = getHref({ isCourse, data, hasUnrevised });
```

apps/web/src/components/user-tools/edit-or-invite/edit-or-invite.tsx
```diff
@@ -81,13 +90,17 @@ export function EditOrInvite({
     if (!data) return undefined
     const revisionId = data.revisionId
     const { type, id } = data
-    const url = getEditUrl(id, revisionId, type.startsWith('Taxonomy'))
+
+    const url = isCourse
```
shn-srl, line 94:
``` tsx
function getUrl(data, isCourse) {
  const { revisionId, type, id, courseId } = data;
  if (!isCourse) return getEditUrl(id, revisionId, type.startsWith('Taxonomy'));

  return `${getEditUrl(courseId ?? id, undefined, false)}#${id}`;
}

// Usage
const url = getUrl(data, isCourse);

```

packages/editor/src/plugins/serlo-template-plugins/course/course.tsx
```diff
@@ -152,14 +136,75 @@ function CourseTypeEditor(props: EditorPluginProps<CourseTypePluginState>) {
...
+                  }}
+                  className={cn(
+                    'serlo-link text-lg leading-browser',
+                    index === activePageIndex &&
+                      'font-semibold text-almost-black hover:no-underline'
+                  )}
+                >
+                  {title.trim().length ? title : '___'}
+                </button>{' '}
+                {/* {index > 0 ? (
```
shn-srl, line 164: Can we remove this comment block?

apps/web/src/components/user-tools/edit-or-invite/edit-or-invite.tsx
elbotho: tbh. this feels like personal taste at first glance.
do you see an upside to this solution that I'm missing?

packages/editor/src/plugins/serlo-template-plugins/course/course.tsx
elbotho: good question 😅

apps/web/src/components/user-tools/edit-or-invite/edit-or-invite.tsx
elbotho: (using ` ``` tsx` gives you syntax highlighting :)

shn-srl: "Cleaner code" but yes I agree its arguably a taste thing, feel free to close the thread :)

elbotho: thanks for the suggestion!
I'll leave it for now.

packages/editor/src/plugins/serlo-template-plugins/course/course.tsx
elbotho: added a comment, usually I'd remove unused code, but this will hopefully become relevant soonish…

packages/editor/src/plugins/serlo-template-plugins/course/course.tsx
```diff
@@ -83,40 +91,17 @@ function CourseTypeEditor(props: EditorPluginProps<CourseTypePluginState>) {
...
-          )
-        })}
-        <div className="mt-24 border-t-2 border-editor-primary-200 pt-12">
-          <AddButton onClick={() => children.insert()}>
+        <div className="ml-side mt-4">
+          <AddButton
+            onClick={() => {
+              children.insert()
+              setActivePageIndex(staticPages.length)
+            }}
```
CodingDive: The code looks great! The only thing that doesn't work as intended I believe is this part. It doesn't transition to the newly inserted active page because the hash of the URL is not updated, and the `useEffect` hook changes the `setActivePageIndex` in the next render to the previous page (current hash). 
Also, it'd be nice to set at a minimum `setCourseNavOpen(true)`, so the behavior as in the gif does not happen where I did not realize, that I have already created many course pages. 

nothing_seemingly_happens_after_add_course_page_click


I tried to fix it here by removing the hash but consequently ran into some issues with the `useLeaveConfirm` hook. https://github.com/serlo/frontend/pull/3621 Feel free to use some of the code and possibly you'll find a more resilient workaround for preventing the dialog from popping up!

elbotho: Thanks, I did not encounter it before.
https://github.com/serlo/frontend/pull/3543/commits/9211bb374db9d334ea001637f0e9097396ae13f8 should fix it.

CodingDive: Works great! No confirmation dialog either :massage_man:

CodingDive (approved):
Works great!

One small thing, that is not too critical and we can address in a follow-up PR
image

On my small laptop screen, the heading is scrolled to the right when loading the page http://localhost:3000/entity/repository/add-revision/155795#155798

PR #3553, opened 2024-03-25
refactor(json-ld): embed context directly

elbotho:
> Easy fix: Inhalt von https://w3id.org/kim/lrmi-profile/draft/context.jsonld direkt reinpacken

@kulla like this?


kulla (approved):
Should work :smile: Thx!

elbotho:
> Should work 😄 Thx!

I'll check with Vicky and hope her parser swallows it.

PR #3528, opened 2024-03-20
docs: Editor package API

shn-srl:

PR #3554, opened 2024-03-25
Export editor as web component

CodingDive:
In follow up PRS, we'll need to define a few more attributes for the web component.

For now, in its first iteration, the PR should be ready for review.


packages/editor/src/plugins/exercise-group/static.tsx
```diff
@@ -45,7 +45,7 @@ export function ExerciseGroupStaticRenderer(
             <ExerciseLicenseNotice exerciseLicenseId={context?.licenseId} />
           </div>
         ) : null}
-        {loaded && auth && context?.uuid ? (
+        {loaded && context?.uuid ? (
```
CodingDive: One thing I forgot to ask earlier, is what we are supposed to do with `auth` dependencies from the frontend. Should we even render the `AuthorToolsExercises` at all?

packages/editor/src/plugins/serlo-template-plugins/exercise-group/static.tsx
```diff
@@ -36,7 +36,7 @@ export function TextExerciseGroupTypeStaticRenderer(
               <ExerciseLicenseNotice exerciseLicenseId={context?.licenseId} />
             </div>
           ) : null}
-          {loaded && auth && context?.uuid ? (
+          {loaded && context?.uuid ? (
```
CodingDive: Same question here. How do we deal with the auth? Do we render `AuthorToolsExercises`? Require some auth of the package consumer who render the editor?

packages/editor/src/plugins/text/static-components/use-lazy-load-katex-styles.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
+
+function loadKaTeXStyles() {
+  const stylesheetHref = '@serlo/frontend/src/../external/katexstyles.css'
```
CodingDive: This works locally, but I have a not-so-good feeling here that this won't work for production or once publishing to npm, because I assume the URL can't be found and we don't tell Vite about this asset either. Any other ideas on how best to lazyload css styles without a Next.js dependency? Should we put the katexstyles into the public assets and load it from there?

packages/editor/vite.config.ts
```diff
@@ -19,13 +19,13 @@ export default defineConfig({
...
     },
     rollupOptions: {
-      external: ['react', 'react-dom'],
-      output: {
-        globals: {
-          react: 'React',
-          'react-dom': 'ReactDOM',
-        },
-      },
+      // external: ['react', 'react-dom'],
```
CodingDive: TODO for me: Delete comments. To make the web component work, we'll need to bundle React and ReactDOM. This will maybe annoy non-web-component consumers of the package, as wed' lock them into our React version (and increase the bundle size).

packages/editor/src/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
```diff
@@ -18,15 +18,19 @@ import { SaveModal } from '@serlo/frontend/src/serlo-editor-integration/componen
...
-import { useLeaveConfirm } from '@/helper/use-leave-confirm'
 
 interface ToolbarMainProps {
   changes?: StateTypeReturnType<(typeof entity)['changes']>
   licenseId?: StateTypeReturnType<(typeof entity)['licenseId']>
   showSubscriptionOptions?: boolean
 }
 
+const isNextApp = () => {
+  return Boolean(document.getElementById('__next'))
```
shn-srl, line 31: Did we not remove the __next tag from root?

Or do we expect this id somewhere else trotzdem?

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,327 @@
...
+function isInitialState(obj: any): obj is InitialState {
+  return (
+    obj !== null &&
+    obj !== undefined &&
+    typeof obj === 'object' &&
+    'plugin' in obj
+  )
+}
+
+const exampleInitialState: InitialState = {
```
elbotho: maybe move to own file?

packages/editor-web-component/src/simple-web-component.tsx
elbotho, line 1: do you still need the simple test / examples?

packages/editor-web-component/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
+import sharedTailwindBase from '@serlo/frontend/shared/tailwind-base.js'
+
+// This file is probably not needed and can be deleted!
```
elbotho: remove comment or file then ;)

packages/editor/src/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
elbotho: I think this PR only changes the targets to something we set ourselves. `__next` get's injected by automatically, I'm not even sure if it can be removed.

apps/web/src/pages/_document.tsx
```diff
@@ -41,7 +41,7 @@ export default class MyDocument extends Document {
       ? getInstanceDataByLang(this.props.__NEXT_DATA__.locale as Instance)
       : undefined
     return (
-      <Html className="bg-brand-100 print:serlo-print-style">
+      <Html className="bg-brand-100 print:serlo-print-style" id="serlo-root">
```
elbotho, line 44: putting it here means that the modals would not be inside `<body>`, right?

elbotho (commented):
Overall this looks great. I added some questions / suggestions.
Also this needs testing if everything works as expected with the changes in the other packages.

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,327 @@
...
+    this.mountReactComponent()
+  }
+
+  mountReactComponent() {
+    const initialStateAttr = this.getAttribute('initial-state')
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const initialState: InitialState = initialStateAttr
+      ? (JSON.parse(initialStateAttr) as unknown as any)
+      : // TODO throw error here instead?
```
shn-srl: Try / Catch ?

Will line 288 currently throw an error or always pass since it has the sample initial  state?

What is desired behaviour?

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,327 @@
...
+        exampleInitialState
+
+    if (!isInitialState(initialState)) {
+      throw new Error('Initial state is not of type InitialState')
+    }
+
+    console.log('Mounting React Component', initialState)
+
+    if (this.reactRoot) {
```
shn-srl: nit: return early

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,327 @@
...
+      this.reactRoot.render(
+        <React.StrictMode>
+          <SerloEditor
+            initialState={initialState}
+            pluginsConfig={{
+              basicPluginsConfig: {
```
shn-srl: nit: Would perhaps be nicer defined as a const

packages/editor/src/plugins/text/static-components/use-lazy-load-katex-styles.tsx
shn-srl: Without much insight, it seems like vite should have some config option for lazy loading stylesheets and that we should not need to do this so manually

packages/editor/src/plugins/serlo-template-plugins/toolbar-main/toolbar-main.tsx
CodingDive: Yeah and one can not change it unfortunately.

apps/web/src/pages/_document.tsx
CodingDive: Correct! I tested with a screen reader chrome extension and could not see that it effects a11y in any negative way.

packages/editor-web-component/src/editor-web-component.tsx
CodingDive: I changed the initial state to a "useful default" for first timers with the editor 

(a heading, paragraph and an equation). This will take some iteration with Vicky and our partners on what default to have here, but I think a "blank slate" in the editor is not desired.
Therefore, I removed the comment for now.

packages/editor-web-component/src/editor-web-component.tsx
CodingDive: Or even better: remove entirely :tada: (moved them to defaults in the editor package)

packages/editor/src/plugins/text/static-components/use-lazy-load-katex-styles.tsx
CodingDive: Moved this to a separate npm package and github repo https://github.com/serlo/serlo-katex-styles which solves the limitations of next.js imports (they allow to import styles dynamically from node_modules but not from anywhere else). For now it lives here https://www.npmjs.com/package/serlo-katex-styles, but happy to move it over to the Serlo npm org :blush:

packages/editor-web-component/package.json
```diff
@@ -0,0 +1,88 @@
...
+    ".": {
+      "import": "./dist/serlo-editor-web-component.js",
+      "require": "./dist/serlo-editor-web-component.js",
+      "types": "./dist/editor-web-component.d.ts"
+    }
+  },
+  "main": "./dist/serlo-editor-web-component.js",
+  "module": "./dist/serlo-editor-web-component.js",
+  "types": "./dist/editor-web-component.d.ts",
+  "files": [
```
yorrd, line 25: this leads to a lot of things being exposed through the package under the dist directory. Common practice would be exposing those types via tree shakable semantic subpaths and strip internals. 

image

Understandable if a semantic structure won't be part of the 1.0 release since tree shaking would really not bring any benefit at the moment and would mostly just bloat the build process atm. I'm bringing this up because it's usually good practice to have well structured import paths - my suggestion for now would be just not exposing the dist directory in the final bundle but rather only the index export (ln 15). Not a hugely important topic.

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,89 @@
...
+import {
+  exampleInitialState,
+  isInitialState,
+  type InitialState,
+} from './initial-state'
+
+// Could probably remove the export entirely, as the customElement is registered
+// below.
+export class EditorWebComponent extends HTMLElement {
+  private reactRoot: ReactDOM.Root | null = null
```
yorrd, line 14: for a lack of a better place to put this, I'll have it at the react dom root:

In vanilla JS applications, there is no `process.env.`. However, the bundle assumes that it exists and bears `"production"` or `"staging"`. Easiest workaround is to include `process = {env: 'production'}` in the banner of the bundle if the environment does not already worry about the global process variable through another bundler. The "real" solution would be not to rely on that variable in the bundle's output but I'm not sure as to how realistic that is on the short term. 

From experience, that's usually a big pain point for package distributors since bundling issues hold back adoption. Putting a workaround in one's own code looks super ugly and is frowned upon, still holding back adoption. On the other hand, even excalidraw has this issue since they also bundle react in their component, but notably they don't pretend to be agnostic and just deliver a react component.

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,89 @@
...
+  private container: HTMLDivElement
+
+  constructor() {
+    super()
+
+    this.container = document.createElement('div')
+  }
+
+  attributeChangedCallback(name: string, oldValue: string, newValue: string) {
+    console.log('AttributeChangedCallback', {
```
yorrd: not sure if we need this ;)

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,89 @@
...
+      oldValue,
+      newValue,
+    })
+    if (oldValue !== newValue) {
+      this.mountReactComponent()
+    }
+  }
+
+  static get observedAttributes() {
+    return ['initial-state']
```
yorrd: this is not black or white but exposing structured data as an attribute seems weird to me. Internally we usually expose through a property (in lit syntax: `<my-el .initialState=${{key: 'value'}}></my-el>`) to avoid confusion around serialization. All webcomponent environments allow some sort of property binding since there's always things you won't be able to do with native DOM APIs. 

To allow for serializing the serlo editor invocation including its initial state, you could allow both a property and the attribute. You'd just need to `??` in ln49 and add the respective class member. 

To the argument "stay as close to DOM APIs as possible" while a good point in principle, I would reply that you're going to be forced to have some sort of more complex API very soon anyways when you want to have any kind of plugin adding API, especially if you're going to allow for custom plugins.

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,89 @@
...
+      this.reactRoot = null
+    }
+  }
+
+  updateProps(/*newProps: any*/) {
+    // Method to update props if necessary, calling this.mountReactComponent() again with new props
+  }
+}
+
+customElements.define('editor-web-component', EditorWebComponent)
```
yorrd: some librarise expose the class rather than the registered element since unfortunately the `customElements` registry is not able to change or remove elements at the moment. The respective W3C github thread has been debated for years but doesn't have a conclusion yet. 

At the very least you should scope the name here (e.g. `serlo-editor`) to avoid naming conflicts with other libraries. Having `web-component` in the name is kind of redundant.

packages/editor-web-component/src/initial-state.tsx
```diff
@@ -0,0 +1,143 @@
+import { SerloEditorProps } from '@serlo/editor'
+
+export type InitialState = SerloEditorProps['initialState']
```
yorrd, line 3: really appreciate you putting a default initial state in here, great thought to be able to plug and play

yorrd (commented):
focused on the webcomponent part

yorrd:
I just completely forgot to add one thing @CodingDive : we need a way to expose the current editor state. Good practice is to dispatch an event on any editor change. Usually that happens on blur, not every keystroke. Does the editor already have infrastructure for this? As an alternative, we can just grab the current data from the outside, but then we wouldn't have visibility of dirty state. We can discuss on Wednesday, but this is obviously a very important part for any integration.

packages/editor-web-component/src/editor-web-component.tsx
CodingDive: Would you recommend me to put this global variable into the vite.config or how should we proceed here?

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -0,0 +1,128 @@
...
+  isInitialState,
+  type InitialState,
+} from './initial-state'
+
+// Could probably remove the export entirely, as the customElement is registered
+// below.
+export class EditorWebComponent extends HTMLElement {
+  private reactRoot: ReactDOM.Root | null = null
+  private container: HTMLDivElement
+  private broadcastChannel: BroadcastChannel
```
CodingDive: Not happy with the broadcastChannel. I believe custom HTML events, or binding a callback through a ref to be a better solution. Asked Kai in Slack and presented a few different options

hejtful:
@CodingDive In the current code, there are:
1. `onChange` property (which doesn't seem to be used by the editor)
2. Event broadcasting
Do we need both of these, or can we remove the `onChange` prop? And did you plan to add some documentation about the event broadcasting (ideally, you could add it to [this PR](https://github.com/serlo/frontend/pull/3528))?

CodingDive:
@hejtful done here https://github.com/serlo/frontend/pull/3658. 

The `onChange`, I see passed down here and used https://github.com/serlo/frontend/blob/staging/packages/editor/src/core/inner-document.tsx#L27. Web component gets the state updates from the editor via `onChange` but then broadcasts it to the web component consumer via the `customEvent`.

packages/editor-web-component/src/editor-web-component.tsx
yorrd: sorry missed github notifications... not too familiar with vite's build system but if you can add global constants that way, that would probably be good. Anywhere that's not in the actual code because it's only for "glueing stuff together"

yorrd: sorry completely missed github notifications. I'm not too familiar with vite's build system. That being said, anywhere that's at compile time to inject the thing into the final bundle. Build config sounds fair enough to me without knowing details about vite stuff.

yorrd:
As discussed in person, need iteration on shadow dom integration because it currently breaks the rendering of CSS due to scoping issues. Detailed feedback on usability will be iterated on after the merge in collaboration with @CodingDive .

Therefore, happy to approve as a preliminary release version for guided internal use.

yorrd (approved):

hejtful (approved):

PR #3557, opened 2024-03-25
fix(plugin-blanks): alternative answers safari focus bug

hejtful:
Slate focus management didn't work reliably in Safari, so blurring the Slate editor manually when adding an alternative answer.

Fix for: https://trello.com/c/u7osCXuw/626-bug-vom-editor-auf-ipad


elbotho (approved):
thank you for fixing this

PR #3560, opened 2024-03-26
fix(plugin-equations): hide transformation for term equation

hejtful:
Fix for: https://trello.com/c/T7wsqBqh/627-kleiner-bug-bei-termumformungs-pluggin

@kulla @elbotho Could you please check [the preview](https://frontend-git-fix-transformation-visible-in-term-eq-b6a90d-serlo.vercel.app/entity/repository/add-revision/304018) to confirm if I fixed this properly, as I'm still not 100% sure that I understood the problem.


elbotho (approved):
love it. minimal fix 👍

PR #3561, opened 2024-03-26
Fix(geogebra): Make multiple geogebra work in `serlo-editor-for-edusharing`

LarsTheGlidingSquirrel:
**Before**: `<Script>` from nextjs did not call `onReady` for each geogebra instance. This meant some geogebra applets did not load. Could not make it work this way.

**After**: Using `document.body.appendChild` to attach and load the necessary geogebra script and then injecting the geogebra applet in a `useEffect`. Also removed `url` from `GeogebraRendererProps` because it is not used / necessary.

### DIFF
--- a/packages/editor/src/plugins/geogebra/renderer.tsx
+++ b/packages/editor/src/plugins/geogebra/renderer.tsx
@@ -1,56 +1,88 @@
-import Script from 'next/script'
-import { useState } from 'react'
+import { memo, useEffect, useState } from 'react'
 import { v4 as uuid_v4 } from 'uuid'
 
 import { cn } from '@/helper/cn'
 
 export interface GeogebraRendererProps {
   geogebraId: string
-  url: string
 }
 
-export function GeogebraRenderer({ geogebraId, url }: GeogebraRendererProps) {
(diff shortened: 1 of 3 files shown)


packages/editor/src/plugins/geogebra/renderer.tsx
```diff
@@ -1,70 +1,100 @@
...
+        // https://javascript.info/global-object
+        // Using 'global' lead to error 'global not defined' in editor package when using vite.
+        const scriptTag = document.createElement('script')
+        scriptTag.src = 'https://www.geogebra.org/apps/deployggb.js'
+        scriptTag.addEventListener('load', () => {
+          resolve()
+        })
+        scriptTag.addEventListener('error', () => {
+          reject()
+        })
```
hejtful, line 78: If it's not too much extra work, maybe you could also handle removing these event listeners?

hejtful (approved):
One less dependency on Next in the editor, thank you!

packages/editor/src/plugins/geogebra/renderer.tsx
LarsTheGlidingSquirrel: Hm, I thought about it but don't think it is necessary. Each time this code is run the listeners are added to a newly created script tag, so it should never happen that we add duplicates. But I might be missing something. (also my first implementation idea appears to make the code a lot more messy :sweat_smile: )

PR #3562, opened 2024-03-27
feat(editor): plugins renderers config object

hejtful:
Replaces plugins and renderers initialization exports with a `pluginsConfig` property on `SerloEditor` component.

Matching Edusharing PR: https://github.com/serlo/serlo-editor-for-edusharing/pull/363


shn-srl (approved):

CodingDive (approved):
Looks good to me! Thank you

packages/editor/src/package/editor.tsx
```diff
@@ -48,21 +49,15 @@ const emptyState = {
...
-  language = 'de',
-}: SerloEditorProps) {
+export function SerloEditor(props: SerloEditorProps) {
+  const { children, pluginsConfig, initialState, language = 'de' } = props
+  const { basicPluginsConfig, customPlugins, customRenderers } = pluginsConfig
   const { instanceData, loggedInData } = editorData[language]
 
   const basicPlugins = createBasicPlugins(basicPluginsConfig)
-  editorPlugins.init([...basicPlugins, ...customPlugins])
+  editorPlugins.init([...basicPlugins, ...(customPlugins || [])])
```
shn-srl: nit: could we move the "empty case" (|| [])  up to line 54 plugins deconstruction, the same way we have `language = 'de'`

shn-srl (approved):

packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -1,22 +1,31 @@
...
+  pluginsConfig,
   language = 'de',
   ...props
 }: SerloRendererProps) {
+  const { customPlugins, customRenderers } = pluginsConfig
   const { instanceData, loggedInData } = editorData[language]
 
+  const basicRenderers = createRenderers(customPlugins)
+  editorRenderers.init({ ...basicRenderers, ...customRenderers })
+
```
LarsTheGlidingSquirrel: I was a little bit confused here. For example, the renderer for `edusharingAsset` is now within `customPlugins` not in `customRenderers`. Only the link renderer is still in `customRenderers`. I understand it now but I think the code would get easier to understand if we continue to consolidate the `renderer` and `editor` part for plugins like you did with the `customPlugins` for all plugins. I think having one array of plugins including both the "renderer" and the "editor" would make the code way more clear. But maybe there was a good reason to separate them, maybe for bundle size?

packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -1,22 +1,31 @@
...
 
 export function SerloRenderer({
+  pluginsConfig,
   language = 'de',
   ...props
 }: SerloRendererProps) {
+  const { customPlugins, customRenderers } = pluginsConfig
   const { instanceData, loggedInData } = editorData[language]
 
+  const basicRenderers = createRenderers(customPlugins)
```
LarsTheGlidingSquirrel, line 26: This includes the `image` plugin, which is specific to serlo.org and not relevant to the other integrations. But not a problem since an additional renderer that is never used does not hurt here.

LarsTheGlidingSquirrel (approved):
Nice! A good step in the right direction in my opinion. 
One thing I noticed is that the interface between integration & editor starts to get different compared to the interface between `frontend` and editor. But I assume this will be temporary until we make `frontend` use the same interface as the integrations at some point. 
Also added some comments to specific lines.

packages/editor/src/package/serlo-renderer.tsx
hejtful: It's not bundle size, `linkRenderer` is separate because `link` is not a plugin. But I guess then `customRenderers` is a confusing name. At first, I named it `customLeafRenderers`, but that's Slate-specific and maybe not 100% accurate. Could you maybe propose a structure that would make sense to you?

packages/editor/src/package/editor.tsx
shn-srl: Sorry to unresolve, we can remove it on this line now right?
:smiling_imp: :angel:

PR #3569, opened 2024-03-28
Prove concept of plugin strings generation script

shn-srl:
Change:

Move strings to json files which are imported by current "[lang]/index.ts" and exported as currently implmented.

Script offers ability to create the plugin strings quicker, by offering CLI prompts, and then updating the JSON files programatically

image


CodingDive:
I like the idea of the script. Next to the small feedback in Slack, I think we could also have a script to create the boilerplate folder & files (renderer / editor / index.ts). Not a high priority in any case, as right now, it doesn't happen too often to create a new plugin!

shn-srl:
> I like the idea of the script. Next to the small feedback in Slack, I think we could also have a script to create the boilerplate folder & files (renderer / editor / index.ts). Not a high priority in any case, as right now, it doesn't happen too often to create a new plugin!

Totally agree for the boilerplate too! Thanks for the feedback

I will close this PR and create a ticket for the idea

PR #3572, opened 2024-04-02
feat(oam): manually add some meta tags to exam urls

elbotho:


shn-srl (approved):

apps/web/src/data/de/math-exams-taxonomies.ts
```diff
@@ -9,3 +9,41 @@ export const mathExamsTaxonomies = [...mathExamTaxNI, ...mathExamTaxBY]
...
+    title: 'Mathe Abschlussprüfungen mit Lösung | Mittlere Reife',
+    metaDescription:
+      'Deine Vorbereitung für die Mittlere Reife Zweig i: Mathe lernen mit Original Prüfungsaufgaben. Mit Serlo schaffst du das!',
+  },
+  76750: {
+    title: 'Mathe Abschlussprüfungen mit Lösung | Mittlere Reife',
+    metaDescription:
+      '	Deine Vorbereitung für die Mittlere Reife Zweig ii und iii: Mathe lernen mit Original Prüfungsaufgaben. Mit Serlo schaffst du das!',
+  },
+} as const
```
shn-srl, line 49: Curious why `as const` is necessary, but not a blocker

elbotho: might not strictly be necessary in this case tbh but it's just so nice in many cases so I just add it most of the times. 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

PR #3591, opened 2024-04-04
refactor(editor-export): Simplify API by using default values for config

CodingDive:
Default values should allow us to change the API from

```TSX
<SerloEditor
  initialState={initialState}
  pluginsConfig={{
    basicPluginsConfig: {
      allowImageInTableCells: true,
      exerciseVisibleInSuggestion: true,
    },
  }}
>
```

to:

```TSX
<SerloEditor
  initialState={initialState}
>
```

### DIFF
--- a/packages/editor/src/editor-integration/create-basic-plugins.tsx
+++ b/packages/editor/src/editor-integration/create-basic-plugins.tsx
@@ -34,20 +34,20 @@ import { TemplatePluginType } from '@editor/types/template-plugin-type'
 export interface CreateBasicPluginsConfig {
   language?: SupportedLanguage
   allowedChildPlugins?: string[]
-  exerciseVisibleInSuggestion: boolean
+  exerciseVisibleInSuggestion?: boolean
   enableTextAreaExercise?: boolean
-  allowImageInTableCells: boolean
+  allowImageInTableCells?: boolean
   multimediaConfig?: MultimediaConfig
 }
 
 export function createBasicPlugins({
   language = 'de',
   enableTextAreaExercise = false,
-  exerciseVisibleInSuggestion,
+  exerciseVisibleInSuggestion = true,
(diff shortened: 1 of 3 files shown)


packages/editor/src/editor-integration/create-basic-plugins.tsx
```diff
@@ -34,20 +34,20 @@ import { TemplatePluginType } from '@editor/types/template-plugin-type'
...
 
 export function createBasicPlugins({
   language = 'de',
   enableTextAreaExercise = false,
-  exerciseVisibleInSuggestion,
+  exerciseVisibleInSuggestion = true,
+  allowImageInTableCells = true,
   allowedChildPlugins,
-  allowImageInTableCells,
   multimediaConfig,
```
LarsTheGlidingSquirrel, line 49: Can we have some default values here as well so that within the function nothing can be undefined any more. Then the whole default config would be centralized here.

packages/editor/src/package/editor.tsx
```diff
@@ -49,7 +49,7 @@ const emptyState = {
 /** For exporting the editor */
 export function SerloEditor(props: SerloEditorProps) {
   const { children, pluginsConfig, initialState, language = 'de' } = props
-  const { basicPluginsConfig, customPlugins = [] } = pluginsConfig
+  const { basicPluginsConfig, customPlugins = [] } = pluginsConfig || {}
```
LarsTheGlidingSquirrel, line 52: Would be great if `basicPluginsConfig` could get its default value here as well. See other comment for my thinking.

LarsTheGlidingSquirrel (approved):
Having default values is nice. Thank you! 
I still see some places where assigning the default values is spread across more than one location. Would be great if they could be assigned in one place only and stuff could not be undefined afterwards. But that can be a separate PR. :)

PR #3593, opened 2024-04-04
refactor(user-profiles): prefer `/user/profile/${username}` and use new api endpoint

elbotho:
first steps to prepare the frontend for a future where users are not uuids any more by using a separate api endpoint (`user{ userByUsername … }`) and giving users some own routes/components.

(only merge to prod after [api is deployed](https://github.com/serlo/api.serlo.org/pull/1344))


apps/web/src/components/user/user-unrevised-revisions.tsx
```diff
@@ -32,62 +28,54 @@ export function UserUnrevisedRevisions({
...
+  { unrevisedEntities: any }
+>
+type UsersRevision = DefinedUsersRevisions['unrevisedEntities']['nodes'][number]
+
+function useUserRevisionsFetch(username?: string) {
+  return useGraphqlSwrPaginationWithAuth<UsersRevision>({
+    query: usersRevisionsQuery,
+    variables: { username },
+    config: { refreshInterval: 1 * 60 * 1000 }, //1min
+    // @ts-expect-error not sure about this one
```
CodingDive, line 67: I also tried to solve this one to no avail. Best hack I could find instead of using @ts-expect-error is :man_shrugging: I don't think it's any better

```
    getConnection(data?: Record<string, unknown>) {
      const knownData = data as UsersRevisionsQuery
      return knownData?.user?.userByUsername?.unrevisedEntities
    },
```

apps/web/src/fetcher/create-title.tsx
```diff
@@ -12,6 +12,8 @@ export function createTitle(uuid: MainUuidType, instance: Instance): string {
   const instanceData = getServerSideStrings(instance)
   const suffix = ` - ${instanceData.title}`
 
+  // TODO: will be solved with upcoming graphql refactoring PR
+  if (uuid.__typename === UuidType.User) return 'Profil'
```
CodingDive, line 16: Very nice, when I first checked out the branch and went to http://localhost:3000/mathe, I guess the Next.js links of the team were still cached and upon clicking them, I got automatically taken from `/user/:id/ ` to the new `/user/profile/username`. Not sure if it's this line of code that is responsible for the redirect, but wanted to point out, how nice it worked even with some stale hrefs in my browser! :raised_hands: :partying_face:

CodingDive (approved):
Looks good to me! Some of the code, I saw for the first time and it took me quite a while to go through, but I couldn't find anything that needs to be changed and no bugs, no matter where on the Serlo platform, I click on a users profile! :blue_heart:

PR #3602, opened 2024-04-05
chore: use node 20

elbotho:


Entkenntnis:
a good moment to update my local node version 😅

PR #3606, opened 2024-04-05
refactor(editor-export): Implement Lars feedback and provide default values again

CodingDive:
I hope this addresses your comments https://github.com/serlo/frontend/pull/3591.

I think it should not change anything about the logic if I'm not mistaken, but maybe it'll prevent future bugs more easily and increase readability by having the default values in a centralized place.

Feel free to make the changes directly in the PR, if I missed something or you had something else in mind.

### DIFF
--- a/packages/editor/src/editor-integration/create-basic-plugins.tsx
+++ b/packages/editor/src/editor-integration/create-basic-plugins.tsx
@@ -40,6 +40,13 @@ export interface CreateBasicPluginsConfig {
   multimediaConfig?: MultimediaConfig
 }
 
+export const defaultPluginConfig = {
+  language: 'de' as const,
+  enableTextAreaExercise: false,
+  exerciseVisibleInSuggestion: true,
+  allowImageInTableCells: true,
+}
+
 export function createBasicPlugins({
   language = 'de',
   enableTextAreaExercise = false,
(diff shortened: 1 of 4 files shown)


packages/editor/src/editor-integration/create-basic-plugins.tsx
```diff
@@ -40,14 +40,21 @@ export interface CreateBasicPluginsConfig {
...
+  enableTextAreaExercise: false,
+  exerciseVisibleInSuggestion: true,
+  allowImageInTableCells: true,
+}
+
 export function createBasicPlugins({
   language = 'de',
   enableTextAreaExercise = false,
   exerciseVisibleInSuggestion = true,
   allowImageInTableCells = true,
```
CodingDive: Still want to add the default values here in case someone passes in e.g

```TypeScript
createBasicPlugin({ language: 'en' as const })
```

which would make all the booleans undefined, as the `defaultPluginConfig` won't be applied. Alternative to that, we could do the merge strategy, I sent you in slack.

CodingDive: This duplication makes me a little bit sad, but I still find it cleaner than 

```
...
enableTextAreaExercise = defaultPluginConfig.enableTextAreaExercise,
exerciseVisibleInSuggestion = defaultPluginConfig.exerciseVisibleInSuggestion,
...
```

shn-srl (approved):

PR #3609, opened 2024-04-08
fix(blanks-exercise): Make empty blanks not get deleted.

LarsTheGlidingSquirrel:
Fixes issue on iPad where a blank gets deleted even if there is content inside.

### DIFF
--- a/packages/editor/src/plugins/blanks-exercise/blank-renderer.tsx
+++ b/packages/editor/src/plugins/blanks-exercise/blank-renderer.tsx
@@ -141,8 +141,8 @@ export function BlankRenderer(props: BlankRendererProps) {
       correctAnswers
         // Trim the text inside of alternative answers inputs
         .map(({ answer }) => ({ answer: answer.trim() }))
-        // Filter out the empty alternative answers
-        .filter(({ answer }) => answer.length > 0)
+        // Filter out the empty alternative answers except the first one (index === 0)
+        .filter(({ answer }, index) => index === 0 || answer.length > 0)
     )
   }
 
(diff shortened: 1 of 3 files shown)


CodingDive:
I guess this will work and fixes the bug, however the right approach would be to somehow keep the focus on the iPad but may not be that trivial.
Should we include a check to only do that for `isSafari()` or will that be too confusing to have a different behavior between Safari and Chrome browsers?

LarsTheGlidingSquirrel:
> I guess this will work and fixes the bug, however the right approach would be to somehow keep the focus on the iPad but may not be that trivial. Should we include a check to only do that for `isSafari()` or will that be too confusing to have a different behavior between Safari and Chrome browsers?

In fact it, I saw it is not liked to Safari but the iPad. There the bug appears both on Safari & Chrome. I would recommend not implementing some platform specific logic unless we absolutely have to.

LarsTheGlidingSquirrel:
Not fully happy with the solution, but it works as a quick fix for now. 
Things I would like to do differently: Get rid of `blanksWithCorrectAnswer` as it makes the code harder to understand. Remove empty blanks (and clean up) on save.

elbotho:
> Not fully happy with the solution, but it works as a quick fix for now.
Things I would like to do differently: Get rid of blanksWithCorrectAnswer as it makes the code harder to understand. Remove empty blanks (and clean up) on save.

Agree. Moving complexity from the renderer to the editor is always preferable for me :)

elbotho (approved):
Seems to work well for me. Safari  on mac (ignoring some other inconsistencies) works fine as well.
Thank you for the quickfix.

PR #3611, opened 2024-04-08
Fix spoiler tracking - use path to check experiment id

shn-srl:


apps/web/src/mutations/use-experiment-create-exercise-submission-mutation.ts
```diff
@@ -13,8 +13,14 @@ const mutation = gql`
...
 `
 
-export function useCreateExerciseSubmissionMutation() {
+export function useCreateExerciseSubmissionMutation(path: string) {
   const mutationFetch = useMutationFetch()
+  const splitPath = path.split('/').filter((p) => p.length > 0)
+  const experimentIds = [30680, 23869, 66809]
+  const pathContainsExperimentId = splitPath.some((p) =>
+    experimentIds.includes(+p)
+  )
```
elbotho: maybe something like this would also work?
``` ts
const isExperimentPath = ['/30680', '/23869', '/66809'].find((slug) => path.includes(path))
```

elbotho (approved):
one suggestion. Take it or leave it :)

apps/web/src/mutations/use-experiment-create-exercise-submission-mutation.ts
elbotho: or
``` ts
const isExperimentPath = [30680, 23869, 66809].find((id) => path.includes(`/${id}`))
```

shn-srl: Thanks, done

PR #3614, opened 2024-04-09
refactor(editor-math): simplify math textarea, merge editor-textarea

elbotho:


packages/editor/src/math/math-editor-textarea.tsx
```diff
@@ -0,0 +1,68 @@
...
+          onMoveOutLeft?.()
+        }
+
+        if (e.key === 'ArrowUp' && !caretAtLeftEnd) {
+          e.stopPropagation()
+        }
+
+        if (e.key === 'ArrowDown' && !caretAtRightEnd) {
+          e.stopPropagation()
+        }
```
LarsTheGlidingSquirrel: I do not understand what these two are for. But, removing them changes nothing in behavior. Maybe obsolete?

LarsTheGlidingSquirrel (approved):
Thank you for the clean up :broom:

packages/editor/src/math/math-editor-textarea.tsx
elbotho: Thanks, I also did not find a change in behaviour and will try without.

PR #3618, opened 2024-04-10
feat: call kratos api endpoint manually after registering

elbotho:
not the nicest code because I have to inject this into the kratos handler but this way it's only happening on the server side and we don't need an additional (unprotected) nextjs endpoint.

(currently not working because of missing secret)

So far it only took 28 new users 😅

### DIFF
--- a/apps/web/src/components/pages/auth/registration.tsx
+++ b/apps/web/src/components/pages/auth/registration.tsx
@@ -91,7 +91,6 @@ export function Registration() {
       })
       .then(() => {
         setIsSuccessfullySubmitted(true)
-        window.scrollTo({ top: 0, behavior: 'smooth' })
       })
       .catch(
         handleFlowError(
@@ -103,6 +102,7 @@ export function Registration() {
       )
       .finally(() => {
         nProgress.done()
+        window.scrollTo({ top: 0, behavior: 'smooth' })
       })
   }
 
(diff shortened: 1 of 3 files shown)


hugotiburtino:
I've tested locally and it worked. But we will have to see it in production, because the error is very hard to reproduce. 

Before deploying, we will have to synchronize, because we'll have to remove the webhook at kratos configuration at the same time.

hugotiburtino (approved):

elbotho:
> I've tested locally and it worked. But we will have to see it in production, because the error is very hard to reproduce.

Nice! Happy to hear that.

> Before deploying, we will have to synchronize, because we'll have to remove the webhook at kratos configuration at the same time.

Is there a downside to having the endpoint being run twice for some time?

elbotho:
we should probably have a sentry call here if the endpoint is not called

hugotiburtino:
> Is there a downside to having the endpoint being run twice for some time?

Actually, it isn't a big deal, one of them is supposed to throw an error. I just want to avoid any kind of concurrence.
If the deployment has some minutes of delay, it's ok for me.

elbotho:
@hugotiburtino actually I could not get it to work locally, but since it worked for you I'd merge this to staging to test, okay?

Hugo's edit: We are still testing locally

PR #3620, opened 2024-04-10
refactor: use new path field, simplify breadcrumbs code

elbotho:


apps/web/src/fetcher/create-breadcrumbs.tsx
```diff
@@ -7,96 +7,69 @@ import {
...
+  //         label: metaMenuEntry.title,
+  //         url: current.alias,
+  //         id: current.id,
+  //       })
+  //       breadcrumbs.unshift({
+  //         label: matching2[0].rootName ?? getSubjectName(current),
+  //         url: matching2[0].landingUrl,
+  //         id: matching2[0].rootId,
+  //       })
+  //       …
```
elbotho: @Entkenntnis tbh. I don't really understand why this is / was needed.
Do you remember?

Entkenntnis: This affects situations like `FOS & BOS` where taxonomies are part of the secondary menu and need some kind of replacement.

grafik

Entkenntnis: If you don't do it, you will get the complete path

root > Mathe > Deutschland  > Bayern > `FOS & BOS`

elbotho: Thanks for the explanation!

elbotho:
@Entkenntnis I think I replicated the behaviour in https://github.com/serlo/frontend/pull/3620/commits/fb02b12e23a77bf76689e040feff0baea0437a8a maybe you could give it a quick glance.
(But most cases seem to work, so we can also merge it and test over time)

PR #3621, opened 2024-04-10
fix(course): Attempt to reset hash to show newly created course page

CodingDive:
Unfortunately, this solution does not work reliably and shows the `useLeaveConfirm` dialog randomly. Almost thinking of using the `#success` hash in the course, just so it leaves us alone when transitioning from one page view to another :grin: 

Maybe you can find a better solution!


elbotho:
see https://github.com/serlo/frontend/pull/3543#discussion_r1560084515

PR #3629, opened 2024-04-11
fix(equations): Reduce vertical space between arrow and explanation for long equations

LarsTheGlidingSquirrel:
## Before: 
image

## After: 
image

Not a super pretty solution but it works. If we want to do more in the future it might be easier to use css grids to position and align the individual elements. 

## Review: 
I tested a lot but maybe you can have a look if you find layout issues.


elbotho (approved):
Thank you!
Generally one of the plugins that should get a proper refactoring at some point so I'm okay with a not so pretty solution here.

PR #3659, opened 2024-04-16
chore(tests): remove unused fetcher tests

elbotho:
we did not run them for quite some time now so it's probably better to remove them, what do you think @Entkenntnis?


Entkenntnis (approved):
Thank you for house keeping. The e2e tests should cover all functionality

PR #3666, opened 2024-04-17
fix(plugin-text): latex cursor jump bug

hejtful:


elbotho (approved):
thank you very much for fixing my bugs 😬

PR #3690, opened 2024-04-22
fix(editor-package): Declare package explicitly as public 

CodingDive:
I saw that Github is running the `publish-to-npm` script, let's see if it works upon merge. If not, I'm about to publish it manually!

Attempts to fix ci
image


CodingDive:
Published it already here https://www.npmjs.com/package/@serlo/editor-web-component

CodingDive:
Even though the CI works already, I'd still like to have the visibility explicit in our package.json. Feel free to close, if you'd rather save a field!

elbotho (approved):

PR #3693, opened 2024-04-22
chore(editor-package): Pull katex-styles from npm serlo org

CodingDive:


packages/editor/package.json
```diff
@@ -39,9 +39,9 @@
     "yalc:publish": "yarn build && yalc publish --push --sig"
   },
   "dependencies": {
+    "@serlo/katex-styles": "^1.0.0",
```
elbotho: why is this package in [a different repo](https://github.com/serlo/serlo-katex-styles) and not part of the monorepo?

elbotho: (then we could also use it with `workspace:*` here)

CodingDive: > (then we could also use it with workspace:* here)

Not with this package, but I ran into some trouble using a yarn workspace as a normal (non dev) dependency and then publishing to npm. It broke npm/pnpm in resolving the dependency tree when I was trying to install from the registry.

That's not to say, I can't make this part of the monorepo but we would probably have to use vite & ts paths to alias the import from the monorepo package, have another package.json + all the config boiler plate files and run the risk of having more of a Eslint/TypeScript mess.

I'd prefer it in an extra repo if it's not a file we are touching too often.

elbotho: Hmm I don't veto this. But I want to make sure…
In my opinion small packages like that are a perfect fit for the monorepo concept and part of the reason why we set it up.
We don't have to use `workspace:*` as the source if that makes it easier.

CodingDive: Cool! Will try to set it up tomorrow :blush:

elbotho (approved):
thanks for moving it into the monorepo :)

PR #3695, opened 2024-04-22
chore: next to v14, try scrollRestoration

elbotho:
smaller bundles is a good start :) and scroll restoration works fine for me


elbotho:
I started this investigation for mathe-skills, hopefully we can use the platform fix.

Entkenntnis (approved):
Thank you for the upgrade, weird that scroll restoration is still experimental 😅

PR #3696, opened 2024-04-22
fix(editor-web-component): Use fixed editor version & remove build before linting 

CodingDive:
- [x] Publish new editor version to get the updated types see https://github.com/serlo/frontend/pull/3698

This should also fix a critical problem that I ran into when installing the package from npm.

```
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:*
```

Updated readme for local development


CodingDive:
This is a major footgun and we have to be very careful. I had to reset my `yarn cache` because of the name collision of `@serlo/editor` from the workspace (that was defined before this PR) to now the `@serlo/editor` from the npm registry. I ended up with the wrong `editor` package in my `node_modules` and ran into the weirdest eslint errors consequently! :scream: 

If you or anyone has trouble with the `editor-web-component` package locally, you may also need to reset the yarn cache and reinstall.

Now it seems to work great, and linter is always running through smoothly!

elbotho (approved):

PR #3699, opened 2024-04-22
chore: minor upgrades

elbotho:
follow ups: there are some major version waiting and some problems to solve with newer react type versions.


apps/web/src/components/auth/node.tsx
```diff
@@ -168,6 +168,8 @@ export function Node({
     }
   }
   // Anchor, Image, Script, Text
+  // eslint-disable-next-line no-console
+  console.log(node)
```
hejtful, line 172: Leftover maybe?

hejtful (approved):
Thank you 🙏

PR #3706, opened 2024-04-23
fix(injection): Not use `TemplatePluginType.TextExerciseGroup` for `exerciseGroup`

LarsTheGlidingSquirrel:
Makes the injection plugin render `exerciseGroups` for example `6193`

Not sure if this solution is good however because I understand too little about why we need to add the `licenceId` and what the difference between `exerciseGroup` and `type-text-exercise-group` is.


elbotho:
@shn-srl could you check if this works with the exercise analytics?

elbotho (approved):

elbotho:
Let's see if Shane has feedback, but it solves an annoying but, so I'll merge it now.

shn-srl:
> Let's see if Shane has feedback, but it solves an annoying but, so I'll merge it now.

LGTM

PR #3710, opened 2024-04-24
WIP: Reduce bundle size in editor package

CodingDive:
- [ ] Find better way than top level await
- [ ] Revert changes from target ES6 to ESNext


CodingDive:
Closing in favor of https://github.com/serlo/frontend/pull/3750. Still interested in reducing the bundle size but lower priority than some of the other tasks

PR #3716, opened 2024-04-25
Remove Prisma package

shn-srl:
- Remove Prisma from frontend as it is no longer used


elbotho (approved):
thanks!

Entkenntnis (approved):
sad to see it go, though - was a nice enabler for quick stuff

elbotho:
@Entkenntnis there is still
- https://turso.tech/
- https://neon.tech/ 
- https://www.cockroachlabs.com
- …

with free tiers. Sure they also might change their pricing models at some point but for experiment's I'm all for it 😀

elbotho:
But maybe we can go with `drizzle` next time :)

PR #3728, opened 2024-04-26
chore(math-skills): use other alpha symbol (⍺) for readability

elbotho:


Entkenntnis (commented):
Can't see a difference in the source code 😅

elbotho:
image
Interesting, the fonts again probably…

Entkenntnis:
github monospaced font
grafik

This is how it looks in-app:

grafik


This is another variant by Yannick, probably bold-italic?

grafik

PR #3730, opened 2024-04-26
feat: move lenabi user journey to staging (wip)

elbotho:
- unfortunately the new app does not work on my phone for some reason so I'll have to find a way (or a person) to test it
- very little code should make it into the production bundle

for https://miro.com/app/board/uXjVKe_jgYA=/

~I'm waiting for exercises to be allowed in coursePages in production, so we can replace the blanks mockup with a real blanks-exercise. (Already worked in staging)~
Example course: https://de.serlo.org/community/307527/teste-dein-wissen


hugotiburtino:
> unfortunately the new app does not work on my phone for some reason so I'll have to find a way (or a person) to test it

It worked on my phone

hugotiburtino:
> I'm waiting for exercises to be allowed in coursePages in production

It isn't a blocker from backend, is it? If it is, could you point me what should be deployed?

elbotho:
> It isn't a blocker from backend, is it? If it is, could you point me what should be deployed?

No, frontend is are currently waiting for content-people to give feedback.

apps/web/src/components/auth/flow.tsx
```diff
@@ -204,6 +205,10 @@ export function handleFlowError<S>(
           unwantedPaths: [verificationUrl, loginUrl, registrationUrl],
         })
         setTimeout(() => {
+          if (!isProduction && window.location.host.startsWith('journey')) {
+            window.location.href =
+              'https://journey.serlo-staging.dev/willkommen'
+          }
```
hugotiburtino: TODO: delete it before merging

apps/web/src/components/pages/auth/login.tsx
```diff
@@ -168,6 +169,10 @@ export function Login({ oauth }: { oauth?: boolean }) {
           showToastNotice(
             strings.notices.welcome.replace('%username%', username)
           )
+          if (!isProduction && window.location.host.startsWith('journey')) {
```
hugotiburtino: Instead we can check if the authentication provider is nbp

PR #3731, opened 2024-04-29
refactor: use react-hot-toast instead of forked outdated package

elbotho:
Instead of having [our own fork of an unmaintained package](https://github.com/serlo/frontend/wiki/Forked-packages#react-notify-toast) refactoring to use package that is [at least somewhat maintained](https://github.com/timolins/react-hot-toast) 😅. 

Downside: ~1kb bundle size increase.


CodingDive:
> Downside: ~1kb bundle size increase.

Apparently, you can get some money for reducing the bundle size of their only dependency! :grinning: https://goober.rocks/the-great-shave-off

Entkenntnis (approved):
notification queue looks good - one outdated dep less

PR #3732, opened 2024-04-29
fix(plugins): Add missing ids to videos, injection and geogebra plugins

CodingDive:


apps/web/src/fetcher/graphql-types/operations.ts
```diff
@@ -2853,4 +2853,4 @@ export type InjectionOnlyContentQueryVariables = Exact<{
...
... | { __typename: 'Event', id: number, licenseId: number, alias: string, title: string, currentRevision?: { __typename?: 'EventRevision', content: string } | null } | { __typename: 'EventRevision', alias: string, title: string } | { __typename: 'Exercise', id: number, licenseId: number, alias: string, title: string, currentRevision?: { __typename?: 'ExerciseRevision', content: string } | null } | { __typename: 'ExerciseGroup', id: number, licenseId: number, alias: string, title: string, currentRevision?: { __typename?: 'ExerciseGroupRevision', content: string } | null } | { __typename: 'ExerciseGroupRevision', alias: string, title: string } | { __typename: 'ExerciseRevision', alias: string, title: string } | { __typename: 'Page', alias: string, title: string } | { __typename: 'PageRevision', alias: string, title: string } | { __typename: 'TaxonomyTerm', alias: string, title: string } | { __typename: 'User', alias: string, title: string } | { __typename: 'Video', id: string, licenseId: number, alias: string, title: string, currentRevision?: { __typename?: 'VideoRevision', content: string, url: string } | null } | { __typename: 'VideoRevision', alias: string, title: string } | null };
```
CodingDive, line 2856: I assume I shouldn't change this, but a lot of ids are classified as `id: number`. Therefore, I either had to go in and do 
`id: uuid.id as unknown as number`, or change it here. Where would I need to update the types to do it properly? In the API?

apps/web/src/components/pages/editor/education-plugins.tsx
```diff
@@ -95,7 +95,11 @@ const pluginData = [
       'With the GeoGebra applet, moving or interactive graphics can be integrated into the content. You can create your own applet or choose from the huge free public applet collection on geogebra.org.',
     example: (
       <StaticRenderer
-        document={{ plugin: EditorPluginType.Geogebra, state: 'd4eNMF5R' }}
+        document={{
+          plugin: EditorPluginType.Geogebra,
+          state: 'd4eNMF5R',
+          id: '1e633412-430d-453f-b621-589a05dd139e',
```
CodingDive, line 101: As it's just an example, I presume using a hard coded id is fine? If not, I can use the uuidv4() fn

elbotho: I think this is a big misunderstanding.
https://github.com/serlo/frontend/pull/3715 is just a temporary branch (that's why I called it tmp:*) to check if the db-migration was successful. I never had the intention to merge it.
I thought about changing the type to make id not optional but that changes core editor behaviour where you can init a new plugin just by providing a plugin type (e.g. `{plugin: EditorPluginType.Article}` is a valid state) which is a nice and simple behaviour imho.

CodingDive:
Closing because we can do this at some point in the future after the migration! 
This PR serves as a good starting point of the changes we'll need to make

PR #3733, opened 2024-04-29
web component: simplify API and expose read/write mode to render static renderer or editor respectively

CodingDive:


LarsTheGlidingSquirrel (approved):
Looks good to me. Thank you!
Just one question: Can we have a small bundle size when being only in 'read' mode using this approach?

yorrd:
@LarsTheGlidingSquirrel yes and no, depends on your build process. Afaik youre using esbuild below vite, so it can work, but only with async chunk loading. When importing the module dynamically before using it, it can be loaded via async request, but that also requires the hosting project to set that up. A "simple" way would be to expose two webcomponents, one for rendering and one for editing. Obviously from an API point of view, @CodingDive 's approach is preferrable. Not an easy decision, I would opt for async loading the chunks via `import()` when the time comes to optimize bundle size. 

Also, there's an avenue here to just prerender the HTML entirely and provide it via a simple functional helper without any webcomponent around it. That's only viable with inlined styles though. Just leaving that here for completeness' sake, I would prefer the encapsulated way (which is still free to load prerendered html if so desired). The whole approach would require a server infra though which I don't think is worth the effort atm.

packages/editor-web-component/README.md
```diff
@@ -5,16 +5,17 @@ This is an early version of the web component wrapping the [Serlo Editor](https:
...
 import { EditorWebComponent } from '@serlo/editor-web-component'
 
+customElements.define('serlo-editor', EditorWebComponent)
+
 // in your render function
 return (
-  <div id="serlo-root">
-    <serlo-editor></serlo-editor>
-  </div>
+  <serlo-editor></serlo-editor>
```
yorrd, line 18: love it!

packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -28,12 +30,12 @@ export class EditorWebComponent extends HTMLElement {
```
yorrd, line 30: 'mode' missing in `observedAttributes` leads to changes in that attribute not being respected. Therefore, only the initial value is respected atm

yorrd (changes requested):
only the attribute changed thing, otherwise lgtm

PR #3734, opened 2024-04-29
refactor: simplify exercise keys / ids

elbotho:
merge after [migration](https://github.com/serlo/db-migrations/pull/339) ran in prod


CodingDive (approved):
Looks good to me and works great! Thank you! I merged latest staging into this branch, because I wasn't sure if there would be some conflicts

What would be the next steps of simplifications? I assume we can remove all these optional ids from here and mark them as required too? https://github.com/serlo/frontend/blob/staging/packages/editor/src/types/editor-plugins.ts#L72

PR #3738, opened 2024-04-30
chore(editor): activate AnchorLinkCopyTool for everyone

elbotho:
Feature was an experiment before (since we could not rely on the editor-ids to be present).
Now after the migration this works a lot nicer.

image

(has crowdin changes)

### DIFF
--- a/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
+++ b/packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
@@ -1,5 +1,4 @@
 import { faHashtag } from '@fortawesome/free-solid-svg-icons'
-import { shouldUseFeature } from '@serlo/frontend/src/components/user/profile-experimental'
 import { useInstanceData } from '@serlo/frontend/src/contexts/instance-context'
 import { useEditorStrings } from '@serlo/frontend/src/contexts/logged-in-data-context'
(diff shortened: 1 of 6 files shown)


packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
```diff
@@ -18,11 +17,7 @@ export function AnchorLinkCopyTool({
   const editorStrings = useEditorStrings()
   const { strings } = useInstanceData()
 
-  if (
-    !navigator.clipboard ||
-    !window.location.href.includes('add-revision') ||
-    !shouldUseFeature('editorAnchorLinkCopyTool')
-  ) {
+  if (!navigator.clipboard || !window.location.href.includes('add-revision')) {
```
LarsTheGlidingSquirrel, line 21: Maybe add a comment explaining why only on paths containing `add-revision`: 
Makes only sense when we have an entity id. We don't have one on `/___editor_preview` for example. Also in `serlo-editor-for-edusharing`.

LarsTheGlidingSquirrel (approved):
Looks good! Thank you! :green_heart: 
Added one comment.

packages/editor/src/editor-ui/plugin-toolbar/plugin-tool-menu/anchor-link-copy-tool.tsx
elbotho: thanks!

PR #3745, opened 2024-05-01
Remove validity check from exercise submission

shn-srl:
Comment in code explains reasoning.

Lösungs Vorschlag is not sending a revision id at the moment.

Sense of urgency to get the experiment tracking working, would look at the revision id issue more closely when I'm back.

I am not working 02.05


Entkenntnis (approved):
👍

PR #3750, opened 2024-05-03
Web Component: Always define process env

CodingDive:


yorrd (approved):
ah directly in the build step. Makes sense. I'm not a vite export, but LGTM.

PR #3751, opened 2024-05-06
chore(editor): activate editorPluginCopyTool for all users

elbotho:
https://serlo.slack.com/archives/CEA1UMW5R/p1705927815805859


CodingDive (approved):
Works amazing! Still feels like magic when I copy a giant plugin, hit paste and it just appears! :sparkles:

PR #3753, opened 2024-05-06
refactor(editor): Publish css extra & release v0.7.0

CodingDive:
The web component will require the css to be exported and published to npm. Meanwhile, the editor package and package consumers would best like to have the css served by the JS bundle (cssInjectedByJsPlugin) which I now removed.

Unfortunately, I'm not sure if this means that we'll need to import the css from edusharing extra.

If we don't want to do that (as it would make the API more complex instead of simplifying it), we may need to have two modes in the vite.config of the editor package. One for including the css and one for injecting it within the JavaScript plugin. What do you think is the right move forward here?

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@serlo/editor",
-  "version": "0.6.0-beta.17",
+  "version": "0.7.0",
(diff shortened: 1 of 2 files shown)


LarsTheGlidingSquirrel (approved):
Fine from my side! Maybe we will find a better solution later.

PR #3758, opened 2024-05-08
feat: integrate VIDIS

hugotiburtino:
There is a way of using the a web component through a CDN, could we try it out? 
I suppose it is important for brand UI button.
```
VIDIS-Login Example
<script src="https://repo.vidis.schule/repository/vidiscdn/latest/vidisLogin.umd.js"></script>
<vidis-login loginurl="https://www.domain.de/path-toauth/" size="L" cookie="true"></vidis-login>
```

### DIFF
--- a/apps/web/src/components/pages/auth/logout.tsx
+++ b/apps/web/src/components/pages/auth/logout.tsx
@@ -32,23 +32,30 @@ export function Logout({ oauth }: { oauth?: boolean }) {
     checkInstance({ redirect: false })
 
     const cookieSession = AuthSessionCookie.parse()
-    const isSSO = cookieSession?.authentication_methods
-      ? cookieSession.authentication_methods[0].method === 'oidc'
-      : false
+    const ssoProvider = cookieSession?.authentication_methods
+      ? cookieSession.authentication_methods[0].provider
+      : undefined
 
-    const redirection = isSSO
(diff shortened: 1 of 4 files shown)


apps/web/src/components/pages/auth/logout.tsx
```diff
@@ -32,23 +32,30 @@ export function Logout({ oauth }: { oauth?: boolean }) {
...
+      ? cookieSession.authentication_methods[0].provider
+      : undefined
 
-    const redirection = isSSO
-      ? 'https://aai.demo.meinbildungsraum.de/realms/nbp-aai/protocol/openid-connect/logout'
-      : filterUnwantedRedirection({
-          desiredPath: sessionStorage.getItem('previousPathname'),
-          unwantedPaths: [settingsUrl, loginUrl, registrationUrl],
-        })
+    const redirection =
```
shn-srl, line 39: Nitpick: Can we make this a key value map or something cleaner than the nested ternary's?

hugotiburtino: I've just kept the style convention of the repo, but sure you can refactor it. I think a function would be the most readable refactoring.

hugotiburtino:
If the integration of the vidis webcomponent isn't that easy, could we merge to start testing it and do the changes afterwards?

hugotiburtino:
I'll merge it and do the improvements in other PR

PR #3779, opened 2024-05-13
chore: Upgrade yarn to v4.2.2

CodingDive:

PR #3780, opened 2024-05-13
Fix ci force yarn to pull package from registry over workspace

CodingDive:
If we are not yet ready for berry v4, here is a fix with the current yarn version (3.6.x) 

Replaces https://github.com/serlo/frontend/pull/3779


elbotho (approved):
nice, I like that this works separate to the decision to move to yarn v4

PR #3783, opened 2024-05-13
types(e2e): Add types and set default timeout to 30s

CodingDive:
After the CI kept running for 3hrs, I wanted to introduce a default timeout. I think it does probably have one by default which makes me wonder why the tests didn't just fail :thinking:


e2e-tests/codecept.config.ts
```diff
@@ -40,7 +41,9 @@ exports.config = {
       inject: 'login',
       users: {
         admin: {
-          login: (I) => {
+          login: (I: CodeceptJS.I) => {
+            I.amOnPage('/')
+            I.amOnPage('/')
```
elbotho: is that line doubled on purpose?

CodingDive: Yikes! Great find, thank you :pray: Fixed now

elbotho (approved):
hard to judge without knowing the default timeout, but it's save enough to try :)

CodingDive:
Global expect timeout should be 5 seconds https://playwright.dev/docs/test-timeouts#introduction

Issue #1985, opened 2023-01-02
remove .babelrc


Entkenntnis:
there are only a way places we actually use `@apply` within styled-jsx.

If we remove them, we can finally remove `.babelrc` and switch to SWC for the main project

elbotho:
Isn't postcss (nesting) the reason we can't switch?

Entkenntnis:
ah right, this was the other reason I completely forgot ...

so no easy fix, unfortunately, see https://github.com/vercel/styled-jsx/issues/782

PR #3785, opened 2024-05-14
chore: remove babel dependency

elbotho:
@Entkenntnis do you remember why we kept it? It now seems to work fine without it but I don't really trust it :)

- edit: found the older issue https://github.com/serlo/frontend/issues/1985
- Vercel Docs on [Babel](https://nextjs.org/docs/pages/building-your-application/configuring/babel) and [PostCSS](https://nextjs.org/docs/pages/building-your-application/configuring/post-css)


Entkenntnis:
In the frontend, we are not using babel anymore (we manually unnested all styled-jsx css 😅) - I would take a closer look at jest though and check if it still works without babel.

elbotho:
> In the frontend, we are not using babel anymore (we manually unnested all styled-jsx css 😅) - I would take a closer look at jest though and check if it still works without babel.

`yarn test` runs fine locally and in the CI, anything else to check?

Entkenntnis:
this should be fine 👍

PR #3795, opened 2024-05-15
chore(editor-web-component): Mark all dev dependencies

kulla:
In my understaning everything is bundled together, right? Then those dependencies should be dev dependencies.


CodingDive (approved):
I haven't considered it, thank you!

PR #3799, opened 2024-05-16
Fix/plugin suggestion menu in shadow dom

CodingDive:
It used to get rendered at the top of the webpage. Problem: Within the shadow DOM, we don't have access to the current selection (`window.getSelection()` defaults to document.body). My changes try to find the active element within the Shadow DOM and based on that, calculates the height offsets from there.


LarsTheGlidingSquirrel (approved):
Code looks good. But I haven't tested it.

PR #3800, opened 2024-05-16
refactor(editor-template-plugins): share components

elbotho:
some preparations for a potential future where the template plugins might be unified completely or moved out of the editor (and into the frontend).


elbotho:
@LarsTheGlidingSquirrel I could also imagine a refactoring that moves all fields from `entity` and from the `serloContext` hack into a frontend-only provider. Something like `SerloEntityProvider`. 

``` ts
export const entity = {
  ...uuid,
  revision: number(),
  licenseId: optional(number()),
  changes: string(),
  meta_title: optional(string()),
  meta_description: optional(string()),
}
```

PR #3803, opened 2024-05-17
chore/slate upgrade to v0.103.0

CodingDive:
# Fixes 

- [x] Shadow DOM firefox: Text placeholder does not disappear as the user is typing
- [x] Shadow DOM: Backspace key in firefox can't be pressed
- [x] Shadow DOM: Big focus issues in firefox in any text plugin (cursor jumping back to the beginning of the line)
- [x] Editor web component:  Allow nested plugins in spoiler
- [x] Suggestion menu (slate-overlay) being positioned wrongly
- [x] Clicking "+" button, causes focus issue
- [x] Slate to DOM node can't be resolved when pressing ArrowKey over the equation plugin (duplicated focus event was called, thank you @hejtful for finding and fixing this with me)
- [x] Slate to DOM node resolution blowing up when pressing the ArrowKey down very quickly (needed a timeout and ensure the whole editor is mounted) 
- [x] Race condition that lead to recursive focus issues when focusing multiple items within <10ms, such as when the ArrowUp key is pressed (fixed by clearing unresolved timeouts) 

Note that the two Slate to DOM node issue happened particularly on the equation plugin. They looked like this 

```
Error: Cannot resolve a DOM node from Slate node: {"children":[{"type":"p","children":[{"text":"let's go"}]}],"operations":[],"selection":{"anchor":{"path":[0,0],"offset":0},"focus":{"path":[0,0],"offset":0}},"marks":null}
``` 

Please review thoroughly, I may have introduced some bugs in fixing the issues above. The editor-web-component now is so much more usable (especially in Firefox), so I'd like to release a new editor-web-component version as soon as possible.

## Automated Tests

I ran and fixed the e2e tests for Chrome. There are 2-3 tests that are failing locally in `e2e-firefox`, but looking at the logs, I don't think they are related.


elbotho:
> chrome: focus gets stuck at hyphen in table when pressing ArrowKey down

I think I have the same behaviour in staging. Makes sense to compare against the status quo here and only track new problems imho.

CodingDive:
> I think I have the same behaviour in staging. Makes sense to compare against the status quo here and only track new problems imho

Oops, you are right! I did check staging before and thought it worked there, but I must have pressed the right arrow key. Can reproduce now myself, that it's happening in staging already.

e2e-tests/README.md
```diff
@@ -41,3 +41,33 @@ yarn e2e-firefox
...
+
+Then ensure to uncomment this line in the .env file.
+
+`FRONTEND_URL=http://localhost:3000`
+
+Finally, run the tests:
+
+```sh
+yarn e2e
+```
```
elbotho: That's why we have `yarn e2e-localhost`. Is there a reason you are recommending the env approach here?

elbotho: also at least for many tests dev server also works. So if you are working on a certain test/feature building in between might not be necessary.

packages/editor/src/plugins/text/components/math-element.tsx
```diff
@@ -53,7 +53,11 @@ export function MathElement({
       <>
         {/* Slate void elements need to set attributes and contentEditable={false}
           See: https://docs.slatejs.org/api/nodes/element#rendering-void-elements */}
-        <span {...attributes} contentEditable={false}>
+        <span
+          {...attributes}
+          contentEditable={false}
+          style={{ userSelect: 'none' }}
```
elbotho:
```suggestion
          className="select-none"
```

packages/editor/src/plugins/text/components/text-leaf-with-placeholder.tsx
```diff
@@ -30,6 +30,7 @@ export function TextLeafWithPlaceholder(
       <span
         className="pointer-events-none absolute -mt-1 block w-full text-gray-400 [user-select:none]"
         contentEditable={false}
+        style={{ userSelect: 'none' }}
```
elbotho:
```suggestion
        className="pointer-events-none absolute -mt-1 block w-full text-gray-400 select-none"
        contentEditable={false}
```

This should work, did it not?

elbotho (commented):
Nice, thanks for seeing this through!
Did comment some small stuff and did some cursory testing in chrome, safari and ff.
I found lots of bugs, but so far no new ones 😉.

e2e-tests/README.md
CodingDive: I didn't know about this command :facepalm: Thank you!

hejtful (approved):
Code looks good to me :)

PR #3821, opened 2024-05-23
Set up plausible analytics

shn-srl:
This is an MVP for testing, if this test works, I will extend the test for tracking the other relevant events with Felix.

https://plausible.io/simple-web-analytics

### DIFF
--- a/apps/web/src/pages/_document.tsx
+++ b/apps/web/src/pages/_document.tsx
@@ -122,7 +122,6 @@ export default class MyDocument extends Document {
           ></script>
 
           {process.env.NEXT_PUBLIC_USE_PLAUSIBLE_ANALYTICS !== undefined && (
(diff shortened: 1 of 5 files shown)


apps/web/src/pages/_document.tsx
```diff
@@ -120,6 +120,15 @@ export default class MyDocument extends Document {
               __html: `window.sa_event=window.sa_event||function(){a=[].slice.call(arguments);sa_event.q?sa_event.q.push(a):sa_event.q=[a]};`,
             }}
           ></script>
+
+          {process.env.NEXT_PUBLIC_USE_PLAUSIBLE_ANALYTICS !== undefined && (
```
elbotho, line 124: why are you using a env var for this? It should only run on staging not prod, right? 
Then the `isProduction` helper should be fine as a check.

PR #3824, opened 2024-05-23
Replace graphql mutation with plausible analytics call

shn-srl:
Replace GraphQl mutation which stored experiment data in GCloud, now it sends events to plausible analytics instead


CodingDive (approved):
LGTM for staging. Reviewed it here on Github, but I didn't run it locally

PR #3825, opened 2024-05-23
fix(editor-web-component): Focus issue

CodingDive:
* Lazy load editor
* Find shadow DOM within useEffect and timeout, otherwise things are not working reliably
* Release new version of editor - web component release will come as a follow up


packages/editor/src/core/helpers/use-get-shadow-root.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from 'react'
+
+export function useGetShadowRoot(containerRef: React.RefObject<HTMLElement>) {
```
hejtful: Minor: This might also simply be called `useShadowRoot`, since it's returning the shadow root itself, and not a getter for it.

packages/editor/src/core/hooks/use-blur-on-outside-click.ts
```diff
@@ -21,16 +24,22 @@ export function useBlurOnOutsideClick(
         !editorWrapperRef.current.contains(clickedElement) && // clicked element is not a child of the provided wrapper
         !clickedElement.closest('.ReactModalPortal') // clicked element is not a part of a modal
       ) {
+        console.log('Outside click detected, blur editor', {
```
hejtful: A leftover maybe?

yarn.lock
```diff
@@ -4848,7 +4848,7 @@ __metadata:
   resolution: "@serlo/editor-web-component@workspace:packages/editor-web-component"
   dependencies:
     "@rollup/plugin-replace": ^5.0.5
-    "@serlo/editor": 0.7.0
+    "@serlo/editor": 0.7.2
```
hejtful, line 4851: In the editor package.json, it changes to 0.7.3, but here it changes to 0.7.2, is that expected?

hejtful (approved):
Nice, great work on fixing the bugs and finding a clean way of extracting the fix into a hook! Just a couple of minor comments

PR #3841, opened 2024-05-27
fix(editor-web-component): AST manipulation to exclude global styles from build

CodingDive:
Custom rollup plugin removes global styles so they don't cause damage outside of the Shadow DOM


LarsTheGlidingSquirrel (approved):
Hm, seems like it creates a bunch of complexity. Maybe we can think of a less complex solution for the future. To me it seems like requiring the integrations to add global styles manually is way less complex. But great that you found a solution for now.

CodingDive:
> Hm, seems like it creates a bunch of complexity. Maybe we can think of a less complex solution for the future. To me it seems like requiring the integrations to add global styles manually is way less complex. But great that you found a solution for now.

I agree! It's a shame to have to add this complexity, but I think some build complexity is fine for easier integrations. I talked this through with Vito and we really don't want React package consumers to have to add styles themselves. I can modularize this code a little bit more, to not have everything in a big vite configuration file!

PR #3842, opened 2024-05-27
 fix(editor-web-component): Render modal inside shadow root

CodingDive:
Migrated to radix/react-dialog. React-modal was really outdated and the `appElement` was killing our integrations. 

Here are the docs. https://www.radix-ui.com/primitives/docs/components/dialog  Behavior and styles should stay the same.


packages/editor/src/plugins/audio/toolbar.tsx
```diff
@@ -26,6 +31,8 @@ export const AudioToolbar = ({
   setShowSettingsModal: Dispatch<SetStateAction<boolean>>
 }) => {
   const audioStrings = useEditorStrings().plugins.audio
+  const ref = useRef<HTMLButtonElement>(null)
+  const shadowRoot = useShadowRoot(ref)
```
elbotho: I don't understand why you pass the opener-button ref as the containerRef to `useShadowRoot`. Could you explain?
Also I'd love to move this complexity inside the modal instead of having it in all the consumers. Would that be possible somehow?

CodingDive: We just need any ref to detect whether it's inside the shadow root or not. 

>Also I'd love to move this complexity inside the modal instead of having it in all the consumers. Would that be possible somehow?

Maybe rendering an empty div inside the modal with a ref. That could be enough! Will try!

CodingDive: Worked great! Thank you for the suggestion :heart:

elbotho: Nice, happy about that!

packages/editor/src/core/helpers/use-shadow-root.ts
```diff
@@ -22,3 +22,19 @@ export function useShadowRoot(containerRef: React.RefObject<HTMLElement>) {
...
+
+export function getFirstElementOrUndefined(
+  shadowRoot: ShadowRoot | null | undefined
+): HTMLElement | undefined {
+  if (!shadowRoot) {
+    return undefined
+  }
+
+  return shadowRoot?.firstElementChild as HTMLElement
+}
```
elbotho, line 40: possible simplifications:

```suggestion
export function getFirstElementOrUndefined(shadowRoot?: ShadowRoot | null) {
  return shadowRoot?.firstElementChild as HTMLElement | undefined
}
```

packages/editor/src/core/hooks/use-blur-on-outside-click.ts
```diff
@@ -22,7 +22,7 @@ export function useBlurOnOutsideClick(
         root.contains(clickedElement) && // clicked element is present in the document
         editorWrapperRef.current && // provided wrapper is defined
         !editorWrapperRef.current.contains(clickedElement) && // clicked element is not a child of the provided wrapper
-        !clickedElement.closest('.ReactModalPortal') // clicked element is not a part of a modal
+        !clickedElement.closest('.serlo-modal') // clicked element is not a part of a modal
```
elbotho, line 35: nice that you thought of those

elbotho (commented):
did not test yet, but code looks good!

elbotho:
Some changes in behaviour: 

**Share modal**
When I open the share modal for  the first time the close button is focused / has an outline:

image

----

**Editor setting modal** (e.g. image-plugin):
restoring/keeping focus works only when I use the close button. If I click outside the focus goes to document.body.
(works in staging)

----

**Editor Article Add Modal** ([example](https://frontend-git-fix-editor-web-component-modal-in-sha-e87021-serlo.vercel.app/entity/repository/add-revision/1565)):
Also looses focus to body on closing (close button and outside click)

image

CodingDive:
Thank you for the thorough review @elbotho 

> When I open the share modal for the first time the close button is focused / has an outline

According to aria recommendations, one should focus the first focusable element. That should clearly not be the close button though! Radix was just picking the first element, so I moved the button down.

https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

Now when I share, the URL gets focused.

image


> Editor Article Add Modal ([example](https://frontend-git-fix-editor-web-component-modal-in-sha-e87021-serlo.vercel.app/entity/repository/add-revision/1565)):
> Also looses focus to body on closing (close button and outside click)

Should be upon ESC/closing via button, but the outside click does not yet work. 

> restoring/keeping focus works only when I use the close button. If I click outside the focus goes to document.body.
(works in staging)

I believe the reason why this does not yet work, is that the toolbar looses focus through `useBlurOnOutsideClick` before the `useEffect` https://github.com/serlo/frontend/pull/3842/files#diff-1c5900ff0d9c5494c672ce3039511959d8a50fe827a74bf903a43c75ed06f787R49 I just added has time to execute. One risky fix I can think of is to restore focus on unmount, but I feel this could have a lot of unintended consequences.

CodingDive:
Please give it another round of testing! I hope I addressed all the issues

apps/web/src/components/user-tools/revision/checkout-reject-buttons.tsx
```diff
@@ -62,7 +62,7 @@ export function CheckoutRejectButtons({
 
       <ModalWithCloseButton
         isOpen={modalMode !== null}
-        onCloseClick={onCloseClick}
+        setIsOpen={(open) => (open ? void null : onCloseClick())}
```
shn-srl, line 65: Maybe we can have setIsOpen as an optional prop? Then here we could have a cleaner prop,


`open && onCloseClick()`

apps/web/src/components/modal-with-close-button.tsx
```diff
@@ -22,74 +24,120 @@ interface ModalWithCloseButtonProps {
...
+    (open: boolean) => {
+      if (!open && confirmCloseDescription) {
+        setShowConfirmation(true)
+        return
+      }
+
+      // Need to add this tiny delay when unmounting the modal, so that the
+      // use-blur-on-outside-click has time to detect whether the modal is open
+      // or not.
+      if (open === false) {
```
shn-srl, line 57: Nit: Can we return early here too? 

```
if(open !== false){
 setIsOpen(open)
 return
}

...
```

apps/web/src/components/modal-with-close-button.tsx
```diff
@@ -22,74 +24,120 @@ interface ModalWithCloseButtonProps {
...
+      } else {
+        previouslyFocusedElementRef.current =
+          document.activeElement as HTMLElement
+      }
+    } else {
+      if (previouslyFocusedElementRef.current) {
+        previouslyFocusedElementRef.current.focus()
+      }
+    }
+  }, [isOpen, shadowRoot])
```
shn-srl, line 82: Nitpick: Return early would be nice here too, to reduce nesting avoid the need for the else of the outer scope.

Asked ChatGPT its opinion and it provided the code below, which is a bigger refactorm looks kinda nice but I haven't tested it

import { useEffect, useRef } from 'react';

const useFocusManagement = (isOpen, shadowRoot) => {
  const previouslyFocusedElementRef = useRef(null);

  useEffect(() => {
    const getActiveElement = () => (shadowRoot ? shadowRoot.activeElement : document.activeElement);

    if (isOpen) {
      previouslyFocusedElementRef.current = getActiveElement();
    } else if (previouslyFocusedElementRef.current) {
      previouslyFocusedElementRef.current.focus();
    }
  }, [isOpen, shadowRoot]);
};

export default useFocusManagement;

apps/web/src/components/user-tools/revision/checkout-reject-buttons.tsx
CodingDive: `setIsOpen` needs to be required for the state management to work. It's just that some consumers exclusively need the `onClose` handler. 
I find the above code more readable than saying

`setIsOpen={(open) => (!open && onCloseClick())}` but it'd save us the ternary :thinking:

packages/editor/src/core/helpers/use-shadow-root.ts
CodingDive: I like the way I wrote it better. Some more lines, but more clarity and easier types

elbotho (approved):
I'm a bit sad that we have to keep track of the previous focus ourselves, but it seems to work alright. thank you.

shn-srl (approved):

CodingDive:
> I'm a bit sad that we have to keep track of the previous focus ourselves, but it seems to work alright. thank you.

Same. This works from the library with ESC already, but for the outside click, we have to manage ourselves :crying_cat_face:

PR #3847, opened 2024-05-29
feat: Make image plugin available to integrations without upload (HACK)

LarsTheGlidingSquirrel:
Integrations can now enable the image plugin like this: 
```typescript
// using editor package
<SerloEditor _enableImagePlugin={true} ...>
  ...
</SerloEditor>

// using editor-web-component package
// -> enabled per default for now
```

Not to be used in production environments!

### DIFF
--- a/packages/editor-web-component/src/editor-web-component.tsx
+++ b/packages/editor-web-component/src/editor-web-component.tsx
@@ -22,7 +22,6 @@ export class EditorWebComponent extends HTMLElement {
   private container: HTMLDivElement
 
   private _mode: Mode = 'read'
-  private _imagePluginEnabled = 'false'
 
   private _initialState: InitialState = exampleInitialState
   private _currentState: unknown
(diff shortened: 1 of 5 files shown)


packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -151,6 +162,9 @@ export class EditorWebComponent extends HTMLElement {
             <Suspense fallback={<div>Loading editor...</div>}>
               <LazySerloEditor
                 initialState={this.initialState}
+                _enableImagePlugin={
```
CodingDive: Just always set to true?

packages/editor/src/package/editor.tsx
```diff
@@ -40,7 +44,32 @@ export function SerloEditor(props: SerloEditorProps) {
   const { instanceData, loggedInData } = editorData[language]
 
   const basicPlugins = createBasicPlugins(pluginsConfig)
-  editorPlugins.init([...basicPlugins, ...customPlugins])
+  let allPlugins = [...basicPlugins, ...customPlugins]
+  // HACK: Temporary solution to make image plugin available in Moodle & Chancenwerk integration with file upload disabled.
+  if (props._enableImagePlugin) {
```
CodingDive, line 49: I think we could even make it always true on this level. What do you think?

CodingDive (approved):
Looks good to me! Let's release a new editor version and then I can publish it to the next editor-web-component release too?

PR #3852, opened 2024-05-31
refactor: `UuidContext` and `ExerciseContext`

LarsTheGlidingSquirrel:
Based on https://github.com/serlo/frontend/pull/3759 and https://github.com/serlo/frontend/pull/3754

Before: 
- `UuidContext` and exercise tracking IDs were mixed up in the code and were used in a way that was confusing to me.  

Changes: 
  - Exercises are enabled in articles, courses and on `___editor_preview` without having to enable experiment in user settings
  - `UuidContext` provides Serlo entity ID -> the uuid you can query in our database. If you are on a page rendering multiple entities with unique uuids (taxonomy page showing multiple exercises / exercise groups) the context provides different values depending on from where you access it. 
  - `ExerciseContext` is used to tell an exercises (A) if it has its own entity ID (B) if it is part of an exercise group (C) the tracking ID used for analytics. 

Followups:
  - Remove `exerciseVisibleInSuggestion` from editor API (no need to customize that any more)

Demo: 
- Exercise folder - 1 is exercise, 2 is exercise group - http://localhost:3000/mathe/97589/aufgaben-zur-schriftlichen-addition-natuerlicher-zahlen
- Exercise group - http://localhost:3000/9743
- Exercise - http://localhost:3000/97600
- Article - http://localhost:3000/277232


apps/web/src/components/entity/entity.tsx
```diff
@@ -144,7 +146,23 @@ export function Entity({ data }: EntityProps) {
...
+      !Array.isArray(document) &&
+      document.plugin === EditorPluginType.Exercise
+    const content = isExercise ? (
+      <ExerciseContext.Provider
+        value={{
+          hasEntityId: true,
+          isInExerciseGroup: false,
+        }}
+      >
+        {' '}
```
elbotho:
```suggestion
```

apps/web/src/contexts/uuids-context.ts
```diff
@@ -1,5 +1,10 @@
 import { createContext, useContext } from 'react'
 
+// Provides ID & revision ID of the Serlo entity in our database. Examples:
+// - On an article page like '.../mathe/75512/rechentricks-für-die-addition' this will return ID:'75512' & revision:'undefined'.
+// - On a page rendering multiple exercises (each having a Serlo entity ID) this will return the corresponding information depending on where you are
+// - On a page not linked to any Serlo entity like '.../math' this will return undefined
+
```
elbotho: maybe format as jsdoc?

apps/web/src/components/entity/entity.tsx
```diff
@@ -144,7 +146,23 @@ export function Entity({ data }: EntityProps) {
...
-    const content = <StaticRenderer document={document} />
+    const isExercise =
+      document &&
+      !Array.isArray(document) &&
+      document.plugin === EditorPluginType.Exercise
+    const content = isExercise ? (
+      <ExerciseContext.Provider
+        value={{
+          hasEntityId: true,
+          isInExerciseGroup: false,
```
elbotho: Since GroupedExercises don't are no longer entities…
I'm wondering if `hasEntityId` would not be clearer as `isStandanloneExercise` or something like that, but that probably is better or worse depending on where it would be used.

elbotho (commented):
Thank you for working on this.

Some small comments/suggestions for now.
I will do some testing with the comment are part later.

apps/web/src/components/entity/entity.tsx
elbotho: or `isEntity`?

LarsTheGlidingSquirrel: I also like `isEntity`. I changed it.

apps/web/src/components/entity/entity.tsx
LarsTheGlidingSquirrel: I guess prettier sneaked that in somehow. :smile:

apps/web/src/contexts/uuids-context.ts
LarsTheGlidingSquirrel: Sure. Changed it.

elbotho (approved):

PR #3853, opened 2024-06-02
refactor(sso): make env check before enabling SSO button

hugotiburtino:
It will simplify [a code in backend](https://github.com/serlo/infra/pull/76/files) and also make a more granular enablement of the OIDC providers (v.g. we may enable VIDIS but keep MeinBildungsraum disabled).


elbotho (approved):
👍

PR #3855, opened 2024-06-05
feat(auth): vidis as experiment with special auto-activate route

elbotho:
- this will allow testing of vidis in production
- ~might conflict with https://github.com/serlo/frontend/pull/3853 (not sure if it's necessary any more)~

route to directly enable vidis experiment is `/auth/___activate_vidis` (but not testable in feature branch since vidis is always active in dev/staging)

### DIFF
--- a/apps/web/src/components/user/profile-experimental.tsx
+++ b/apps/web/src/components/user/profile-experimental.tsx
@@ -23,6 +23,12 @@ export const features = {
     activeInDev: true,
     hideInProduction: true,
   },
+  authVidis: {
+    cookieName: 'useAllowVidisLogin',
+    isActive: false,
+    activeInDev: false,
+    hideInProduction: true,
+  },
 }
 
 const showExperimentsStorageKey = 'showExperiments'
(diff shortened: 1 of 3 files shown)


hugotiburtino:
> might conflict with https://github.com/serlo/frontend/pull/3853 (not sure if it's necessary any more)

It is still necessary, because we need (or better said: it is easier) to enable both providers. So, we still need to hide NPB.
I resolved the conflict.

hugotiburtino (approved):

apps/web/src/components/user/profile-experimental.tsx
```diff
@@ -119,6 +125,16 @@ export function ProfileExperimental() {
           </p>
         </div>
       ) : null}
+      {shouldBeVisible('authVidis') ? (
+        <div>
+          <h3 className="serlo-h3 mb-3">
+            {renderFeatureButton('authVidis')} ✹
```
hugotiburtino, line 131: > ✹

strange things may appear 😄

PR #3868, opened 2024-06-07
feat(plugin-image): expose `upload` prop in Editor config

### LINKED ISSUE #3866: editor: Expose upload of ImageConfig in Editor API
## Problem which needs to be solved

In the Moodle plugin we want to use the `image` plugin with a specific configuration so that the files are uploaded to Moodle.

## Solution

1. Add `ImageConfig` to `createBasicPlugins()` + `PluginsConfig` -> When the upload-function is defined add the image plugin as well.
2. Move `validate` function into the editor
3. Make the `validate` function the default function which does not need to be specified.
4. Document the solution in the README with a note that the API might change in the future.

hejtful:
Resolves #3866 

Usage:
```
<SerloEditor
  pluginsConfig={{
    image: {
      upload: uploadFile
    }
  }}
  initialState={this.initialState}
>
```
where `uploadFile` is of type `UploadHandler`:
```
type UploadHandler = (file: File) => Promise<string>
```

### DIFF
--- a/packages/editor/src/editor-integration/create-basic-plugins.tsx
+++ b/packages/editor/src/editor-integration/create-basic-plugins.tsx
@@ -43,7 +43,7 @@ export function createBasicPlugins(config: CreateBasicPluginsConfig) {
     },
     {
       type: EditorPluginType.Multimedia,
-      plugin: createMultimediaPlugin(config.multimedia),
+      plugin: createMultimediaPlugin(createMultimediaPluginConfig(config)),
       visibleInSuggestions: true,
       icon: <IconMultimedia />,
     },
@@ -147,4 +147,23 @@ function hasImagePluginConfigUploadPropertyDefined(
   imagePluginConfig: CreateImagePluginConfig
 ): imagePluginConfig is ImageConfig {
   return imagePluginConfig.upload !== undefined
+}
+
+function createMultimediaPluginConfig(config: CreateBasicPluginsConfig) {
+  const { image, multimedia } = config
+
+  return {
+    ...multimedia,
+    allowedPlugins: multimedia.allowedPlugins.filter((allowedPlugin) => {
+      if (
+        // If the user didn't provide the `upload` prop for the Image plugin,
+        // remove the Image plugin from Multimedia config allowed plugins
+        allowedPlugin === EditorPluginType.Image &&
+        !hasImagePluginConfigUploadPropertyDefined(image)
+      ) {
+        return false
+      }
+      return true
+    }),
+  }
 }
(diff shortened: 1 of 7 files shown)


packages/editor/src/editor-integration/create-basic-plugins.tsx
```diff
@@ -77,15 +83,27 @@ export function createBasicPlugins(props: Required<PluginsConfig>) {
...
+    ...(hasImagePluginConfigUploadPropertyDefined(config.image)
+      ? [
+          {
+            type: EditorPluginType.Image,
+            plugin: createImagePlugin(config.image),
+            visibleInSuggestions: true,
+            icon: <IconImage />,
+          },
+        ]
+      : []),
```
elbotho, line 97: We would need to add this to multimedia plugin (and soon image drag and drop) as well, right?

hejtful: Good point, I'll look into multimedia plugin for integrations.

I don't think that anything changes on the web/frontend side however, as the Image plugin there is initialized in the same way as it used to be.

hejtful: Alright, the solution that comes to my mind is to remove the Image plugin from `allowedPlugins` in the Multimedia plugin config, if Image plugin `upload` config prop is not provided by the user of the Editor. What do you think about that?

elbotho: >Alright, the solution that comes to my mind is to remove the Image plugin from allowedPlugins in the Multimedia plugin config, if Image plugin upload config prop is not provided by the user of the Editor. What do you think about that?

Makes sense. but trying to integrate the editor without image support is also an edge case and I'd be also fine with an descriptive error message somewhere for example :)

hejtful: Added a condition for now, let's see how messy the edge cases get and consider error messages then. I wouldn't deal with error messages in this PR because, [here for example](https://github.com/serlo/frontend/blob/staging/packages/editor/src/plugins/image/index.ts#L96C50-L96C65), we still use the web/frontend code in the editor. That should be cleaned first, before any more error messages, I think.

Thank you for noticing the Multimedia edge case! For the Dropzone plugin (I prefer to use a one-work name until we get a real name from @GregorZupan), it's easier, we just disable it completely if Image upload isn't provided.

hejtful:
Closing in favor of the Serlo-hosted bucket approach (as a short-term solution).

PR #3871, opened 2024-06-07
refactor(share-modal): Modernize copy pasta, share dialog

CodingDive:
* Get rid of deprecated document.execCommand('copy') and document.queryCommandSupported('copy')

### DIFF
--- a/apps/web/src/components/user-tools/share/share-modal.tsx
+++ b/apps/web/src/components/user-tools/share/share-modal.tsx
@@ -10,7 +10,7 @@ import {
   faEnvelope,
 } from '@fortawesome/free-solid-svg-icons'
 import QRCode from 'qrcode.react'
-import { MouseEvent, useRef, useState, useEffect } from 'react'
+import { MouseEvent, useState, useEffect } from 'react'
 
 import { FaIcon, FaIconProps } from '../../fa-icon'
 import { ModalWithCloseButton } from '@/components/modal-with-close-button'
@@ -42,7 +42,6 @@ export function ShareModal({
   showPdf,
   path,
 }: ShareModalProps) {
-  const shareInputRef = useRef<HTMLInputElement>(null)
   const { strings, lang } = useInstanceData()
   const { entityId } = useEntityData()
   const pathOrId = path ?? entityId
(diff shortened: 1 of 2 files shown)


apps/web/src/components/user-tools/share/share-modal.tsx
```diff
@@ -46,18 +46,29 @@ export function ShareModal({
...
-    document.execCommand('copy')
-    target.focus()
-    showToastNotice(
-      '👌 ' + (text ? text : strings.share.copySuccess),
-      'success'
-    )
+  async function copyToClipboard(event: MouseEvent, text?: string) {
+    try {
+      if (shareInputRef.current) {
+        await navigator.clipboard.writeText(shareInputRef.current.value)
```
elbotho:
```suggestion
        await navigator.clipboard.writeText(shareUrl)
```

I think `shareInputRef` is generally no longer needed and can be removed.

apps/web/src/components/user-tools/share/share-modal.tsx
```diff
@@ -46,18 +46,29 @@ export function ShareModal({
...
-  function copyToClipboard(event: MouseEvent, text?: string) {
-    const target = event.target as HTMLAnchorElement
-    shareInputRef.current!.select()
-    document.execCommand('copy')
-    target.focus()
-    showToastNotice(
-      '👌 ' + (text ? text : strings.share.copySuccess),
-      'success'
-    )
+  async function copyToClipboard(event: MouseEvent, text?: string) {
```
elbotho: event is unused and and be removed

apps/web/src/components/user-tools/share/share-modal.tsx
```diff
@@ -99,7 +110,7 @@ export function ShareModal({
         copyToClipboard(
           event,
           'Link in die Zwischenablage kopiert.\r\nEinfach auf Mebis (www.mebis.bayern.de) einfügen!'
-        )
+        ).catch(() => void null)
```
elbotho:
```ts 
void copyToClipboard(
    event,
    'Link in die Zwischenablage kopiert.\r\nEinfach auf Mebis (www.mebis.bayern.de) einfügen!'
)
```
is what we'd usually do here, the catch does not really make sense

elbotho (changes requested):
Thanks for the 🍝 äh, modernisation. 
Suggested some more cleanup.

elbotho (approved):
thank you

PR #3879, opened 2024-06-11
Refactor(editor) replace NEXT env variables with custom rollup plugin

CodingDive:
Consumers of the React package (@serlo/editor) had to define the process.env.NEXT_PUBLIC_ENV. Now we already remove it in the built of the editor, instead of the built of the editor-web-component which makes the life of consumers of the package slightly easier.

Works great for me. @LarsTheGlidingSquirrel I did check edusharing with yalc, but better double check that we are not breaking anything there as it uses Next.js under the hood. To be clear, this strips away the NEXT env variables at build time, so doesn't influence or overwrite the ones we are defining as part of edusharing.


kulla (approved):

CodingDive:
Merging now @LarsTheGlidingSquirrel if there should be an issue, we still have time to fix it before upgrading EduSharing :blush:

PR #3884, opened 2024-06-12
refactor(taxonomy): Remove `taxonomyId`

kulla:
After https://github.com/serlo/db-migrations/issues/346 there is no taxonomy table / taxonomy ID for a taxonomy term => Thus I would delete the quest of it in the query so that I can delete this field in the API in https://github.com/serlo/api.serlo.org/pull/1561


elbotho:
user journey is now [merged into staging](https://github.com/serlo/frontend/pull/3730/files) and we should not have to keep this branch up to date

PR #3894, opened 2024-06-18
fix(dropzone-image): relative position and layout

hejtful:
Done:
- Smaller empty canvas size
- Use relative values (percentage) instead of absolute values for positioning and size of answer zones
- Replace the "scaler" responsive approach with percentage-based one
- Snap the answer zone to the edges of canvas, when dropping them outside while positioning them
- Enable all four resize handles for answer zones
- Limit resizing within canvas limits (snap on overflow)
- General code cleanup and bugfixes
- Feedback from @elbotho 

### [demo in article](https://frontend-git-fix-dropzone-image-sizing-serlo.vercel.app/277232)


packages/editor/src/plugins/dropzone-image/components/answer-zone/answer-content.tsx
```diff
@@ -25,31 +25,15 @@ export function AnswerContent(props: AnswerContentProps) {
...
         data-qa="plugin-dropzone-image-answer-content-text"
-        className={classNames}
+        className={cn(
+          'mx-0 block rounded-full border border-brand px-2',
+          className
+        )}
       >
-        {content}
+        <StaticSlate element={text} pClassName={pClassName} />
       </span>
```
elbotho: this would result in a `p` inside a `span` right?

elbotho:
unrelated, but if the popup gets to long it is not usable any more
image

elbotho:
Unrelated UX/UI stuff: 
button has no hover effect and text of button does not make sense if I edit an existing zone.
image
Also makes me unsure what happens if I close the popup (Actually everything is changes already and I never **have** to press the button)

elbotho:
unrelated ux:
settings and plus buttons get really small.
image
Also: They don't have tooltips.

packages/editor/src/plugins/dropzone-image/components/answer-zone/answer-content.tsx
elbotho: I'd prefer some css magic (like `[&>p]:…`) instead of passing this all the way, but that probably personal preference tbh.

apps/web/src/data/en/index.ts
```diff
@@ -710,7 +710,7 @@ export const loggedInData = {
           },
           answerZone: {
             description: 'Description (optional)',
-            sizeLabel: 'Configure the size of the drop zone',
+            sizeLabel: 'Configure the size of the drop zone (in percentages)',
```
elbotho: I think adding a `%` after the inputs below this text might work nicely instead.

elbotho (commented):
some feedback, more maybe after our call

packages/editor/src/plugins/dropzone-image/components/answer-zone/answer-content.tsx
hejtful: 1. Changed to a `div`, good point.
2. I like the css way more as well, just didn't think of it. 👍

shn-srl (approved):

PR #3904, opened 2024-06-24
refactor(event): delete RemoveEntityLink

hugotiburtino:
It's causing a bug in the notifications and the event will anyway stop to exist after merging
the course pages. So it was already removed from api, see https://github.com/serlo/api.serlo.org/pull/1607

By the way, it could go directly to `production`, but I'll leave the decision for the frontend people.


elbotho:
thank you @hugotiburtino I will deploy a minimal version of this to prod now

PR #3923, opened 2024-07-02
feat(image-plugin-redesign): Add licence selection to image plugin settings

shn-srl:
closes https://github.com/serlo/produktentwicklung/issues/14
Issue Link: https://github.com/orgs/serlo/projects/45/views/8?pane=issue&itemId=65345579


packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+}) => {
+  const editorStrings = useEditorStrings()
+  const imageStrings = editorStrings.plugins.image
+  const { licences, licenceHelpText } = imageStrings
+  const licenceNames = Object.values(licences)
+
+  const [selectedLicense, setSelectedLicense] = useState(licenceNames[0])
+  const [isConfirmed, setIsConfirmed] = useState(false)
+  const [isOpen, setIsOpen] = useState(false)
+  const [triggerWidth, setTriggerWidth] = useState(0)
```
shn-srl: @hejtful @elbotho I didn't find a better way to get the content of the dropdown and the select box itself to keep the same width. If you guys are unhappy with this implementation I would be grateful for a short pairing session to check it out together

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+    setIsConfirmed(true)
+  }
+
+  return (
+    <label
+      ref={triggerRef}
+      className="mx-auto mb-0 mt-5 flex items-center text-almost-black"
+    >
+      <span className="serlo-tooltip-trigger w-1/4">
+        Lizenz
```
elbotho: i18n

elbotho:
A [Demo Link](https://frontend-git-240702-select-image-licence-serlo.vercel.app/___editor_preview?state=%7B%22plugin%22%3A%22rows%22%2C%22state%22%3A%5B%7B%22plugin%22%3A%22image%22%2C%22state%22%3A%7B%22src%22%3A%22https%3A%2F%2Fassets.serlo.org%2Flegacy%2F8493_2pjCUI3IZd.png%22%2C%22caption%22%3A%7B%22plugin%22%3A%22text%22%2C%22state%22%3A%5B%7B%22type%22%3A%22p%22%2C%22children%22%3A%5B%7B%22text%22%3A%22%22%7D%5D%7D%5D%2C%22id%22%3A%2248a0ab11-0aab-4d57-b10e-0bf3edc46dce%22%7D%7D%2C%22id%22%3A%228679f418-1b18-4b9a-9214-0bc9ac86b05f%22%7D%5D%7D) helps me a lot when reviewing or testing ✌️

packages/editor/src/plugins/image/components/licence-dropdown.tsx
elbotho: I would not mind if box and dropdown don't have the same width and would probably not add the complexity for a select box in a settings modal

packages/editor/src/plugins/image/controls/settings-modal-controls.tsx
```diff
@@ -27,7 +34,15 @@ export function SettingsModalControls({ state }: Pick<ImageProps, 'state'>) {
...
-        onChange={(e) => src.set(e.target.value)}
+        onChange={(e) => {
+          src.set(e.target.value)
+        }}
+      />
+      <LicenseDropdown
+        defaultLicense={licence?.defined ? licence.value : undefined}
+        onLicenseChange={updateOrCreateLicence}
+        // eslint-disable-next-line @typescript-eslint/no-base-to-string
+        isPixabayImage={src.get().toString().includes('Pixabay Lizenz')}
```
elbotho: When would the src ever include `'Pixabay Lizenz'`?

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+                <Select.ItemText>{license}</Select.ItemText>
+              </Select.Item>
+            ))}
+          </Select.Content>
+        </Select.Root>
+        {isConfirmed && (
+          <span className="pointer-events-none ml-2 inline flex items-center pr-3 text-green-500">
+            &#10003;
+          </span>
+        )}
```
elbotho: I don't understand this tick tbh. Maybe more an UX Problem but if I want it to be CC BY-SA 4.0 I will never get a tick because I did not change it, correct?

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+      ref={triggerRef}
+      className="mx-auto mb-0 mt-5 flex items-center text-almost-black"
+    >
+      <span className="serlo-tooltip-trigger w-1/4">
+        Lizenz
+        <FaIcon className="ml-2" icon={faQuestionCircle} />
+        <EditorTooltip text={licenceHelpText} />
+      </span>
+      <span className="flex w-3/4 flex-row border-2 border-transparent">
+        <Select.Root
```
elbotho: I have some problems with this component. Not sure if it's the comp or the implementation.
- keyboard: I can open the options, but not change them
- opening the options changes the height of the select, causes the layout to shift

shn-srl:
> A [Demo Link](https://frontend-git-240702-select-image-licence-serlo.vercel.app/___editor_preview?state=%7B%22plugin%22%3A%22rows%22%2C%22state%22%3A%5B%7B%22plugin%22%3A%22image%22%2C%22state%22%3A%7B%22src%22%3A%22https%3A%2F%2Fassets.serlo.org%2Flegacy%2F8493_2pjCUI3IZd.png%22%2C%22caption%22%3A%7B%22plugin%22%3A%22text%22%2C%22state%22%3A%5B%7B%22type%22%3A%22p%22%2C%22children%22%3A%5B%7B%22text%22%3A%22%22%7D%5D%7D%5D%2C%22id%22%3A%2248a0ab11-0aab-4d57-b10e-0bf3edc46dce%22%7D%7D%2C%22id%22%3A%228679f418-1b18-4b9a-9214-0bc9ac86b05f%22%7D%5D%7D) helps me a lot when reviewing or testing ✌️

Thanks, noted for next time :+1: :pray:

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
+import { cn } from '@/helper/cn'
+
+interface LicenseDropdownProps {
+  defaultLicense?: string
+  onLicenseChange?: (license: string) => void
+  isPixabayImage?: boolean
+}
+
+export const LicenseDropdown: React.FC<LicenseDropdownProps> = ({
```
elbotho: only usage of React.FC in this codebase.

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+import React, { useState, useEffect, useRef } from 'react'
+
+import { FaIcon } from '@/components/fa-icon'
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
+import { cn } from '@/helper/cn'
+
+interface LicenseDropdownProps {
+  defaultLicense?: string
+  onLicenseChange?: (license: string) => void
+  isPixabayImage?: boolean
```
elbotho: would it maybe make more sense to pass down the `src` here or some other field? 
So in the future we can also have special behaviour for other providers as well.

shn-srl:
Note: e2e tests failing for dropzone image plugin and video + injection, investigating

image

packages/editor/src/plugins/image/controls/settings-modal-controls.tsx
```diff
@@ -4,14 +4,21 @@ import { useEditorStrings } from '@serlo/frontend/src/contexts/logged-in-data-co
...
   const imageStrings = useEditorStrings().plugins.image
 
   const isTemp = isTempFile(src.value)
   const isFailed = isTempFile(src.value) && src.value.failed
 
+  const updateOrCreateLicence = (newLicence: string) => {
+    if (!licence?.defined) licence.create(newLicence)
+    else {
+      licence.set(newLicence)
+    }
```
elbotho: maybe?
```suggestion
    if (!licence?.defined) licence.create(newLicence)
    else licence.set(newLicence)
```

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { faCaretDown, faCaretUp } from '@fortawesome/free-solid-svg-icons'
+import * as Select from '@radix-ui/react-select'
+import React, { useState, useEffect, useRef } from 'react'
+
+import { FaIcon } from '@/components/fa-icon'
+import { useEditorStrings } from '@/contexts/logged-in-data-context'
+import { cn } from '@/helper/cn'
+
+interface LicenseDropdownProps {
+  defaultLicense?: string
```
elbotho: this seems to be unused?

apps/web/src/data/en/index.ts
```diff
@@ -795,6 +795,17 @@ export const loggedInData = {
...
+            'External content with the following licenses may be integrated on serlo.org:',
+          licences: {
+            CC4: 'CC BY-SA 4.0',
+            CC3: 'CC BY-SA 3.0',
+            CC2: 'CC BY-SA 2.0',
+            CCBY: 'CC BY',
+            CCO: 'CCO',
+            publicDomain: 'Public Domain',
+            pixabay: 'Pixabay License',
+          },
```
elbotho: Will there actually be translations? Or could we also use the same strings for all language versions here?

packages/editor/src/plugins/image/index.ts
```diff
@@ -23,6 +23,7 @@ const imageState = object({
   src: upload(''),
   link: optional(object({ href: string('') })),
   alt: optional(string('')),
+  licence: optional(string('')),
```
elbotho, line 27: I would probably prefer a id here otherwise we would need a migration to change a license title, right?

elbotho (changes requested):
Hi there, there are some things I'm not sure about in this one (see comments).
It's on purpose, that the license is not displayed anywhere yet, right?

packages/editor/src/plugins/image/index.ts
elbotho: Without thinking about it too much I think I'd go with an enum with fixed title strings that are the same for all languages for now.

elbotho:
Bit unrelated, but settings modal is quite low now and could easily be clipped on smaller screens:
image

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Done

packages/editor/src/plugins/image/controls/settings-modal-controls.tsx
shn-srl: Done :heavy_check_mark:

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Removed :+1:

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Done

packages/editor/src/plugins/image/controls/settings-modal-controls.tsx
shn-srl: Typo, thanks, fixed

apps/web/src/data/en/index.ts
shn-srl: I think probably only for the pixabay licence and public domain

packages/editor/src/plugins/image/index.ts
shn-srl: Its now saving the key which is usable as ID :+1:

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Correct - I just pushed the correct behaviour for the default value.

From the issue:

> Per Default wird die Standart-Lizenz CC BY-SA 4.0 ausgewählt
> 
> Sobald eine Änderung an den Lizenzen vorgenommen wurde: Grünes Häkchen setzen
> 
> für spätere Iteration (bei Lust jetzt schon einbauen): Falls keine Änderung an der Lizenz vorgenommen wird noch einmal via Popup beim User nachfragen, ob das Bild unter CC BY-SA 4.0 steht. Dann automatisch grünes Häkchen setzen

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Removed

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: I tend to agree but since the focus of the ticket is about redesign so it seems that design and UX is important and I wouldn't feel comfortable having it not match the design exactly without having input from Gregor

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Fixed :heavy_check_mark:

apps/web/src/data/en/index.ts
```diff
@@ -795,6 +795,18 @@ export const loggedInData = {
...
           change: 'Change image',
+          licence: 'Licence',
+          licenceHelpText:
+            'External content with the following licenses may be integrated on serlo.org:',
+          licences: {
+            CC4: 'CC BY-SA 4.0',
+            CC3: 'CC BY-SA 3.0',
+            CC2: 'CC BY-SA 2.0',
+            CCBY: 'CC BY',
+            CCO: 'CCO',
```
hejtful: I don't know much about these licences, so it might be a super silly question 😄 But why do they need to be i18n strings?

e2e-tests/tests/460-dropzone-image-plugin.ts
```diff
@@ -32,12 +32,12 @@ Scenario('Create a drag drop exercise with two dropzones', async ({ I }) => {
   I.type(
     'https://assets.serlo.org/5ab7c782ad7f7_f4111037d697776c337e7bffd142cedd01324bc9.png'
   )
-  I.click('$modal-close-button')
   I.seeNumberOfElements('$answer-zone-answerZone-0', 1)
 
   I.seeNumberOfElements('$plugin-dropzone-image-answer-content-image', 2)
 
   I.say('Click settings button of first answer zone')
+  I.click('$modal-close-button')
```
hejtful, line 40: Please pay attention to the `I.say` statements. They are there to group lines into specific actions. Does clicking of the close modal belong to this group?

When you add new e2e tests, you don't have to use `I.say` if you don't want to, of course. But if you change a test with a certain structure, please respect it 🙏 

That being said, could you please explain why this change was needed?

packages/editor/src/plugins/image/components/licence-dropdown.tsx
hejtful: >I wouldn't feel comfortable having it not match the design exactly without having input from Gregor

The question mark icon isn't right-aligned, and it's right aligned in the design, so it's not matching the design exactly anyway.

Plus, on Safari, there's this tiny hole in the border.
Screenshot 2024-07-03 at 16 57 45

My point is, no need to go for pixel perfect, if something increases complexity in the code and doesn't bring a lot of value in the UX, it's good to avoid the complexity. If you do believe that this brings value to UX however, then there should be an editor-wide dropdown component detached from the licencing logic, where this UI fix would belong.

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,118 @@
+import { EditorTooltip } from '@editor/editor-ui/editor-tooltip'
+import { faQuestionCircle } from '@fortawesome/free-regular-svg-icons'
+import { faCaretDown, faCaretUp } from '@fortawesome/free-solid-svg-icons'
+import * as Select from '@radix-ui/react-select'
+import React, { useState, useEffect, useRef } from 'react'
```
hejtful: No need to import React

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <Select.Trigger
+            className={cn(
+              'serlo-input-font-reset rounded-md border-2 border-editor-primary-100 bg-editor-primary-100 px-2 text-left focus:border-2 focus:border-editor-primary',
+              'flex flex-row justify-between',
+              isOpen &&
+                'rounded-b-none border-editor-primary-300 border-b-transparent'
+            )}
+            style={{ width: `${triggerWidth}px` }}
+          >
+            <Select.Value placeholder="Select an option" />
```
hejtful: Not sure if this placeholder prop will ever be visible. If it will, could you use an i18n string? If it won't, maybe remove it?

apps/web/src/data/en/index.ts
shn-srl: I don't know if the lettering of the licences depends on language.
For sure we need to translate the last two keys, pixabay and public domain. I guess I could move the rest to the code.

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Removed

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: Removed

e2e-tests/tests/460-dropzone-image-plugin.ts
shn-srl: This change was needed because the test was failing at this step, and when I investigated I saw that it was stuck with an open modal, and needed this line to pass the test

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: I will try to make a seperate dropdown component as you suggest

packages/editor/src/plugins/image/components/licence-dropdown.tsx
elbotho: for @GregorZupan: As a user testing this feature I was mostly unsure what that tick meant and was slightly confused.

apps/web/src/data/en/index.ts
elbotho: > I don't know if the lettering of the licences depends on language.

No, it does not.

apps/web/src/data/de/license-data-short.ts
```diff
@@ -1,15 +1,67 @@
 import { LicenseData } from '@/data-types'
 
+const ccBaseLicence = {
+  title: 'Dieses Werk steht unter der freien Lizenz CC BY-SA __VERSION__.0',
+  shortTitle: 'CC BY-SA __VERSION__.0',
+  url: 'https://creativecommons.org/licenses/by-sa/__VERSION__.0/deed.de',
+  agreement:
+    'Mit dem Speichern dieser Seite versicherst du, dass du deinen Beitrag (damit sind auch Änderungen gemeint) selbst verfasst hast bzw. dass er keine fremden Rechte verletzt. Du willigst ein, deinen Beitrag unter der <a href="https://creativecommons.org/licenses/by-sa/__VERSION__.0/deed.de">Creative Commons Attribution/Share-Alike Lizenz __VERSION__.0</a> und/oder unter einer gleichwertigen Lizenz zu veröffentlichen, welche der Serlo Education e. V. entsprechend der Regelungen in den <a href="/terms">Nutzungsbedingungen</a> festlegen darf. Falls du den Beitrag nicht selbst verfasst hast, muss er unter den <a href="/terms">Nutzungsbedingungen</a> verfügbar sein und du stimmst zu, notwendigen Lizenzanforderungen zu folgen.',
```
elbotho: Will those infos be needed for the image-licenses as well at some point or did you just add it to make it fit the structure?

shn-srl: Added it to make it fit the structure

packages/editor/src/plugins/image/components/licence-dropdown.tsx
shn-srl: @hejtful I improved the component and made it an shared editor-wide dropdown component, it seems to work well now.

apps/web/src/data/en/index.ts
shn-srl: Moved to licence data

apps/web/src/data/de/license-data-short.ts
elbotho: what's the upside to integrating it here?

shn-srl: They fit well here, since this is where the other "licence data" is stored. It seems like a better option than hardcoding them in the component. 

If you have a suggestion for a better place I'm open for it :pray:

elbotho: My thoughts:
- since we don't use the same licenses for images and content I think they should have their own data structure/variable
- content licenses might move to the editor as well but until they do they are part of the frontend code, the image licenses are immediately part of the editor and might be used in the integrations as well so the code should already be in the editor imho

elbotho (changes requested):
I would prefer the license info in another place as I explained in my comment.

Also the keyboard experience is still completely broken for me in the select component.
Key events get passed down so using arrow keys focused the next plugin and somehow breaks the whole editor. Enter and Escape also close the popup instead of interacting with the select and also break the editor. Not sure why that is happening.
(current chromium)
Honestly I prefer the less pretty native dropdown that actually works instead of this component if it's not reliable and accessible.

apps/web/src/data/de/license-data-short.ts
shn-srl: Moved the licences to the code.

shn-srl:
> I would prefer the license info in another place as I explained in my comment.
> 
> Also the keyboard experience is still completely broken for me in the select component. Key events get passed down so using arrow keys focused the next plugin and somehow breaks the whole editor. Enter and Escape also close the popup instead of interacting with the select and also break the editor. Not sure why that is happening. (current chromium) Honestly I prefer the less pretty native dropdown that actually works instead of this component if it's not reliable and accessible.

I've improved the behaviour related to keyboard. Let's maybe take a look together if its still at an unacceptable point for you?

packages/editor/src/plugins/image/components/licence-dropdown.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onLicenseChange,
+  src,
+  currentLicence,
+}) => {
+  const editorStrings = useEditorStrings()
+  const { licence, licenceHelpText } = editorStrings.plugins.image
+
+  const [isConfirmed, setIsConfirmed] = useState(
+    src ? src.includes('pixabay') : false
+  )
```
elbotho: code golf ;) 
```suggestion
  const [isConfirmed, setIsConfirmed] = useState(!!src?.includes('pixabay'))
```

shn-srl: Nice !!

packages/editor/src/core/components/nice-dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  <Select.ItemText>{label}</Select.ItemText>
+                </Select.Item>
+              ))}
+            </Select.Viewport>
+          </Select.Content>
+        </Select.Root>
+      </span>
+      {isConfirmed && (
+        <span className="pointer-events-none ml-2 inline-flex items-center pr-3 text-green-500">
+          &#10003;
```
elbotho: optional: I usually just use the char directly (✓) or add a comment for readability

elbotho: aside: in most contexts this confirmed logic is not needed so we should probably have a prop in the comp to activate it.

packages/editor/src/core/components/nice-dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }, [isOpen, hasOpened])
+
+  // Update selectedValue if defaultValue changes
+  useEffect(() => {
+    setSelectedValue(defaultValue)
+  }, [defaultValue])
+
+  return (
+    <label
+      className={`h-100 mx-auto mb-0 mt-5 flex items-center text-almost-black ${className}`}
```
elbotho: works but that's also a good place to use the `cn` helper

packages/editor/src/editor-ui/overlay-input.tsx
```diff
@@ -8,13 +8,15 @@ export interface OverlayInputProps
...
+  tooltip?: JSX.Element
 }
 
 export const OverlayInput = forwardRef<HTMLInputElement, OverlayInputProps>(
-  function OverlayInput({ label, autoFocus, ...props }, ref) {
+  function OverlayInput({ label, autoFocus, tooltip, ...props }, ref) {
     return (
       <label className="mx-auto mb-0 mt-5 flex items-center justify-between text-almost-black">
         <span className="w-1/5">{label}</span>
+        {tooltip && <span className="w-25] ml-[-25px]">{tooltip}</span>}
```
elbotho: typo at `w-25]`

packages/editor/src/plugins/image/index.ts
```diff
@@ -23,6 +23,8 @@ const imageState = object({
   src: upload(''),
   link: optional(object({ href: string('') })),
   alt: optional(string('')),
+  imageSource: optional(string('')),
+  licence: optional(string('')),
```
elbotho, line 27: Hmmm, as I said yesterday in out call, it feels weird to store a integer based id in a string here…

Please use strings as keys in your array or an enum or an object (`ImageLicense.Pixabay` would be a nice, readable way to call it) or if you want to keep the numeric ids please store them as a number as well…

elbotho (changes requested):

packages/editor/src/plugins/image/index.ts
shn-srl: Done, thanks :+1:

packages/editor/src/core/components/nice-dropdown.tsx
shn-srl: Added

packages/editor/src/core/components/nice-dropdown.tsx
shn-srl: Done

packages/editor/src/plugins/image/index.ts
elbotho: Nice!

hejtful (approved):

elbotho (approved):
finally :D

PR #3931, opened 2024-07-03
feat(image-plugin-redesign): Add Pixabay Search functionality

shn-srl:
Issue link: https://github.com/orgs/serlo/projects/45/views/8?pane=issue&itemId=65339969

[Demo Link](http://localhost:3000/___editor_preview?state=%7B%22plugin%22%3A%22rows%22%2C%22state%22%3A%5B%7B%22plugin%22%3A%22text%22%2C%22state%22%3A%5B%7B%22type%22%3A%22p%22%2C%22children%22%3A%5B%7B%22text%22%3A%22%22%7D%5D%7D%5D%2C%22id%22%3A%225cacb48f-1444-4d6f-9da6-cd33ea143020%22%7D%2C%7B%22plugin%22%3A%22image%22%2C%22state%22%3A%7B%22src%22%3A%22%22%2C%22caption%22%3A%7B%22plugin%22%3A%22text%22%2C%22state%22%3A%5B%7B%22type%22%3A%22p%22%2C%22children%22%3A%5B%7B%22text%22%3A%22%22%7D%5D%7D%5D%2C%22id%22%3A%2221a82234-38f4-4139-837b-2e79e281fe46%22%7D%7D%2C%22id%22%3A%2241a93f13-8754-411a-959b-48f6c4b30087%22%7D%5D%7D)

Todo: Add env variable for pixabay api key

PM if API key needed for testing


packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,158 @@
...
+      >
+        {/* Loading overlay */}
+        {(isLoadingImage || isSearching) && (
+          <div className="absolute inset-0 flex items-center justify-center bg-white bg-opacity-75 backdrop-blur-sm">
+            <h1 className="text-center text-xl font-bold">
+              {isLoadingImage
+                ? imageStrings.loadingImage
+                : isSearching
+                  ? imageStrings.searching
+                  : 'Hmm'}
```
elbotho: If "Hmm" is never reached, you could also remove the last conditional, right?

```suggestion
              {isLoadingImage
                ? imageStrings.loadingImage
                : imageStrings.searching }
```

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const imageStrings = editorStrings.plugins.image
+  const apiKey = 'secret'
+
+  useEffect(() => {
+    if (isLoadingImage) {
+      document.body.classList.add('overflow-hidden')
+    } else {
+      document.body.classList.remove('overflow-hidden')
+    }
+  }, [isLoadingImage])
```
elbotho: Not sure why this is needed, but maybe this should have a cleanup function to make sure on unmount the body is def. scrollable again.
``` ts
//...
return () => {
    document.body.classList.remove('overflow-hidden')
};
```

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const [query, setQuery] = useState('')
+  const [images, setImages] = useState<PixabayImage[]>([])
+  const [isLoadingImage, setIsLoadingImage] = useState(false)
+  const [isSearching, setIsSearching] = useState(false)
+
+  const [showTags, setShowTags] = useState(true)
+
+  const editorStrings = useEditorStrings()
+  const imageStrings = editorStrings.plugins.image
+  const apiKey = 'secret'
```
elbotho: this would be replaced by env?

apps/web/src/data/en/index.ts
```diff
@@ -795,6 +795,32 @@ export const loggedInData = {
...
+            science: 'Naturbeobachtungen',
+            class: 'Klassenzimmer',
+            books: 'Schulbücher',
+            hof: 'Schulhof',
+            teaching: 'Lehrmaterialien',
+            board: 'Schultafel',
+            groupWork: 'Gruppenarbeit',
+            projects: 'Schulprojekte',
+            games: 'Lernspiele',
+            creative: 'Kreatives Lernen',
```
elbotho: since those suggestions might change more often I again wonder if they should live in the i18n strings or if we should hardcode them. Not sure.

packages/editor/src/plugins/image/components/image-selection-screen.tsx
```diff
@@ -30,15 +32,32 @@ export function ImageSelectionScreen(
   const imgUrl = src.get() as string
   const showErrorMessage = imgUrl.length > 5 && !isImageUrl(imgUrl)
 
+  const [showPixabayModal, setShowPixabayModal] = useState(false)
+
+  const onSelectPixabayImage = (imageUrl: string) => {
```
elbotho: tiny thing, but  `imgUrl`, `isImageUrl` and `imageUrl` following each other so closely makes me look twice.

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const handleSearch = async (searchQuery: string) => {
+    if (!searchQuery) return
+    setIsSearching(true)
+    try {
+      const response = await fetch(
+        `https://pixabay.com/api/?key=${apiKey}&q=${encodeURIComponent(searchQuery)}&image_type=photo`
+      )
+      const data = (await response.json()) as PixabayResponse
+      setImages(data.hits)
+      setShowTags(false)
```
elbotho: currently I think showTags can be calculated by the value of `images` and does not need to be an own state, is that correct?

elbotho (changes requested):
Nice, not that much code for this feature!
Left some comments and did not test yet.
(I'll take that API key)

packages/editor/src/plugins/image/components/image-selection-screen.tsx
shn-srl: Changed imgUrl to imagUrl now for consistency, since  the API returns something with "largeImageUrl"

apps/web/src/data/en/index.ts
shn-srl: Will keep them here for now since they will need to be translated. Have taken a note to discuss with Gregor

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: Not needed, its a remnant, removed now, thanks !

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: Correct

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: Great, thanks, done

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: So far I've looked at it more as the "front search page" with the tags, and the "search results" page with the images, and if 0 images are returned then we are informed with text, but the tags are not present

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
elbotho: why not do it now? less chance of someone like me accidentally pushing the key when I put it into a local env file

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
elbotho: My point is that you should not put stuff in the state that can be calculated by existing state.
Maybe I still don't understand correctly but I still think that might be happening here.

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: Wasn't sure if the env is different for plugins or inside the editor package. Doing soon

elbotho: No idea how that works with integrations but if we use nextjs is should be available everywhere.

shn-srl: Done

elbotho:
I don't know how precious those API keys are but with the current approach we would expose them to the client and they can easily be stolen.
If we want to protect them we'd have to proxy the API calls to pixabay with a nextjs function (or cf-worker or api…) and make it not available to the integrations for now (fine by me).
This way we can also cache at least the suggested searches.

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,148 @@
...
+  return (
+    <div className="min-h-[60vw] pt-10">
+      <h2 className="mb-6 ml-10 mt-10 font-bold">{imageStrings.licenceFree}</h2>
+      {/* Search input */}
+      <div className="w-full">
+        <input
+          type="text"
+          value={query}
+          onChange={handleInputChange}
+          placeholder="Suche"
```
elbotho: i18n

shn-srl: Done

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: I've updated the code, please let me know what you think

elbotho:
Some things that came to mind while testing a bit:
also ping @GregorZupan for the ux stuff

- No option to get back to the suggestions (like a "x" button to remove the query, also it does not reset when I remove or change the query without pressing enter)
- I can click in the margin between two images (horizontally) and don't know which image will load
- As a user I need to know that I can send my query by pressing enter right now, button or hint would be nice for inexperienced users
- Images load in very slowly. First request almost timed out at least once
- no loading state / visual feedback while the larger version is loading after the popup is already closed. While waiting I thought it did not work
- "Searching for images" screen could use a spinner
- "Informatikkurs" has no results (and maybe other suggestions)

Text overflowing
image

"Bild Ändern" is a somewhat dangerous button :)
(People might click it on accident or because they want to see if they can edit the image or something like that and their image is gone)
image

Maybe a more helpful placeholder?
image

view while waiting for results is confusing. Image placeholders would help
image

elbotho (dismissed):
More comments after testing and also one issue with the secret.

shn-srl:
> More comments after testing and also one issue with the secret.

@elbotho 

I don't think its' much of a risk with the api key:

>  By default, you can make up to 100 requests per minute
https://pixabay.com/api/docs/

**Fixed**:
- Boundaries / Padding on images, images clickable but gaps not clickable
- Faster loading of preview images by loading in smaller sizes.
- Loading spinner added
- No text overflow

**Looking into fixing**:
- Lazy loading and skeletons for loading images (in image grid)
- _no loading state / visual feedback while the larger version is loading after the popup is already closed. While waiting I thought it did not work_ (not sure about solution for this)

**Need feedback from @GregorZupan :**

- _No option to get back to the suggestions (like a "x" button to remove the query, also it does not reset when I remove or change the query without pressing enter)_

- UX so its clear you have to hit "enter" - or search button beside search field?

-  _"Informatikkurs" has no results (and maybe other suggestions)_

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,149 @@
...
+    setQuery(tag)
+    await handleSearch(tag)
+  }
+
+  const handleKeyDown = async (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key !== 'Enter') return
+    await handleSearch(query)
+  }
+
+  const showTags = images.length === 0 && !isSearching
```
elbotho: nice

hejtful:
>I don't know how precious those API keys are but with the current approach we would expose them to the client and they can easily be stolen.
If we want to protect them we'd have to proxy the API calls to pixabay with a nextjs function (or cf-worker or api…) and make it not available to the integrations for now (fine by me).
This way we can also cache at least the suggested searches.

@elbotho @shn-srl It would become precious if we requested full API access for retrieving high resolution images. We may be able to get it as a non-profit, but then we might lose it if someone abuses our access. Good to keep in mind.

shn-srl:
Awaiting new search tags from @GregorZupan , and approval from @elbotho @hejtful (assuming API token is fine for now (I think it is))

elbotho:
Will this go to testing next or be public to everyone (including the editor integrations) directly?

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
```diff
@@ -0,0 +1,191 @@
...
+        </div>
+      )}
+
+      <div
+        className={cn(
+          'max-h-[500px]',
+          'mt-4 flex flex-wrap px-8',
+          isLoadingImage && 'max-h-100 border-1 border border-red-500'
+        )}
+        style={{ overflow: isLoadingImage ? 'hidden' : 'auto' }}
```
elbotho: why is this using style and is not part of tailwind classes above?

apps/web/src/data/en/index.ts
```diff
@@ -806,6 +807,32 @@ export const loggedInData = {
...
```
elbotho: If you want to keep them in the i18n strings the tags need to be english here.
Also then you need to test the english and the german versions as well. 
Personally I'd prefer to move them out and only offer the feature in the german versions anyway since that is less work overall.

elbotho (changes requested):

apps/web/src/data/en/index.ts
GregorZupan: Agreed with @elbotho. Let's enable it in the German version only
```
{
math: 'Mathematik',
nature: "Natur',
plants: 'Pflanzen',
adventure: 'Abenteuer',
teamwork: 'Teamwork',
journey: 'Reise",
sports: 'Sport',
chemistry: 'Chemie',
lab: "Labor',
tech: 'Technologie',
physics: 'Physik',
humans: 'Menschen',
pyramide: 'Pyramide',
cylinder: 'Zylinder',
art: 'Kunst',
music: 'Musik',
school: 'Schule'
}
```

GregorZupan:
@elbotho 
Due to time constraints I was not intending to do extensive testing, but after release I will invite our Authors to try it out and give some feedback.

packages/editor/src/plugins/image/components/pixabay-image-search.tsx
shn-srl: Fixed, thanks

apps/web/src/data/en/index.ts
shn-srl: Moved to code, enabled only for German

shn-srl:
Still awaiting approval from @hejtful @elbotho

apps/web/src/data/en/index.ts
elbotho: @GregorZupan is that the revised list? "Labor" does find the english Labor (arbeit) instead.
Menschen has only one hit and Physik has no hits for me.

elbotho (approved):

apps/web/src/data/en/index.ts
shn-srl: @elbotho I just pushed once more, added the lang param to the search query, it improves the results quite a bit

shn-srl: Menschen gets loads of hits, but unforunately Physik still gets none

GregorZupan: It's weird. Sometimes i get more hits sometimes less. I added more than enough words so if Physik doe not have any hit, we'll remove it

PR #3932, opened 2024-07-03
fix(editor): pass testingSecret through web-component

hejtful:
- Add editor-web-component testingSecret prop and pass it to the Editor as a config prop
- Fix image upload to temp bucket


packages/editor-web-component/src/editor-web-component.tsx
```diff
@@ -155,10 +166,21 @@ export class EditorWebComponent extends HTMLElement {
     this.reactRoot.render(
       <React.StrictMode>
         <div id="serlo-root" className="relative">
+          <span>{testingSecretAttr}</span>
```
CodingDive: Leftover?

CodingDive (approved):
Looks good to me other than the span rendering the secret

PR #3942, opened 2024-07-08
SEO: add noIndex to many pages, refactor head code

elbotho:
We index a lot of pages, that are not really relevant for almost all users (like author tools etc) and some of them are actually hurting our SEO ratings. So I added a `noindex` to most routes.

You can now also just add the `noIndex` prop to the `FrontendClientBase` component (`<FrontendClientBase noIndex>`).
This should make it easier and more visible that those pages are not indexed.

Reviewers could double check if I noindexed some important pages :)


Entkenntnis (approved):
I have looked through the pages and it looks fine

PR #3944, opened 2024-07-08
fix(image-plugin-redesign): fix url regex to be case insensitive

shn-srl:


elbotho (approved):
👍

elbotho:
gOoD jOB

hejtful (approved):
Bug fixed 👍 

Note: When you edit the URL, just do a backspace and the URL is invalid. This makes the modal disappear and the empty Image plugin state is shown. But, the URL field is not focused any more. This seems not really optimal from UX perspective.

shn-srl:
> Bug fixed 👍
> 
> Note: When you edit the URL, just do a backspace and the URL is invalid. This makes the modal disappear and the empty Image plugin state is shown. But, the URL field is not focused any more. This seems not really optimal from UX perspective.

Will discuss the UX of this with Gregor tomorrow

PR #3964, opened 2024-07-15
e2e(image-plugin): fix tests

shn-srl:
This comes from a change in the Image Plugin settings modal.

There are two seperate properties for an image in the image plugin:

- source: src property used in the 
- link: link to which the user should be sent if they click on the image.

The new [simplified](https://github.com/orgs/serlo/projects/45/views/8?pane=issue&itemId=65339991) Image Settings modal UX does not have a link field.

The previous behaviour of the image being clicked and having the link as href has been removed


hejtful:
This looks like a draft PR to me

shn-srl:
> This looks like a draft PR to me

Ready for review now, confirmed with Gregor

hejtful:
The tests are failing.

Also, please stop force pushing, it's desired to have the commit history.

elbotho (approved):

PR #3968, opened 2024-07-17
fix(exercise-generation): MenuButton crashes with empty string as value

CodingDive:
While recording a gif for Kulla and Vicky for some of our partners, I found this small bug of the MenuButton producing errors


hejtful (approved):
Nice, thanks for fixing so quickly

PR #3969, opened 2024-07-18
feat(dropzone-image): Enable DropzoneImage plugin in production

shn-srl:


elbotho (approved):
code is good. did no testing and I'm relying on you and Gregor that you did enough testing :)

PR #3986, opened 2024-07-23
Redesign Plugin suggestions

shn-srl:
[Issue Link](https://linear.app/serlo/issue/PE-34/ui-redesign-plugin-selection)

[Demo Link](https://frontend-git-240723-plugin-selection-menu-serlo.vercel.app/___editor_preview)

- [ ] Fix failing tests
- [ ] Split Fill Gaps plugin into two exercises
- [ ] Include Exercise Plugin container around inserted exercises


apps/web/src/components/modal-with-close-button.tsx
```diff
@@ -90,20 +91,27 @@ export function ModalWithCloseButton({
...
             data-modal-state={isOpen ? 'open' : 'closed'}
+            aria-describedby={undefined}
           >
             {title ? (
               <Dialog.Title className={cn('serlo-h2', extraTitleClassName)}>
                 {title}
               </Dialog.Title>
-            ) : null}
+            ) : (
+              <VisuallyHidden.Root asChild>
```
shn-srl: This was added as there was a console warning that Dialog.Title is necessary

packages/editor/src/plugin/helpers/get-plugin-title.tsx
```diff
@@ -3,8 +3,12 @@ import { useEditorStrings } from '@serlo/frontend/src/contexts/logged-in-data-co
...
   pluginType: string
-) {
+): string {
   return Object.hasOwn(pluginStrings, pluginType)
-    ? pluginStrings[pluginType as keyof typeof pluginStrings].title
+    ? (
+        pluginStrings[pluginType as keyof typeof pluginStrings] as {
+          title: string
+        }
+      ).title
```
shn-srl: This change was necessary to fix eslint errors in the file

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -20,10 +23,11 @@ interface useSuggestionsArgs {
   id: string
   focused: boolean
   isInlineChildEditor?: boolean
+  refocus?: () => void
```
hejtful: What's the reason for having this as an optional argument, instead of just defining the refocus function in this file?

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -34,30 +38,42 @@ const hotkeyConfig = {
   scopes: ['global'],
 }
 
-export const useSuggestions = (args: useSuggestionsArgs) => {
-  const { editor, id, focused, isInlineChildEditor } = args
-
+export const interactivePluginTypes = new Set([
```
hejtful: Why is this exported?

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -34,30 +38,42 @@ const hotkeyConfig = {
...
+  isInlineChildEditor,
+  refocus,
+}: useSuggestionsArgs) => {
   const dispatch = useAppDispatch()
-  const [selected, setSelected] = useState(0)
-  const suggestionsRef = useRef<HTMLDivElement>(null)
-  const pluginsStrings = useEditorStrings().plugins
-
+  const editorStrings = useEditorStrings()
+  const { plugins: pluginsStrings } = editorStrings
```
hejtful: Minor: This was one line - `const pluginsStrings = useEditorStrings().plugins`. What's the benefit of splitting it into two lines now?

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -34,30 +38,42 @@ const hotkeyConfig = {
...
-    const allVisible = editorPlugins
-      .getAllWithData()
+    const allWithData = editorPlugins.getAllWithData()
+    const allVisible = allWithData
       .filter(({ visibleInSuggestions }) => visibleInSuggestions)
       .map(({ type }) => type)
 
     const allowedByContext = allowedContextPlugins ?? allVisible
-
-    // Filter out plugins which can't be nested inside of the current plugin or ancestor plugins
```
hejtful, line 59: Why remove this comment? If it was misleading, updating it would have been great.

hejtful:
Inline placeholder with the button and its behaviour should be cleaned up now:
Screenshot 2024-08-06 at 12 58 20
Screenshot 2024-08-06 at 12 58 25

hejtful:
Moving onto the first suggestion on arrow down, when focus is on the search bar, would have been a nice addition.

\+ Maybe, moving onto the search bar on arrow up, when focus is on any of the suggestions in the top row.

hejtful:
Bug: When I'm focused on "Non interactive exercise" and press arrow down key, "Video" gets focused.

hejtful:
A Text plugin is added on plus button click. When the modal is closed, the Text plugin stays there. Could we adjust the code, so that nothing is added on plus button click, and a desired plugin is added only once selected in the modal?

apps/web/src/components/modal-with-close-button.tsx
hejtful: The console error:

>`DialogContent` requires a `DialogTitle` for the component to be accessible for screen reader users.

> If you want to hide the `DialogTitle`, you can wrap it with our VisuallyHidden component.

I don't think that we should just silence the error by adding an empty `VisuallyHidden`. If we're addressing the error, let's actually help the screen reader users by adding a helpful title for them?

apps/web/src/serlo-editor-integration/create-plugins.tsx
```diff
@@ -189,16 +193,34 @@ export function createPlugins({
...
       type: EditorPluginType.InputExercise,
+      visibleInSuggestions: true,
       plugin: createInputExercisePlugin(),
+      icon: <IconTextArea />,
+    },
+    {
+      type: EditorPluginType.ScMcExercise,
+      visibleInSuggestions: true,
+      plugin: createScMcExercisePlugin(),
+      icon: <IconScMcExercise />,
```
hejtful: Super minor: It would be super neat to keep the property order consistent across all the plugins. It makes it easier to scan the list and see what sticks out, the brain (mine, at least) is way faster if it can expect a certain prop in a certain position. Again, minor.

e2e-tests/tests/420-text-plugin.ts
```diff
@@ -4,32 +4,52 @@ Feature('Serlo Editor - Text plugin basic interactions')
...
+const selectors = {
+  textEditor: '$plugin-text-editor',
+  addNewPluginButton: '$add-new-plugin-row-button',
+  multimediaWrapper: '$plugin-multimedia-wrapper',
+  toolbarUndoButton: '$editor-toolbar-undo',
+  toolbarRedoButton: '$editor-toolbar-redo',
+  numberedListButton: '$plugin-toolbar-button-nummerierte-liste',
+}
+
+function addNewTextPlugin(I) {
```
hejtful, line 18: Nice idea with the helper functions. Maybe even consider creating a file for editor helpers and exporting all of them from there?

e2e-tests/tests/420-text-plugin.ts
```diff
@@ -43,252 +63,207 @@ Scenario('Add new line in plugin using Enter', async ({ I }) => {
...
 
   I.pressKey(['CommandOrControl', 'Z'])
-
   I.dontSee('Some text')
 
-  // ! For some reason, the first redo does not work. The second one does. If
-  // one puts a pause() here and runs the command only once through the
-  // interactive shell , it works just as fine as clicking the button.
-  // Therefore, I thought the Ctrl+Y was maybe happening too quickly after the
-  // Ctrl+Z, but even with I.wait(1) inbetween, two executions were needed.
```
hejtful: Please, please, don't remove comments like this. Whoever added it, had a good reason to do it. If nothing else, it points out to everyone that there is a reason for the double redo in the next two lines.

packages/editor/src/plugin/helpers/get-plugin-title.tsx
hejtful: The eslint error happened because of this change: https://github.com/serlo/frontend/pull/3986/files?short_path=c59cfd3#diff-e08687cc24fc586c93a42e9fa2ebb10f4a4f3e07819cf1e54b3b9cb8f3699b09R680-R681

Could you consider another place for those two strings? Like, instead of editor.plugins, create a new namespace editor.addPluginsModal? And then the a11y modal title string could go there as well?

packages/editor/src/plugins/rows/editor.tsx
```diff
@@ -18,7 +18,7 @@ export function RowsEditor({ state, config, id }: RowsProps) {
   function insertRowWithSuggestionsOpen(insertIndex: number) {
```
hejtful: This function's name is outdated now. It only made sense when it added a Text plugin with `/`, where the `/` would trigger a suggestions list to show.

packages/editor/src/plugins/text/hooks/use-slate-render-handlers.tsx
```diff
@@ -108,7 +110,7 @@ export const useSlateRenderHandlers = ({
           mayManipulateSiblings
             ? () => {
                 ReactEditor.focus(editor)
-                editor.insertText('/')
+                setShowSuggestions?.(true)
```
hejtful: The dependency array is now missing the expected `setShowSuggestions` dependency.

packages/editor/src/plugins/text/components/text-editor.tsx
```diff
@@ -141,7 +143,17 @@ export function TextEditor(props: TextEditorProps) {
...
+          if (e.key === '/') {
+            const text = Node.string(editor)
+            // "Show suggestions when the user types '/' and the editor is empty"
+            if (text.length === 0) {
+              setShowSuggestions(true)
+              e.preventDefault()
+            }
+          }
+          handleEditableKeyDown(e)
+        }}
```
hejtful: Why add inline code, couldn't this all go inside the `useEditableKeydownHandler` hook?

packages/editor/src/plugins/text/components/text-editor.tsx
```diff
@@ -163,11 +175,14 @@ export function TextEditor(props: TextEditorProps) {
...
-      {showSuggestions ? (
-        <SlateOverlay width={620}>
-          <Suggestions {...suggestionsProps} />
-        </SlateOverlay>
-      ) : null}
+      <ModalWithCloseButton
+        className="top-8 max-h-[90vh] w-auto min-w-[700px] translate-y-0 overflow-y-scroll pt-0"
+        isOpen={showSuggestions}
+        setIsOpen={setShowSuggestions}
+        title=""
```
hejtful: 👎

packages/editor/src/editor-ui/editor-tooltip.tsx
```diff
@@ -19,14 +21,22 @@ export function EditorTooltip({
     isMac ? '⌘' : strings.keys.ctrl
   )
 
+  const positionClasses = {
```
hejtful: This could have been done before using the `className` property. Having a `position` property is fine with me. Did you check for previous usages of EditorTooltip, to see if there were any places where you could replace the `className` method with your new method of positioning?

packages/editor/src/plugins/text/components/suggestion-item.tsx
```diff
@@ -0,0 +1,70 @@
...
+    onBlur,
+    tooltipPosition,
+  } = props
+
+  const { pluginType, title, icon, description } = option
+
+  const handleMouseEvents = {
+    onMouseEnter,
+    onMouseLeave,
+  }
```
hejtful: Minor: I don't like this, simply because it's more lines than just doing it inline.

packages/editor/src/plugins/text/components/suggestion-item.tsx
```diff
@@ -0,0 +1,70 @@
...
+  onMouseEnter: () => void
+  onMouseLeave: () => void
+  onBlur: () => void
+  tooltipPosition?: 'right' | 'left'
+}
+
+export const SuggestionItem = forwardRef<
+  HTMLButtonElement,
+  SuggestionItemProps
+>((props, ref) => {
```
hejtful:
```suggestion
>(function SuggestionItem(props, ref) {
```

And then you don't need to manually set displayName at the bottom of the file.

packages/editor/src/plugins/text/components/suggestions.tsx
```diff
@@ -1,65 +1,116 @@
...
+          selected={currentIndex === currentlyFocusedItem}
+          ref={(el) => (itemRefs.current[currentIndex] = el)}
+          onClick={(event) => handleItemClick(event, item.pluginType)}
+          onFocus={() => handleItemFocus(currentIndex)}
+          onBlur={handleItemBlur}
+          onMouseEnter={() => handleItemFocus(currentIndex)}
+          onMouseLeave={() => {
+            handleItemBlur()
+            itemRefs.current[currentIndex]?.blur()
+          }}
```
hejtful, line 86: It would be nice to do the same / a similar thing on arrow button click. Otherwise, it can happen that two tooltips are displayed one over the other.

Screenshot 2024-08-06 at 13 45 27

packages/editor/src/plugins/text/components/suggestions.tsx
```diff
@@ -1,65 +1,116 @@
...
-            </div>
-          </div>
-        )
-      })}
+    <div className="mt-2">
+      <div className="sticky top-0 z-10 bg-white pb-3 pl-6 pt-7 shadow-stickysearch">
+        <EditorInput
+          ref={searchInputRef}
+          autoFocus
+          placeholder="Search..."
```
hejtful: i18n?

packages/editor/src/plugins/text/components/suggestions.tsx
```diff
@@ -1,65 +1,116 @@
...
+          value={searchString}
+          onChange={(e) => setSearchString(e.target.value)}
```
hejtful, line 101: Minor: On line 74 you used `event`, here you're using `e`. Sticking to one, one per file at least, would be nice.

hejtful (changes requested):

packages/editor/src/plugins/text/components/suggestions.tsx
shn-srl: Fixed :heavy_check_mark:

packages/editor/src/plugins/text/components/suggestion-item.tsx
shn-srl: Thanks, Fixed :heavy_check_mark:

packages/editor/src/plugins/text/components/suggestion-item.tsx
shn-srl: Removed

packages/editor/src/plugins/text/components/text-editor.tsx
shn-srl: Removed :+1:

packages/editor/src/plugins/rows/editor.tsx
shn-srl: Renamed

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
shn-srl: Fixed :heavy_check_mark:

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
shn-srl: Fixed :heavy_check_mark:

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
shn-srl: Restored

packages/editor/src/plugin/helpers/get-plugin-title.tsx
shn-srl: Fixed, thanks

e2e-tests/tests/420-text-plugin.ts
shn-srl: Restored

packages/editor/src/plugins/text/hooks/use-slate-render-handlers.tsx
shn-srl: Fixed

apps/web/src/serlo-editor-integration/create-plugins.tsx
shn-srl: Fixed

packages/editor/src/plugins/text/components/text-editor.tsx
shn-srl: Thanks, fixed

e2e-tests/tests/420-text-plugin.ts
shn-srl: Created a ticket for this: https://linear.app/serlo/issue/PE-61/create-helper-functions-for-tests

packages/editor/src/editor-ui/editor-tooltip.tsx
shn-srl: Created a ticket for this too since there are a lot of places to be checked

https://linear.app/serlo/issue/PE-62/standardise-tooltip-usage

CodingDive:
Try the keyboard navigation with tab and hit enter. It always opens n-1 of the focused plugin for me. E.g I have the Code plugin in focus and it'll open Geogebra upon pressing Enter. 
image


Box opens spoiler, etc.

packages/editor/src/plugins/text/components/suggestions.tsx
shn-srl: This seems quite tricky, will have to investigate a bit more. Will work on fixing the e2e tests first

shn-srl:
> Try the keyboard navigation with tab and hit enter. It always opens n-1 of the focused plugin for me. E.g I have the Code plugin in focus and it'll open Geogebra upon pressing Enter. image
> 
> Box opens spoiler, etc.

Well spotted thanks ! Fixed now

apps/web/shared/tailwind-base.js
```diff
@@ -110,6 +110,8 @@ export default {
         menu: 'rgba(0, 0, 0, 0.2) 0px 2px 4px',
         modal: '0px 0px 10px 0px rgba(0, 0, 0, 0.25)',
         input: `0 0 4px 0 ${brand}`,
+        suggestions: '2px 2px 8px -1px rgba(51,51,51,0.55)',
+        stickysearch: ' 0px 7px 7px 3px rgba(255,255,255,0.8)',
```
elbotho: Just checking:
Are those visually clearly different from the ones we already created or the default tailwind ones?
I usually try to choose one of the existing ones, that actually helps the visual consistency (and if it's not very different is also in Gregors interest I believe).

apps/web/src/components/modal-with-close-button.tsx
```diff
@@ -90,20 +93,27 @@ export function ModalWithCloseButton({
...
```
elbotho: by the way `sr-only` [from tailwind](https://tailwindcss.com/docs/visibility) does the same thing as radix' `VisuallyHidden` so we should don't need another package here.

packages/editor/src/editor-ui/assets/plugin-icons/icon-fill-the-gap.svg
elbotho, line 1: @GregorZupan we should probably rework those to fit the style / quality of the other plugins a bit better.

packages/editor/src/plugins/multimedia/editor.tsx
```diff
@@ -99,7 +99,11 @@ export function MultimediaEditor(props: MultimediaProps) {
         <AreImagesDisabledInTableContext.Provider value>
           <MultimediaRenderer
             media={<>{multimedia.render()}</>}
-            explanation={<>{explanation.render()}</>}
+            explanation={
+              <>
+                {explanation.render({ config: { isInlineChildEditor: true } })}
+              </>
+            }
```
elbotho, line 106: Why did you add this here?

packages/editor/src/plugins/text/components/suggestions.tsx
```diff
@@ -1,65 +1,126 @@
...
+    itemRefs,
+    currentlyFocusedItem,
+    setCurrentlyFocusedItem,
+    insertSelectedPlugin,
+  } = props
+
   const editorStrings = useEditorStrings()
 
-  if (options.length === 0) {
-    return <div>{editorStrings.plugins.text.noItemsFound}</div>
```
elbotho, line 25: If you remove the last usage of a string, please also remove it from the i18n file. 
But in this case I feel like we should have some kind of fallback if a category or at least if all categories are empty.
@GregorZupan did you think about something here?

packages/editor/src/plugins/text/components/suggestions.tsx
```diff
@@ -1,65 +1,126 @@
...
+  useEffect(() => {
+    const focusTimeout = setTimeout(() => {
+      searchInputRef.current?.focus()
+    }, 0)
+    return () => clearTimeout(focusTimeout)
+  }, [searchInputRef])
```
elbotho, line 42: If this is needed it means the `autoFocus` on the `EditorInput` does not work, right?
Do you figure out why?

packages/editor/src/plugins/text/components/suggestions.tsx
elbotho: (interesting edge cases when I move the focus with the keyboard, trigger a scroll and that triggers a mouseenter on a previous item 😶‍🌫️)

packages/editor/src/plugins/text/components/suggestion-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+      onMouseLeave={onMouseLeave}
+      className={cn(
+        'group serlo-tooltip-trigger flex cursor-auto flex-col items-center rounded-md border border-2 border-transparent pb-0'
+      )}
+    >
+      <EditorTooltip position={tooltipPosition} text={description} />
+      <span
+        className={cn(
+          'w-full cursor-pointer rounded-md p-2 hover:shadow-suggestions',
+          selected && 'shadow-suggestions'
```
elbotho: @GregorZupan is that the intended design for a selected/hovered item (with the shadow + the outline)?
image

packages/editor/src/plugins/text/components/text-editor.tsx
```diff
@@ -163,11 +166,15 @@ export function TextEditor(props: TextEditorProps) {
...
-          <Suggestions {...suggestionsProps} />
-        </SlateOverlay>
-      ) : null}
+      <ModalWithCloseButton
+        className="top-8 max-h-[90vh] w-auto min-w-[700px] translate-y-0 overflow-y-scroll pt-0"
+        isOpen={showSuggestions}
+        setIsOpen={setShowSuggestions}
+        altText={editorStrings.addPluginsModal.altText}
+      >
+        {/* TODO: Remove added text plugin if user closes modal without selecting text plugin */}
```
elbotho, line 175: maybe now it's easier to not create a text plugin in the first place?

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
```diff
@@ -91,147 +113,205 @@ export const useSuggestions = (args: useSuggestionsArgs) => {
...
-  const description =
-    pluginStrings && Object.hasOwn(pluginStrings, 'description')
-      ? pluginStrings.description
-      : pluginData.plugin.defaultDescription
-  const { icon } = pluginData
+  const title: string =
+    pluginStrings?.title ?? pluginData.plugin.defaultTitle ?? pluginType
+  const description: string | undefined =
+    pluginStrings?.description ?? pluginData.plugin.defaultDescription
+  const icon: JSX.Element | undefined = pluginData.icon
```
elbotho: why do you initialize the variables with types in this file?

elbotho (commented):
Some questions / ideas / feedback.

elbotho:
@GregorZupan separation in Basic Plugins and Interactive Plugins feels a bit weird (Geogebra/Video are interactive). 
Maybe Content and Exercises?

packages/editor/src/editor-ui/assets/plugin-icons/icon-fill-the-gap.svg
GregorZupan: Which one? Fill in the Gap (typing)?

packages/editor/src/plugins/text/components/suggestions.tsx
GregorZupan: Can you elaborate? Not sure what you are referring to

packages/editor/src/plugins/text/components/suggestion-item.tsx
GregorZupan: only when using the keyboard, on mouse hover the outline should not appear (but removing it is not suuuuper crucial)

packages/editor/src/plugins/text/components/text-editor.tsx
GregorZupan: If that's possible, I'd welcome it :)

packages/editor/src/editor-ui/editor-tooltip.tsx
```diff
@@ -19,14 +22,22 @@ export function EditorTooltip({
...
+    top: 'bottom-full mb-2.5',
+    right: 'left-full top-1/3 transform -translate-y-1/3 translate-x-2.5',
+    left: 'right-full top-1/3 transform -translate-y-1/3 -translate-x-2.5',
+    bottom: 'top-full mt-2.5',
+  }
```
elbotho: I also don't mind the position approach in general but for those tooltips from a design perspective I'd prefer them being on the same vertical position. So basically I'd want them to be position top, but aligned `-left-6` `-right-6` for the sides. That would be a lot easier and cleaner with `className`,

GregorZupan:
@elbotho the copy of the category is definitely up for debate ;)

apps/web/src/components/modal-with-close-button.tsx
shn-srl: Thanks, cool fix, removed

packages/editor/src/plugins/text/components/suggestions.tsx
elbotho: Sure. If you search for "Tetris" for example you get no results 😿. 
Only this view that is not really helpful:
image

shn-srl: @GregorZupan  :

@elbotho is saying we should have a "no basic plugins found" or similar text under each category heading if there are no filtered results

packages/editor/src/editor-ui/assets/plugin-icons/icon-fill-the-gap.svg
elbotho: all but Dropzones:
image
Or alternatively only Dropzones and Non-Interactive.

packages/editor/src/plugins/text/components/suggestions.tsx
elbotho: I wonder if decoupling mouse interaction and keyboard interaction would be easier/better here. 
Hovering an item would not move the focus to that item but show the hover state as before. Focus would then usually stay on the input field for mouse users. You can still start filtering by typing at any point.
Double tooltips would be a (maybe ignorable) edge case.
What do you think?

packages/editor/src/plugins/text/components/suggestion-item.tsx
GregorZupan: I can only replicate this issue once i interqcted with my keyboard and **then** use the mouse. Once i click into empty space, this behaviour stops

packages/editor/src/editor-ui/editor-tooltip.tsx
shn-srl: Pushed a change, it seems `-right-0` and `-left-0` look good, please let me know what you think.

packages/editor/src/plugins/text/hooks/use-suggestions.tsx
shn-srl: Removed

apps/web/shared/tailwind-base.js
shn-srl: The stickysearch is definitely a unique box shadow situation.

Will check with Gregor if the one of the standard tailwind ones is acceptable for the suggestion items

packages/editor/src/plugins/multimedia/editor.tsx
shn-srl: The current implementation of the Suggestions and how the modal opens, required this prop.

```

const [showSuggestions, setShowSuggestions] = useState(
    !isInlineChildEditor &&
      focused &&
      filteredOptions.length > 0 &&
      !isSelectionWithinList(editor)
  )
  
  ```

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 24 above)
Comment (shn-srl) [reply packages/editor/src/plugins/text/components/suggestions.tsx:42]: Thats correct, it doesn't work here. 

I assume it's related to editor focus hacks, or because it's inside a modal. I've created a [ticket](https://linear.app/serlo/issue/PE-64/investigate-autofocus-in-editorinput) in Linear to investigate.

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 17 above)
Comment (shn-srl) [reply packages/editor/src/plugins/text/components/suggestions.tsx:86]: I think the reason I restored the mouse interaction and focus change was for the case where the user hovers the item with the mouse and hits the enter key, it should insert the hovered plugin

Comment (shn-srl) [reply packages/editor/src/plugins/text/components/suggestions.tsx:86]: Probably a good idea to discuss this on a call to see what the best case behavior is

File: packages/editor/src/editor-ui/assets/plugin-icons/icon-fill-the-gap.svg
Comment (GregorZupan) [reply packages/editor/src/editor-ui/assets/plugin-icons/icon-fill-the-gap.svg:1]: The icons are not accurate yet anyway, i'd wait

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 23 above)
Comment (GregorZupan) [reply packages/editor/src/plugins/text/components/suggestions.tsx:25]: So we just need some form of emty state icon/illustration/notification?

Comment (GregorZupan) [reply packages/editor/src/plugins/text/components/suggestions.tsx:25]: we should def implement a tetris plugin!

File: apps/web/shared/tailwind-base.js (hunk 20 above)
Comment (shn-srl) [reply apps/web/shared/tailwind-base.js]: Removed the custom suggestion shadow and used the tailwind one

File: packages/editor/src/plugins/text/components/text-editor.tsx (hunk 26 above)
Comment (shn-srl) [reply packages/editor/src/plugins/text/components/text-editor.tsx:175]: The whole suggestions logic still lives in the text plugin, investigating how to avoid that but I do believe it's out of scope for this ticket

File: packages/editor/src/plugins/multimedia/editor.tsx (hunk 22 above)
Comment (elbotho) [reply packages/editor/src/plugins/multimedia/editor.tsx:106]: I'm not sure about the implementation, I'd have to check.
But only the multimedia explanation used for the article introduction is actually inline, usually explanations are not inline so this makes so I don't think we should set this here.

Also this PR changes the current behaviour. Suggestions should open inside a normal multimedia plugin but not inside an Article introduction. Compare:
- https://de.serlo-staging.dev/entity/repository/add-revision/277232 (staging)
- https://frontend-git-240723-plugin-selection-menu-serlo.vercel.app/entity/repository/add-revision/277232 (this PR)

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 23 above)
Comment (elbotho) [reply packages/editor/src/plugins/text/components/suggestions.tsx:25]: > So we just need some form of emty state icon/illustration/notification?

I think that would be nice. We could also remove the headlines for categories without results and only have a fallback message when there is actually no plugin found overall.

File: packages/editor/src/editor-ui/editor-tooltip.tsx (hunk 28 above)
Comment (elbotho) [reply packages/editor/src/editor-ui/editor-tooltip.tsx]: Thank you.
Personally I liked `-[direction]-6` better, that's why suggested it but I don't mind `[direction]-0`.
(but `-[direction]-0` hurts a tiny bit 😉)

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 23 above)
Comment (GregorZupan) [reply packages/editor/src/plugins/text/components/suggestions.tsx:25]: I'll come up with something! We will remove the headers then and have either a quick note or illustration.

File: packages/editor/src/plugins/text/components/suggestions.tsx (hunk 17 above)
Comment (elbotho) [reply packages/editor/src/plugins/text/components/suggestions.tsx:86]: > I think the reason I restored the mouse interaction and focus change was for the case where the user hovers the item with the mouse and hits the enter key, it should insert the hovered plugin

For that you could also query for items with `:hover` on enter.

File: packages/editor/src/plugins/multimedia/editor.tsx (hunk 22 above)
Comment (shn-srl) [reply packages/editor/src/plugins/multimedia/editor.tsx:106]: I don't think it makes sense that the user clicks add plugin, then the modal opens, then they click "Multimedia plugin" and straight away see another Suggestions Modal. 

Will compare tomorrow and see what an alternative is

File: apps/web/src/components/modal-with-close-button.tsx [hunk 29]
```diff
@@ -90,20 +92,25 @@ export function ModalWithCloseButton({
...
-              </Dialog.Title>
-            ) : null}
+            <Dialog.Title
+              className={cn(
+                'serlo-h2',
+                extraTitleClassName,
+                title ? '' : 'sr-only'
+              )}
+            >
+              {title || altText}
```
kulla, line 104: Why have you added `altText` like this in the modal? For me I would rather have a call like

```typescript
<ModalWithCloseButton text={title ?? altText}/>
```

apps/web/src/serlo-editor-integration/create-plugins.tsx
```diff
@@ -189,16 +194,35 @@ export function createPlugins({
       plugin: exercisePlugin,
       visibleInSuggestions: true,
     },
-    { type: EditorPluginType.Solution, plugin: solutionPlugin },
-    { type: EditorPluginType.H5p, plugin: H5pPlugin },
+    {
+      type: EditorPluginType.Solution,
+      plugin: solutionPlugin,
+      visibleInSuggestions: true,
```
kulla, line 200: Because I am curios: Why is this change necessary?

packages/editor/src/plugins/rows/components/row-editor.tsx
```diff
@@ -28,7 +28,8 @@ export function RowEditor({
   const plugins = editorPlugins.getAllWithData()
   const dropContainer = useRef<HTMLDivElement>(null)
 
-  const isLastRow = index === rows.length - 1
+  // TODO: Remove this after reviewing behaviour
+  // const isLastRow = index === rows.length - 1
```
kulla, line 32: Leftover todo :smile:

packages/editor/src/plugins/rows/index.tsx
```diff
@@ -53,6 +53,7 @@ export interface RowsPluginConfig {
   allowedPlugins?: (EditorPluginType | string)[]
   parentType: string
   textPluginPlaceholder?: string
+  isInlineChildEditor?: boolean
```
kulla, line 56: What is the meaning of this flag?

kulla (changes requested):

kulla (changes requested):
Nice, I like the new add button + the design of the new popup. Things I have noticed:

### Bug with adding multimedia plugin

After adding the multimedia plugin the add button does not seem to work again: 
2024-08-07-203652_849x410_scrot

### Bug with adding interactive plugins

Adding interactive plugins (sc/mc, non-interactive-exercise, input-field, ...) opens the pop up again: 
2024-08-07-203345_181x203_scrot

It also seems that some of those plugins are integrated without a compounding exercise plugin...

packages/editor/src/plugins/text/components/text-editor.tsx
kulla: > The whole suggestions logic still lives in the text plugin

After the change of the issue there should be IMHO a refactoring to move the code out into the rows plugin. still integrating a text plugin might be a reason for bugs in the future...

apps/web/src/components/modal-with-close-button.tsx
```diff
@@ -90,20 +92,25 @@ export function ModalWithCloseButton({
...
-            {title ? (
-              <Dialog.Title className={cn('serlo-h2', extraTitleClassName)}>
-                {title}
-              </Dialog.Title>
-            ) : null}
+            <Dialog.Title
+              className={cn(
+                'serlo-h2',
+                extraTitleClassName,
+                title ? '' : 'sr-only'
```
kulla, line 101: This looks strange. `sr-only` should be part of `extraTitleClassName`

packages/editor/src/plugins/multimedia/editor.tsx
kulla: Is it intentional, that in a multimedia plugin with this change no new elements can be added on the left side of the image? (@GregorZupan)

kulla (changes requested):

apps/web/src/components/modal-with-close-button.tsx
kulla: From our meeting:

`<ModelWithCloseButton title="..." extraTitleClassName="sr-only" />`

----

* ModelWithCloseButton { screenReaderTitle: string }
* ModelWithTitleAndCloseButton { title: string, extraTitleClassName?: }

hejtful:
@kulla @shn-srl 

> After adding the multimedia plugin the add button does not seem to work again:

The add button does appear, it's under both the multimedia explanation and the multimedia itself.

Screenshot 2024-08-08 at 13 48 45

kulla (commented):
@shn-srl from our meeting today, here are my notes:

```
Beforhand:

* Only one single plugin could be added => list of addible plugins => flag for each plugin

After:

* Single plugins + compound of plugins
* Add button for MultipleChice
* Add button for Single Choice

const suggestionConfig = [{
  title: "Interaktive Aufgabe",
  elements: [
    { icon: …, name: ..., state: …
   }
   ]
```

shn-srl:
Closing this for new approach, see PR https://github.com/serlo/frontend/pull/4017

PR #4012, opened 2024-08-02
e2e: fix failing tests

shn-srl:
Fixes failing dropzone tests and focus tests for deployment


shn-srl:
I can't currently reproduce the last failing test locally. Have tried it with and without a Pixabay env variable, doesn't seem to have an impact. Will take a further look on monday

LarsTheGlidingSquirrel (approved):
Maybe include a comment in `ImageEditor` that `ImageProps.focused` should be true when the buttons are focused but is not for some reason.

PR #4017, opened 2024-08-09
feat(editor): Redesigned Plugin Selection

shn-srl:
[Demo Link](https://frontend-git-new-add-plugin-modal-serlo.vercel.app/___editor_preview)

[Issue Link](https://linear.app/serlo/issue/PE-34/ui-redesign-plugin-selection)

**Todo: **

- [x] Fix Tests
      Most tests are fixed, flakiness when running locally is making it unsustainable to keep testing locally, as soon as this PR gets merged I will make a follow up PR for any failing tests.
      
      
- [x] Include all plugins (`visibleInSuggestions`)
- [x] Code cleanup

- [x] Usability Test 
- Output: 
  - [x]  blue focus box size same as shadow
  - [x]  Uniform SVG sizes
  - [x]  Less space on 2nd header
  - [x]  x button inside search input field (maybe new ticket)
               New Ticket: https://linear.app/serlo/issue/PE-73/add-clear-x-button-to-editor-input
  - [x] add row floating button above first plugin


hejtful:
Nice improvement over the first PR, imo. 👍 

A couple of suggestions:
- Rather than rendering the context provider in the `InnerDocument`, render it in the `RowsEditor.` Multiple scoped contexts are better for performance, as any changes to it will only trigger re-renders within that one `RowsEditor`, not the whole editor.
- Consider merging the `AllowedChildPlugins` context with your new context, as they share the same concern, adding plugins.
- `useReducer` in your new context might make the code cleaner and easier to understand.

shn-srl:
> Nice improvement over the first PR, imo. 👍
> 
> A couple of suggestions:
> 
>     * Rather than rendering the context provider in the `InnerDocument`, render it in the `RowsEditor.` Multiple scoped contexts are better for performance, as any changes to it will only trigger re-renders within that one `RowsEditor`, not the whole editor.
> 
>     * Consider merging the `AllowedChildPlugins` context with your new context, as they share the same concern, adding plugins.
> 
>     * `useReducer` in your new context might make the code cleaner and easier to understand.

Thanks I've implemented what you've suggested, the first two points look good.

I'm not sure if I'm making the cleanest (or most sensible use) of useReducer but in a way I guess it is more readable and stable

shn-srl:
Sub ticket created 

https://linear.app/serlo/issue/PE-69/split-filinthegaps-plugin

packages/editor/src/plugins/rows/components/rows-inner-editor.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  function handleInsertPlugin(pluginType: EditorPluginType) {
+    const isInteractivePlugin = interactivePluginTypes.has(pluginType)
+
+    const pluginToInsert = isInteractivePlugin
+      ? wrapExercisePlugin(pluginType)
+      : { plugin: pluginType }
+
+    pluginMenuState.insertCallback?.()
```
kulla:
```suggestion
    pluginMenuState.onBeforeInsert?.()
```

kulla (changes requested):

e2e-tests/tests/440-box-plugin.ts
```diff
@@ -100,13 +101,13 @@ Scenario(
   }
 )
 
-Scenario('Set box type using keyboard', async ({ I }) => {
+Scenario.skip('Set box type using keyboard', async ({ I }) => {
```
shn-srl: Todo: Fix

packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx
```diff
@@ -41,11 +48,38 @@ export const useEditableKeydownHandler = (
...
+          const parent = selectChildTreeOfParent(store.getState(), id)
+
+          if (Object.hasOwn(node, 'text') && node.text.length === 0 && parent) {
+            pluginMenuDispatch({
+              type: PluginMenuActionTypes.OPEN_WITH_SLASH_KEY,
+              payload: {
+                insertIndex: parent.children?.findIndex(
+                  (child) => child.id === id
+                ),
+                onInsertComplete: () => {
```
shn-srl: @hejtful I think this is undesired behaviour, perhaps we can discuss this on a call? 

I introduced the callback for a different reason, after your changes it was unnecessary for me to use it that way. I see that its removing the text plugin here and am slightly unsure why

An example of how this is broken is seen with the failed test, the last scenario in 423-text-plugin-list.ts

shn-srl:
@elbotho @hejtful 

Would appreciate another review round :pray: 

I am conducting a full usability test with Gregor today at 16, and still looking over the tests to ensure nothing is failing

packages/editor/src/core/inner-document.tsx
```diff
@@ -107,9 +106,7 @@ export function InnerDocument({ children, onChange, ...props }: EditorProps) {
 
   return (
     <div className="relative" ref={wrapperRef}>
-      <PreferenceContextProvider>
-        <EditorChildren>{children}</EditorChildren>
-      </PreferenceContextProvider>
+      <EditorChildren>{children}</EditorChildren>
     </div>
```
elbotho: why did you remove this?

packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx
shn-srl: Removing the onInsertComplete callback shows that the insertIndex is incorrect, and the new plugin gets inserted before the existing one

shn-srl: I have added a change (hopefully fix) in `packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx: 66`

elbotho: Could we maybe simplify this by handling Click and Slash behaviour in the same way?
I think removing the current text-plugin if it's empty would work nicely for both.

elbotho (commented):
looks good so far.
- tiny ux think (optional): when pressing `esc` and there is a search string I expect the search to to `""`. If the input is empty already and I press esc then I expect the modal to close
- there is already a discussion about this but right now I can't replace the text-plugin with the new plugin I select, it always get's appended

e2e-tests/tests/440-box-plugin.ts
hejtful: You can also use `Scenario.todo`, if you so desire

packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx
shn-srl: Open with slash key is opening the Modal from within a text plugin, while normal open is clicking the Floating add button, which to me as a User I would expect to add an additional row

packages/editor/src/core/inner-document.tsx
shn-srl: Restored

packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx
hejtful: The way I understood it:
- Click on the plus button between two rows:
  1. Open the modal
  2. Select the plugin
  3. Insert the selected plugin where the button was
- Slash in an empty Text plugin:
  1. Open the modal
  2. Select the plugin
  3. Insert the selected plugin after the empty Text plugin
  4. Remove the empty Text plugin

@elbotho Are we aligned here? If yes, could you please make a more concrete suggestion about handling both in the same way / removing the current text plugin for both?

elbotho: but I would also not mind to replace an empty text plugin. Currently the only way to add a new plugin in place of the currently selected empty text plugin is to add it and the manually remove the text plugin.

elbotho: I'm talking about this case:

- Click on the plus button between two rows when the button is below an empty text plugin:
  1. Open the modal
  2. Select the plugin
  3. Insert the selected plugin after the empty Text plugin
  4. Remove the empty Text plugin

Which would be the steps as `Slash in an empty Text plugin` so that's why I wondered if we could simplify:

``` diff
   OPEN = 'OPEN',
-  OPEN_WITH_SLASH_KEY = 'OPEN_WITH_SLASH_KEY',
   CLOSE = 'CLOSE',
  ```

Comment (shn-srl) [reply packages/editor/src/plugins/text/hooks/use-editable-key-down-handler.tsx]: I think @GregorZupan should have some input here, I don't agree that i makes sense to remove the empty text plugin if they clicked on the button

**Comment by shn-srl** [comment]
> * there is already a discussion about this but right now I can't replace the text-plugin with the new plugin I select, it always get's appended

Fixed

**Comment by hejtful** [comment]
Search bar autofocus doesn't work for me in local, on Chrome. Not sure if you're already on it

**Comment by shn-srl** [comment]
> Search bar autofocus doesn't work for me in local, on Chrome. Not sure if you're already on it

It works for me in Chromium and Firefox

Could it be your mouse is over a plugin Item when the modal opens and steals the focus?

**Comment by hejtful** [comment]
@shn-srl 
>Could it be your mouse is over a plugin Item when the modal opens and steals the focus?

Yes, but won't that happen most of the time? Can we expect the user's mouse not to be on a plugin item?

**Comment by hejtful** [comment]
@shn-srl If you change `onMouseEnter` to `onMouseMove` in `PluginMenuItem`, the bug seems to be fixed. Just a suggestion.

**Comment by GregorZupan** [comment]
Good Idea @hejtful

**Comment by shn-srl** [comment]
> @shn-srl If you change `onMouseEnter` to `onMouseMove` in `PluginMenuItem`, the bug seems to be fixed. Just a suggestion.

Done

File: e2e-tests/tests/440-box-plugin.ts (hunk 2 above)
Comment (shn-srl) [reply e2e-tests/tests/440-box-plugin.ts]: Thanks, fixed the test

**Comment by shn-srl** [comment]
Updated the description regarding the tests:

  Most tests are fixed, flakiness when running locally is making it unsustainable to keep testing locally, as soon as this PR gets merged I will make a follow up PR for any failing tests.

**Comment by hejtful** [comment]
>Most tests are fixed, flakiness when running locally is making it unsustainable to keep testing locally, as soon as this PR gets merged I will make a follow up PR for any failing tests.

Sounds good to me. Just to mention, if you run `yarn build` then `yarn start` before running tests, it reduces the flakiness compared to just `yarn dev`.

**Comment by hejtful** [comment]
Sometimes, the shortcuts won't work after the modal is opened. They start working after a few clicks on the modal area. Can't find a stable reproduction method, it's on Chrome. Can you reproduce it?

Edit: Fixed by https://github.com/serlo/frontend/pull/4017/commits/96e41ed161a8736352c0d375d074aa66a1f7e99a

**Comment by hejtful** [comment]
Up and Down arrows don't work in the Text plugin any more. They should move the cursor a line up or down, but instead do nothing.

Edit: Fixed by https://github.com/serlo/frontend/pull/4017/commits/43404ac101fdcebfa341aac10ed34ee187021b65

**Review by hejtful — APPROVED** [review]

**Comment by shn-srl** [comment]
From Botho:

One weird case: When I insert some plugins (e.g. equations & code) with enter I get an additional textplugin below my inserted plugin (I guess the enter propagates)

**Comment by hejtful** [comment]
@elbotho Fixed the additional Text plugin bug with: https://github.com/serlo/frontend/pull/4017/commits/1d863e745ee569d28878ec5cf0f839954c02f355

Could you please confirm that the removed code is deprecated? I believe it's from when we would add a new Text plugin on Enter key, but we moved away from that feature.

File: packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx [hunk 5]
```diff
@@ -0,0 +1,224 @@
...
+      isEmpty: basicOptions.length === 0 && interactiveOptions.length === 0,
+    }
+  }, [allowedPlugins, pluginsStrings, searchString])
+
+  const handleItemFocus = (index: number) => {
+    setFocusedItemIndex(index)
+  }
+
+  const handleItemBlur = () => {
+    setFocusedItemIndex(-1)
```
elbotho: you use null everywhere else, is that `-1` on purpose?

elbotho:
> Could you please confirm that the removed code is deprecated? I believe it's from when we would add a new Text plugin on Enter key, but we moved away from that feature.

I think it is. Thank you.

packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx
elbotho: (changed in https://github.com/serlo/frontend/pull/4021)

hejtful: >you use null everywhere else, is that -1 on purpose?

I introduced the `null` and overlooked this one instance. My bad.

packages/editor/src/plugins/rows/components/plugin-menu-item.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <li>
+      <button
+        data-qa={`plugin-suggestion-${pluginType}`}
+        ref={ref}
+        onClick={() => onInsertPlugin(pluginType)}
+        onFocus={onFocus}
+        onBlur={onBlur}
+        onMouseMove={onMouseMove}
```
elbotho: @shn-srl do you see a downside to using `onMouseEnter` here?
(upsides: less events and I think the bug with focus stealing when the modal scrolls because of arrow key usage would be solved)

hejtful: @elbotho For context:
- https://github.com/serlo/frontend/pull/4017#issuecomment-2289112270
- https://github.com/serlo/frontend/pull/4017#issuecomment-2289214314

PR #4021, opened 2024-08-15
small keyboard improvements for plugin menu

elbotho:
- `enter` inserts first option if there is one
- `escape` does not close modal if there is text in the filter input but remove the text first

### DIFF
--- a/packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx
+++ b/packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx
@@ -60,33 +60,37 @@ export function PluginMenuModal({ onInsertPlugin }: PluginMenuModalProps) {
       : allowedByContext
   }, [allWithData, pluginMenuState.allowedChildPlugins])
 
-  const { basicOptions, interactiveOptions, isEmpty } = useMemo(() => {
-    const allOptions = allowedPlugins.map((type) => {
-      return createOption(type as EditorPluginType, pluginsStrings)
-    })
-    const allowed = filterOptions(allOptions, searchString)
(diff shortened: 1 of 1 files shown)


hejtful:
The code looks good to me, but I see @GregorZupan as the real reviewer here, as in, is this the behaviour that we want? :)

GregorZupan:
Is there a way for me to test it?

hejtful:
@GregorZupan Sure, here: https://frontend-git-keyboard-improvements-plugin-menu-serlo.vercel.app/___editor_preview

hejtful:
@elbotho A bug:
1. Open modal, type text
2. Press Enter
3. Text plugin inserted
4. Modal re-opened

Step 4. is not desired, right?

GregorZupan:
I wasn't able to reproduce this error @hejtful. 
The look & feel was amazing, but this tested only the keyboard usage, right? Not the other issues we've been discussing today?

elbotho:
>@elbotho A bug:
> Open modal, type text
Press Enter
Text plugin inserted
Modal re-opened
Step 4. is not desired, right?

Weird and def. not desired 😅. Only happens if I open the Modal with Button click, not if using the keyboard so I suspect it might be related to the `OPEN_WITH_SLASH_KEY` code. Maybe we remove that from `new-add-plugin-modal` pr first?

hejtful:
> Only happens if I open the Modal with Button click, not if using the keyboard so I suspect it might be related to the OPEN_WITH_SLASH_KEY code.

Can't see the logic in this

>Maybe we remove that from new-add-plugin-modal pr first?

But sure, let's try

hejtful (approved):

PR #4023, opened 2024-08-16
simplify plugin menu item

elbotho:
Less dom elements, more semantic, removing a lot of unnecessary classes.
Should not change behaviour / style.

(hold on, found a small problem with the tooltip positioning)


hejtful (approved):
🧹

PR #4025, opened 2024-08-16
feat!: move custom plugins to editor package & only allow text on left side of multimedia

LarsTheGlidingSquirrel:
Changes (including some breaking changes): 
- Custom plugins `edusharingAsset` and `serloInjection` are now part of the editor package. 
- Editor API property `customPlugins` is gone as its no longer needed. 
- Editor component accepts an optional list of plugin types to make it possible to enable/disable plugins. 
- Now, multimedia plugin only allows text plugin on the left side. Before, it allowed also table, equations, code. Discussion on slack: https://serlo.slack.com/archives/C05QWHQP1HT/p1723124720805939
- Some plugin configuration simplification & removing obsolete code
- Editor API property `pluginsConfig` is gone
- Box, spoiler & multimedia have the same nesting restrictions everywhere

These changes are in preparation for making sure that the content format in the editor integrations (excluding serlo.org) is consistent. Next up: Create a shared content format (including `dateModified`, `editorVersion`, ...) and methods for performing client-side migrations. 

Potential followups: 
- Use `@radix-ui/react-dialog` instead of `react-modal` in edusharing plugin


apps/web/src/serlo-editor-integration/create-plugins.tsx
```diff
@@ -82,7 +82,20 @@ export function createPlugins({
...
+          EditorPluginType.Video,
+          ...(isProduction ? [] : [EditorPluginType.Audio]),
+          EditorPluginType.Geogebra,
+        ],
+        explanation: {
+          plugin: EditorPluginType.Rows,
+          config: {
+            allowedPlugins: [EditorPluginType.Text],
+          },
+        },
```
hejtful:
```suggestion
        explanation: {
          plugin: EditorPluginType.Text,
        },
```

Did you try this? Perhaps we don't need a Rows plugin there any more.

Or maybe we can refactor multimedia plugin in a follow-up, so that it renders explanation as a Text plugin all the time, and doesn't accept explanation as a config prop.

packages/editor/src/package/index.ts
```diff
@@ -1,9 +1,12 @@
 export { SerloEditor, type SerloEditorProps } from './editor'
 export { SerloRenderer, type SerloRendererProps } from './serlo-renderer'
 
+export { defaultPlugins } from './config'
+
 export type { BaseEditor } from '@editor/core'
 
 export { EditorPluginType } from '@editor/types/editor-plugin-type'
+export { TemplatePluginType } from '@editor/types/template-plugin-type'
```
hejtful: I'm out of the loop, why do we need to export this?

packages/editor/src/package/index.ts
```diff
@@ -1,9 +1,12 @@
 export { SerloEditor, type SerloEditorProps } from './editor'
 export { SerloRenderer, type SerloRendererProps } from './serlo-renderer'
 
+export { defaultPlugins } from './config'
```
hejtful, line 5: Can you please explain the idea behind exporting this?

apps/web/src/serlo-editor-integration/create-plugins.tsx
```diff
@@ -94,13 +107,13 @@ export function createPlugins({
...
       type: EditorPluginType.Spoiler,
-      plugin: createSpoilerPlugin(),
+      plugin: createSpoilerPlugin({}), // Passing empty config object here allows all plugins
       visibleInSuggestions: true,
       icon: <IconSpoiler />,
     },
     {
       type: EditorPluginType.Box,
-      plugin: createBoxPlugin({}),
+      plugin: createBoxPlugin(),
```
hejtful: Can you please explain why you passed the empty object to the spoiler, but removed it from the box?

hejtful (commented):
Just some questions to better understand the PR

apps/web/src/serlo-editor-integration/create-plugins.tsx
elbotho: note: without a migration we still have non text based plugins as explanations on serlo.org

hejtful: Oh right, good point. And if we want to go back, it would take a second migration...

LarsTheGlidingSquirrel: Good idea. Maybe long term, only allowing a text plugin makes sense. But for right now & since it is kind of an experiment, I would keep the rows plugin.

packages/editor/src/package/index.ts
LarsTheGlidingSquirrel: What plugins are available can be set by passing `plugins?: (EditorPluginType | TemplatePluginType)[]` in the editor API. So, we need to export these two types in the package.

packages/editor/src/package/index.ts
LarsTheGlidingSquirrel: What plugins are available can be set by passing `plugins?: (EditorPluginType | TemplatePluginType)[]` in the editor API. However, if none are passed `defaultPlugins` will be used. The edu-sharing integration wants to use `defaultPlugins` but remove `image` & `dropzone` and enable some custom plugins. Then it could look like this: 
```typescript
      plugins={[
        ...defaultPlugins.filter(
          (plugin) =>
            plugin !== EditorPluginType.Image &&
            plugin !== EditorPluginType.DropzoneImage,
        ),
        EditorPluginType.EdusharingAsset,
        EditorPluginType.SerloInjection,
        EditorPluginType.TextAreaExercise,
      ]}
```

apps/web/src/serlo-editor-integration/create-plugins.tsx
LarsTheGlidingSquirrel: When I changed the default `allowedPlugins` for the integrations that messed with the `allowedPlugins` for serlo.org. This is a quick (but very unclear) solution to restore the original nesting restrictions on serlo.org without changing more things.

packages/editor/src/package/index.ts
hejtful: If this is only an exception for Edusharing, could you please add a comment that this is exported only for Edusharing for this case?

If this is something that you imagine all integrations using, could you please document it in the Readme?

packages/editor/src/package/index.ts
hejtful: Sorry if I'm too slow 😬 but I still don't get why `TemplatePluginType` needs to be exported, and how it would be used...

LarsTheGlidingSquirrel: I removed it. Currently, it is not used and I think there will be no eslint errors. If some integration wants to customize `TemplatePluginType` plugins, we can add it again. Thank you for questioning it.

packages/editor/src/package/index.ts
LarsTheGlidingSquirrel: That would be helpful. I added a code comment. Other integration should try to avoid this.

packages/editor/src/editor-integration/create-basic-plugins.tsx
```diff
@@ -145,23 +187,6 @@ export function createBasicPlugins(config: Required<PluginsConfig>) {
...
-
-function createMultimediaPluginConfig(config: Required<PluginsConfig>) {
-  const { general, multimedia } = config
 
-  return {
-    ...multimedia,
-    allowedPlugins: multimedia.allowedPlugins.filter((allowedPlugin) => {
-      if (
-        // If the user didn't provide the Serlo Editor testing secret,
-        // remove the Image plugin from Multimedia config allowed plugins
```
hejtful, line 158: How is this logic handled now?

hejtful (commented):
Just one more question, sorry

packages/editor/src/editor-integration/create-basic-plugins.tsx
LarsTheGlidingSquirrel: Good find! I misunderstood this. With the new changes the image plugin will be deactivated when the testing secret is missing.

hejtful (approved):

PR #4026, opened 2024-08-19
feat(plugin-rows): larger last add plugin button

hejtful:
Task: https://linear.app/serlo/issue/PE-31/redesign-add-plugin-button

Screenshot 2024-08-19 at 12 00 29
Screenshot 2024-08-19 at 12 00 34


elbotho (approved):
👍

PR #4029, opened 2024-08-19
fix(editor): add plugin modal follow-ups

hejtful:
- Fix e2e tests
- Remove unused code
- Clean up custom dynamic placeholder logic (also fixes a focus bug)


e2e-tests/tests/410-plugin-toolbar.ts
```diff
@@ -14,22 +14,24 @@ Scenario('Duplicate plugin', async ({ I }) => {
 
   I.seeNumberOfElements('$plugin-text-editor', initialTextPluginCount)
 
-  addNewTextPlugin(I)
-  I.seeNumberOfElements('$plugin-text-editor', initialTextPluginCount + 1)
+  I.click(locate('$plugin-text-editor').inside('.plugin-rows'))
+  I.type('Test')
```
elbotho: what is the `.type()` for (if you don't check it later)?

e2e-tests/tests/420-text-plugin.ts
```diff
@@ -20,24 +20,11 @@ function focusTextPlugin(I, position = 'first') {
...
-  I.seeNumberOfElements(selectors.textEditor, initialTextPluginCount + 1)
-})
-
 Scenario('Add new line in plugin using Enter', async ({ I }) => {
   I.amOnPage(pageUrl)
   I.seeNumberOfElements(selectors.textEditor, initialTextPluginCount)
 
-  addNewTextPlugin(I)
-  I.seeNumberOfElements(selectors.textEditor, initialTextPluginCount + 1)
+  I.click(locate(selectors.textEditor).inside('.plugin-rows'))
```
elbotho, line 28: I don't mind tests to be verbose. In this case you could consider maybe moving this to a helper. This way it would also be easier to read (was not sure that this does at first).

packages/editor/src/store/focus/selectors.ts
```diff
@@ -12,27 +10,3 @@ export const selectIsFocused = createSelector(
...
-  [(state: State) => state, (_state, id: string) => id],
-  (state, id: string) => {
-    const tree = selectChildTree(state, id)
-    if (!tree || !tree.children) return false
-    const focused = selectFocused(state)
-    return R.any((node) => node.id === focused, tree.children)
-  }
-)
-
-export const selectHasFocusedDescendant = createSelector(
```
elbotho, line 26: crazy that we only needed this for the placeholder, would have thought we use it for other behaviours as well.

elbotho (approved):

packages/editor/src/store/focus/selectors.ts
hejtful: We didn't use it for placeholder either 😁

e2e-tests/tests/420-text-plugin.ts
hejtful: I added `I.say` statements to improve readability.

packages/editor/src/store/focus/selectors.ts
elbotho: ah okay :)

PR #4045, opened 2024-08-26
feat(editor-package)!: Add storage format and client-side migrations.

LarsTheGlidingSquirrel:
# TODO
- [x] Implement `editorVersion` according to schema.org / AMB -> Sadly, I found nothing to use. 

# Changes
- Breaking: Now, the editor package takes and outputs state of type `StorageFormat` instead of just the editor state `{plugin: ..., state: ...}` 
``` ts
type StorageFormat {
  // Constant values (set at creation)
  id: string // uuid
  type: 'https://serlo.org/editor',
  variant: 'edusharing' | 'lti-tool' | 'serlo-org' | ...,

  // Variable values (can change when state modified)
  version: number, // Index of the next migration to apply (if there is one). Example: 2 -> Apply migration[2], migration[3], ... until end of array
  editorVersion: string, // For example 0.10.1
  dateModified: string,
  document: {
    plugin: string,
    state: unknown,
  }
}
```
- Breaking: New required parameter `editorVariant` in editor API
- Breaking: Parameter `onChange`

# Notes

- `SerloRenderer` also performs migrations but content cannot be saved. 
- serlo.org is unaffected by this since it does not consume the editor npm package yet. In the future it should however. 

# Followups

- Migrate content in database on serlo.org to new storage format & consume editor package

### DIFF
--- a/packages/editor/src/package/editor.tsx
+++ b/packages/editor/src/package/editor.tsx
@@ -8,6 +8,7 @@ import { LtikContext } from '@editor/plugins/edusharing-asset/ltik-context'
 import { EditorPluginType } from '@editor/types/editor-plugin-type'
 import { SupportedLanguage } from '@editor/types/language-data'
 import { TemplatePluginType } from '@editor/types/template-plugin-type'
+import { getCurrentDatetime } from '@editor/util/get-current-datetime'
 import React from 'react'
 
 import { defaultSerloEditorProps } from './config'
@@ -16,12 +17,12 @@ import { getEditorVersion } from './editor-version'
 import {
   type StorageFormat,
   createEmptyDocument,
-  getCurrentDatetime,
   migrate,
   type EditorVariant,
 } from './storage-format'
 import { InstanceDataProvider } from '@/contexts/instance-context'
 import { LoggedInDataProvider } from '@/contexts/logged-in-data-context'
+
 import '@/assets-webkit/styles/serlo-tailwind.css'
 
 export interface SerloEditorProps {
--- a/packages/editor/src/package/editor.tsx
+++ b/packages/editor/src/package/editor.tsx
@@ -105,11 +105,9 @@ export function SerloEditor(props: SerloEditorProps) {
         const document = payload.getDocument()
         if (!document) return null
         return {
+          ...migratedState,
           dateModified,
           editorVersion: getEditorVersion(),
-          type: migratedState.type,
-          variant: migratedState.variant,
-          version: migratedState.version,
           document,
         }
       },
--- a/packages/editor/src/package/editor.tsx
+++ b/packages/editor/src/package/editor.tsx
@@ -1,4 +1,5 @@
 import { Editor, type EditorProps } from '@editor/core'
+import { type GetDocument } from '@editor/core/types'
 import { createBasicPlugins } from '@editor/editor-integration/create-basic-plugins'
 import { createRenderers } from '@editor/editor-integration/create-renderers'
 import { editorPlugins } from '@editor/plugin/helpers/editor-plugins'
@@ -9,18 +10,30 @@ import { SupportedLanguage } from '@editor/types/language-data'
 import { TemplatePluginType } from '@editor/types/template-plugin-type'
 import React from 'react'
 
-import '@/assets-webkit/styles/serlo-tailwind.css'
 import { defaultSerloEditorProps } from './config'
 import { editorData } from './editor-data'
+import { getEditorVersion } from './editor-version'
+import {
+  type StorageFormat,
+  createEmptyDocument,
+  getCurrentDatetime,
+  migrate,
+  type EditorVariant,
+} from './storage-format'
 import { InstanceDataProvider } from '@/contexts/instance-context'
 import { LoggedInDataProvider } from '@/contexts/logged-in-data-context'
+import '@/assets-webkit/styles/serlo-tailwind.css'
 
 export interface SerloEditorProps {
   children: EditorProps['children']
   plugins?: (EditorPluginType | TemplatePluginType)[]
-  initialState?: EditorProps['initialState']
-  onChange?: EditorProps['onChange']
+  initialState?: unknown // Either type `StorageFormat` or outdated storage format that will be migrated to `StorageFormat`
+  onChange?: (payload: {
+    changed: boolean
+    getState: () => StorageFormat | null
+  }) => void
   language?: SupportedLanguage
+  editorVariant: EditorVariant
   _testingSecret?: string
   _ltik?: string
 }
@@ -29,7 +42,7 @@ export interface SerloEditorProps {
 export function SerloEditor(props: SerloEditorProps) {
   const {
     children,
-    initialState,
+    editorVariant,
     onChange,
     language,
     plugins,
@@ -38,6 +51,21 @@ export function SerloEditor(props: SerloEditorProps) {
   } = {
     ...defaultSerloEditorProps,
     ...props,
+  }
+
+  const initialState = !props.initialState
+    ? createEmptyDocument(editorVariant)
+    : props.initialState
+
+  const { migratedState, stateChanged } = migrate(initialState)
+
+  if (onChange && stateChanged) {
+    onChange({
+      changed: true,
+      getState: () => {
+        return migratedState
+      },
+    })
   }
 
   const { instanceData, loggedInData } = editorData[language]
@@ -53,7 +81,10 @@ export function SerloEditor(props: SerloEditorProps) {
       <LoggedInDataProvider value={loggedInData}>
         <LtikContext.Provider value={_ltik}>
           <div className="serlo-editor-hacks">
-            <Editor initialState={initialState} onChange={onChange}>
+            <Editor
+              initialState={migratedState.document}
+              onChange={handleDocumentChange}
+            >
               {children}
             </Editor>
           </div>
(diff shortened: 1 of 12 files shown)


packages/editor/src/package/editor.tsx
```diff
@@ -9,18 +10,30 @@ import { SupportedLanguage } from '@editor/types/language-data'
...
 import { InstanceDataProvider } from '@/contexts/instance-context'
 import { LoggedInDataProvider } from '@/contexts/logged-in-data-context'
+import '@/assets-webkit/styles/serlo-tailwind.css'
 
 export interface SerloEditorProps {
   children: EditorProps['children']
   plugins?: (EditorPluginType | TemplatePluginType)[]
-  initialState?: EditorProps['initialState']
-  onChange?: EditorProps['onChange']
+  initialState?: unknown // Either type `StorageFormat` or outdated storage format that will be migrated to `StorageFormat`
```
elbotho, line 31: idea: union type here.
more explicit type and we wouldn't need the comment maybe.

packages/editor/src/package/storage-format.ts
```diff
@@ -0,0 +1,133 @@
...
+    migratedState.dateModified = getCurrentDatetime()
+  }
+
+  return { migratedState, stateChanged }
+}
+
+/** The variant of the Serlo editor that created this saved data */
+const EditorVariantType = t.union([
+  t.literal('https://github.com/serlo/serlo-editor-for-edusharing'),
+  t.literal('https://github.com/serlo/serlo-editor-as-lti-tool/'),
```
elbotho: wouldn't the serlo.org editor also be a variant?

packages/editor/src/package/editor.tsx
```diff
@@ -40,6 +53,21 @@ export function SerloEditor(props: SerloEditorProps) {
     ...props,
   }
 
+  const initialState = !props.initialState
+    ? createEmptyDocument(editorVariant)
+    : props.initialState
+
+  const { migratedState, stateChanged } = migrate(initialState)
```
elbotho, line 58: this would run on every render, right?

packages/editor/src/package/storage-format.ts
```diff
@@ -0,0 +1,133 @@
...
+            {
+              plugin: 'text',
+              state: [
+                {
+                  type: 'p',
+                  children: [{ text: '' }],
+                },
+              ],
+            },
+          ],
```
elbotho: I think this might not be necessary.

packages/editor/src/package/storage-format.ts
elbotho: (ah I remember you telling me this is only inside the package right now)

elbotho (commented):
some unsolicited comments :)

packages/editor/src/package/editor.tsx
LarsTheGlidingSquirrel: We could type it to the current (and all previous) `StorageFormat` variants. But then every integration would need to do the type checking before passing it to the `SerloEditor` component. So, I figured putting the check in the editor package might be easier. What do you think?

packages/editor/src/package/storage-format.ts
LarsTheGlidingSquirrel: Yes, hopefully soon we can add `serlo.org` to this list. :)

packages/editor/src/package/editor.tsx
elbotho: Sounds reasonable 👍 Thanks for the explanation.

packages/editor/src/package/editor.tsx
LarsTheGlidingSquirrel: Yes. If prop `initialState` changed, we need to migrate again. If it didn't change, I think it probably does not hurt to just redo the migration. Should not happen often I think.

packages/editor/src/package/storage-format.ts
LarsTheGlidingSquirrel: Ah, because it will be added anyway by the rows plugin at runtime? Hm, I actually prefer this to be more explicit and already a "valid" state from the beginning.

elbotho: What I'd personally prefer about something `{plugin: EditorPluginTypes.Rows}` instead of this:
- using the enum type like we do almost everywhere else
- leaving the correct initial/empty state of the plugin in the responsibility of the plugin
- less code

But I'm also okay with this way since if the type name or default initial state would change in the future we would have to manually search for cases like this in the code anyway ^^

LarsTheGlidingSquirrel: Oh, sure. We should use the enum values. Did not see that until now. :see_no_evil: 

Hm, making the plugin responsible for the initial state sounds nice actually. You convinced me! Thank you for the explanation!

elbotho: Nice, let's hope it actually works ;)

packages/editor/src/package/editor.tsx
elbotho: I learned to never trust react not to rerender, but let's see :)

packages/editor/src/package/serlo-renderer.tsx
```diff
@@ -2,26 +2,29 @@ import { createRenderers } from '@editor/editor-integration/create-renderers'
...
 
 export function SerloRenderer(props: SerloRendererProps) {
-  const { language, _ltik } = {
+  const { language, _ltik, state } = {
     ...defaultSerloEditorProps,
     ...props,
   }
 
+  // Side note: Migrated state will not be persisted since we cannot save in static renderer view
+  const { migratedState } = migrate(state)
```
elbotho, line 26: Not sure if that is relevant for other integrations but this might increase the bundle size.

LarsTheGlidingSquirrel: Because of the `io-ts` npm package?

packages/editor/src/package/index.ts
```diff
@@ -7,6 +7,8 @@ export { defaultPlugins } from './config'
 export type { BaseEditor } from '@editor/core'
 
 export { EditorPluginType } from '@editor/types/editor-plugin-type'
+export { TemplatePluginType } from '@editor/types/template-plugin-type'
```
hejtful: Sorry if I'm repeating myself, but could you please explain why integrations will need this now?

packages/editor/src/package/index.ts
```diff
@@ -7,6 +7,8 @@ export { defaultPlugins } from './config'
...
+export { type EditorVariant } from '@editor/package/storage-format'
```
hejtful: How will this be used? Could you perhaps add an example to Readme?

packages/editor/src/package/serlo-renderer.tsx
elbotho: yeah. https://bundlephobia.com/package/io-ts is down today so I could not easily check if it actually matters.

packages/editor/src/package/storage-format.ts
```diff
@@ -0,0 +1,129 @@
...
+    document: {
+      plugin: TemplatePluginType.GenericContent,
+      state: {
+        content: {
+          plugin: EditorPluginType.Rows,
+        },
+      },
+    },
+  }
+}
```
hejtful, line 66: Would you consider extracting this function out of this file, like you did `getEditorVersion`?

packages/editor/src/package/storage-format.ts
```diff
@@ -0,0 +1,129 @@
...
+  document: t.type({
+    plugin: t.string,
+    state: t.unknown,
+  }),
+})
+export type StorageFormat = t.TypeOf<typeof StorageFormatType>
+
+export function getCurrentDatetime() {
+  return new Date().toISOString()
+}
```
hejtful: Also this one, seems like it's not strictly related to storage format?

hejtful (commented):
Two questions to understand the usage, and two super minor comments. 😁

packages/editor/src/package/serlo-renderer.tsx
LarsTheGlidingSquirrel: Hm, we can get rid of it in case the bundle size becomes too big but it is pretty handy for these things. :) Let's test this later and if it is an issue then we can find a solution.

packages/editor/src/package/index.ts
LarsTheGlidingSquirrel: Sorry, forgot to remove it. It's not needed.

packages/editor/src/package/index.ts
LarsTheGlidingSquirrel: I updated the readme. While doing so, I realized we can delete a whole bunch of other package exports because they were only necessary because we had `customPlugins`. :partying_face:

packages/editor/src/package/storage-format.ts
LarsTheGlidingSquirrel: Hm, I tried to move it but it depends on `documentType` & `currentVersion` which only make sense in context of `storage-format`. I think I would leave it there for now if that is fine for you?

packages/editor/src/package/storage-format.ts
LarsTheGlidingSquirrel: Yes! Moved it. :)

LarsTheGlidingSquirrel:
I changed `VariantType` to the less verbose `edusharing`, ... instead of `https://github.com/serlo/serlo-editor-for-edusharing`.

hejtful (approved):

PR #4050, opened 2024-08-28
feat(image-gallery): add images toolbar button (PE-74)

hejtful:
Linear ticket: PE-74

Todo:
- [x] Image and Image Gallery plugins code cleanup
- [x] Add add images button to toolbar
- [x] Show loader when adding additional images
- [x] Append new images to existing images in masonry
- [x] Limit upload to 8 in total

### DIFF
--- a/packages/editor/src/plugins/image-gallery/editor.tsx
+++ b/packages/editor/src/plugins/image-gallery/editor.tsx
@@ -63,16 +63,15 @@ export function ImageGalleryEditor(props: ImageGalleryProps) {
     dispatch(focus(id))
   }
 
-  function pruneEmptyImagePlugins() {
-    state.images.forEach((image, index) => {
-      if (selectIsDocumentEmpty(store.getState(), image.id)) {
-        state.images.remove(index)
-      }
-    })
+  function removeDefaultEmptyImagePlugin() {
+    const defaultImagePluginId = state.images[currentImageIndex].id
+    if (selectIsDocumentEmpty(store.getState(), defaultImagePluginId)) {
+      state.images.remove(currentImageIndex)
+    }
   }
 
   function handleImageModalClose() {
-    pruneEmptyImagePlugins()
+    removeDefaultEmptyImagePlugin()
     setIsModalOpen(false)
   }
 
--- a/packages/editor/src/plugins/image-gallery/editor.tsx
+++ b/packages/editor/src/plugins/image-gallery/editor.tsx
@@ -6,7 +6,7 @@ import {
   useAppSelector,
 } from '@editor/store'
 import { EditorPluginType } from '@editor/types/editor-plugin-type'
-import { MouseEvent, useEffect, useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import type { ImageGalleryProps } from '.'
 import { AddImagesButton } from './components/add-images-button'
@@ -34,9 +34,7 @@ export function ImageGalleryEditor(props: ImageGalleryProps) {
   )
 
   function handleAddImagesButtonClick() {
-    if (state.images.length === 0) {
-      state.images.insert(0, { plugin: EditorPluginType.Image })
-    }
+    state.images.insert(state.images.length, { plugin: EditorPluginType.Image })
     setIsModalOpen(true)
   }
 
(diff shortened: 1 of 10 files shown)


packages/editor/src/plugins/image-gallery/editor.tsx
```diff
@@ -34,31 +39,44 @@ export function ImageGalleryEditor(props: ImageGalleryProps) {
...
   }
 
   function handleMultipleImageUpload(files: File[]) {
-    for (const file of files) {
-      const newImagePluginState = imagePlugin.onFiles?.([file])
-      const newImagePlugin = {
-        plugin: EditorPluginType.Image,
-        state: newImagePluginState?.state as unknown,
+    if (state.images.length + files.length > MAX_IMAGES) {
+      showToastNotice(imageGalleryStrings.tooManyImagesMessage, 'warning')
```
elbotho, line 49: You could use %max_images% in the i18n string and replace it here with the constant.  Not sure how likely that is to change though.

elbotho: maybe an early return would make sense here? Not sure we actually want to close the modal after that error.

packages/editor/src/plugins/image-gallery/editor.tsx
```diff
@@ -34,31 +39,44 @@ export function ImageGalleryEditor(props: ImageGalleryProps) {
...
       }
-      state.images.insert(currentImageIndex, newImagePlugin)
     }
+
     setHasImages(true)
     setIsModalOpen(false)
     dispatch(focus(id))
   }
 
+  function pruneEmptyImagePlugins() {
```
elbotho: when/how do empty images happen?

elbotho: maybe we should not init the plugin with one empty image plugin in the beginning, is that the problem?

packages/editor/src/plugins/image-gallery/editor.tsx
elbotho: Because `handleMultipleImageUpload` is run after the normal upload the first image is uploaded anyway. 
This should not be a problem but it's a bit unexpected. We could also instead upload all images up to 8 and change the error message to say that not all images were uploaded.

maybe a decision for @GregorZupan?

elbotho (commented):
looking good so far, commented on two things.

packages/editor/src/plugins/image-gallery/editor.tsx
hejtful: @GregorZupan In short, how do we handle the situation when the user uploads more than 8 images in total? 😁

packages/editor/src/plugins/image-gallery/editor.tsx
hejtful: In order to show the empty Image plugin in the modal, we need to have an empty Image plugin in the state.

When the user closes the modal without uploading anything, we get rid of that plugin using this function. I wouldn't say there's a problem at all, it's just how it works.

Do you see a better flow for this?

elbotho: Nice, thanks for the explanation. Maybe we can call it removeDefaultImagePlugin or something then?

hejtful: Good point, I also changed the function itself to match the name, we don't really need to iterate. Thanks!

elbotho (approved):

packages/editor/src/plugins/image-gallery/editor.tsx
GregorZupan: I would send a gentle reminder that there are only 8 images allowed:

Copy for when the User tries to Upload more than 8 images at once:

English:
`You can only upload 8 images in this element. Please select fewer images and try again.
`
German:
`Du kannst maximal 8 Bilder in der Galerie hochladen. Bitte wähle weniger Bilder aus und versuche es erneut.`



When 8 images have been uploaded and the user tries to upload more images via the add image button:

English
`Maximum of 8 images reached. PLease remove one or more images to upload new ones`

German
`Maximal 8 Bilder erlaubt. Bitte entferne ein oder mehrere Bilder, um neue hochzuladen.`


I would recommend to implement a _Toast Message_ for this case (less intrusive than a popup). Do you need a Design for that?

elbotho: Thank you! 

>I would recommend to implement a Toast Message for this case (less intrusive than a popup). Do you need a Design for that?

No thats fine. Vito built it already 🚀

PR #4052, opened 2024-08-28
feat(image-gallery): single image hover state + delete image (PE-59)

hejtful:
Linear issue: PE-59

To do:
- [x] Hover state backround gradient
- [x] Show image caption on hover
- [x] Show remove image button on hover
- [x] Remove image from gallery on remove image button click


packages/editor/src/plugins/image-gallery/components/image-grid.tsx
```diff
@@ -17,42 +34,67 @@ export function ImageGrid({ images, onImageClick }: ImageGridProps) {
...
+              'pointer-events-auto absolute right-4 top-4 h-8 w-8 rounded-full p-1',
+              'bg-black bg-opacity-20 text-white hover:bg-opacity-50'
+            )}
+            onClick={(event) => {
+              event.stopPropagation()
+              onRemoveImageButtonClick(index)
+            }}
+          >
+            <FaIcon icon={faTrashCan} className="text-sm" />
+          </button>
```
elbotho: So far `ImageGrid` is used like a renderer so this is not 100% the place for editor ui.
But this is very small and the ways to inject custom ui in the image grid are also not super elegant so we could also keep it here.

packages/editor/src/plugins/image-gallery/components/image-grid.tsx
```diff
@@ -1,58 +1,95 @@
...
+          </div>
         )
       })}
     </div>
   )
+
+  function renderHoverOverlay(caption: Descendant, index: number) {
+    return (
+      <div className={hoverOverlayClassNames}>
+        <StaticSlate element={caption} />
```
elbotho: The caption can include links. They have bad contrast (@GregorZupan) and they are not clickable right now (@hejtful)
image

elbotho: (code formatting and math formulas work okay)

elbotho (commented):
I left some more UI suggestions in linear for Gregor, some smallish code comments.

packages/editor/src/plugins/image-gallery/components/image-grid.tsx
hejtful: Good point, I will do the same as you did with extraChildren then, to be consistent 👍

packages/editor/src/plugins/image-gallery/components/image-grid.tsx
hejtful: ~~I'll fix~~ Fixed the clickability

elbotho: Something like this might be enough:
```suggestion
    <div className="pointer-events-auto [&_a]:text-brand-400">
        <StaticSlate element={caption} />
    </div>
```

hejtful: Also added the color suggestion. @GregorZupan this is what it looks now, good enough?

Screenshot 2024-08-29 at 16 30 05

elbotho (approved):
👍 thank you!

PR #4054, opened 2024-08-29
refactor(plugin-gallery): simpler image grid

elbotho:
@hejtful this simplifies the js part of the image grid by using rows. 
also integrates the single image case into the normal logic to make it a bit more compact.


hejtful (approved):
Neat

PR #4060, opened 2024-09-02
feat(image-gallery): static carousel (mobile)

hejtful:
Linear issue: PE-56


elbotho (commented):
Nice, works nicely (at least on simulated mobile).

Did you consider CSS scroll snap? That might simplify the code (less or no js event stuff).
- https://caniuse.com/?search=CSS%20scroll%20snap
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll_snap
- https://css-tricks.com/practical-css-scroll-snapping/

PR #4062, opened 2024-09-03
feat(plugin-rows): split BlanksExercise in plugin menu

hejtful:
Linear issue: PE-69

A new plugin type is introduced, `EditorPluginType.BlanksExerciseDragAndDrop`. But, it still uses the existing components. Additionally, BlanksExercise is now initialized as a function that accepts the config as an argument, so that the `defaultMode` could be provided programmatically.

It's a lot of changes, but I didn't think of a way to do it with fewer changes. Feedback very welcome


apps/web/src/data/en/index.ts
```diff
@@ -1096,6 +1101,7 @@ export const loggedInData = {
           textAreaExercise: 'Text Box Exercise',
           dropzoneImage: 'Image Dropzones Exercise',
           blanksExercise: 'Fill In The Blanks Exercise',
+          blanksExerciseDragAndDrop: 'Fill In The Blanks Exercise',
```
elbotho: This should be a different string, right?

apps/web/src/data/en/index.ts
```diff
@@ -1043,6 +1043,11 @@ export const loggedInData = {
           alternativeAnswers: 'Alternative answers',
           acceptMathEquivalents: 'Accept all equivalent mathematical values',
         },
+        blanksExerciseDragAndDrop: {
+          title: 'Drag&Drop exercise',
```
elbotho: Is that a good title since we also have other exercises that use drag&drop?
Maybe "Drag & Drop Fill in the blanks"? 
(maybe ping @GregorZupan)

apps/web/src/data/en/index.ts
```diff
@@ -1043,6 +1043,11 @@ export const loggedInData = {
...
+          description:
+            'Create a fill-in-the-gap text or a table where the correct answers are inserted via drag and drop.',
```
elbotho:
```suggestion
            'Create a fill in the blanks exercise (text or table) where learners add the answers via drag and drop.',
```

btw: the "fill-in-the-gap" part is also incorrect in the original string above.

elbotho (changes requested):

elbotho:
A little off topic:
@GregorZupan This view is not really working any more, but it's also not really supported any more, right?
image

(currently you can for example remove an input exercise from the plugin toolbar and get this view)

Maybe we should remove it and add an "Textaufgabe" Option in the Exercise Plugin Toolbar instead?
(then this select menu would be the only way to change an exercise)

image

apps/web/src/data/en/index.ts
hejtful: The strings come from [this document](https://docs.google.com/document/d/1Jqd3qqRtBX-J3dKwC7-rohDqZGTSGyNNvPq5XXaKllY/edit#heading=h.ubjbe6nuxwf2) and were discussed with @GregorZupan , so I'll leave this to him.

apps/web/src/data/en/index.ts
hejtful: Good point, thanks! I'll wait for [this comment](https://github.com/serlo/frontend/pull/4062/files#r1743439889) to be resolved, and then fix this.

apps/web/src/data/en/index.ts
GregorZupan: Good Point! Do we need both the english and german strings for the plugin(s)?

apps/web/src/data/en/index.ts
GregorZupan: Proposal accepted! 

I'll edit it in my tooltip document

GregorZupan: I also changed the german copy a bit

apps/web/src/data/en/index.ts
elbotho: Having both would be easiest.

apps/web/src/data/en/index.ts
```diff
@@ -1025,9 +1025,9 @@ export const loggedInData = {
           description: 'Import an interactive task from H5P via URL.',
         },
         blanksExercise: {
-          title: 'Fill In The Blanks',
+          title: 'Fill in the Blanks',
```
elbotho, line 1028:
```suggestion
          title: 'Fill in the Blanks (typing)',
```

elbotho (approved):
:godmode:

PR #4063, opened 2024-09-03
chore(editor): remove unnecessary hotkeys scope

hejtful:


CodingDive (approved):
Thank you! Seems to work great https://frontend-git-chore-remove-unnecessary-hotkeys-scope-serlo.vercel.app/entity/repository/add-revision/277232. Let's see what the e2e tests think about it :grin: :crossed_fingers:

packages/editor/src/core/editor.tsx
```diff
@@ -13,9 +13,7 @@ export function Editor(props: EditorProps) {
   return (
     <Provider store={store}>
       <DndWrapper>
-        <HotkeysProvider
-          initiallyActiveScopes={['global', 'root-up-down-enter']}
-        >
+        <HotkeysProvider initiallyActiveScopes={['global']}>
```
CodingDive, line 16: Because we now have only one scope, we could also remove it entirely. However, leaving the scope may make it easier to add another scope and hotkey in the future

hejtful: Since you don't seem to have a preference, I'll leave it like this 😁

PR #4066, opened 2024-09-04
Fix editor web component new storage format

CodingDive:
Next up is releasing a new editor-web-component version


CodingDive:
Eslint: 

image


Chicken egg problem I don't know if I can fix these lint errors without releasing a new version of the editor first :grin:

kulla (changes requested):
- [x] add editor variant "unknown"
- [x] Add editor variants "scobees" | "kiron"
- [x] save domain

packages/editor/src/package/storage-format.ts
```diff
@@ -9,10 +9,34 @@ import { getEditorVersion } from './editor-version'
...
+        domainOrigin: window.location.origin,
+        editorVersion: getEditorVersion(),
+        dateModified: getCurrentDatetime(),
+        document: state,
+      }
+    }
+
+    // State is neither in the new format nor the old format, throw!
+    throw new Error('Invalid state format')
+  },
```
kulla: This shouldn't be a migration

kulla (changes requested):

packages/editor/src/package/storage-format.ts
```diff
@@ -9,10 +9,34 @@ import { getEditorVersion } from './editor-version'
...
+  // Migration 1: Add `editorVersion` and `id`
```
LarsTheGlidingSquirrel: ... and `domainOrigin`

packages/editor/src/package/storage-format.ts
```diff
@@ -35,6 +59,7 @@ const migrations: Migration[] = [
     return {
       ...state,
       editorVersion,
+      domainOrigin: window.location.origin,
```
LarsTheGlidingSquirrel, line 40: If the package is already deployed to production (could be the case in RLP, not sure) we should not change existing migrations. But instead we can create a new migration at the end of the `migrations` array.

packages/editor/src/package/storage-format.ts
```diff
@@ -65,25 +91,47 @@ export function createEmptyDocument(
...
+  const nextMigrationIndex = migratedState.version
 
-  // Create deep copy
-  let migratedState = JSON.parse(JSON.stringify(stateBeforeMigration)) as {
-    version: number
-  }
-  let stateChanged = false
   for (let i = nextMigrationIndex; i < migrations.length; i++) {
-    migratedState = migrations[i](migratedState)
+    migratedState = migrations[i](migratedState, variant) as StorageFormat
```
LarsTheGlidingSquirrel: I would remove the `as StorageFormat` because it will have this runtime type only after the last migration ran not in between.

packages/editor/src/package/storage-format.ts
```diff
@@ -65,25 +91,47 @@ export function createEmptyDocument(
...
+    // deep copy
+    migratedState = JSON.parse(
+      JSON.stringify(stateBeforeMigration)
+    ) as StorageFormat
+  } else {
     throw new Error(
-      `Missing property 'version: number' in state while trying to perform migrations. Got ${JSON.stringify(stateBeforeMigration)}`
+      `Unknown, invalid state format. Got ${JSON.stringify(stateBeforeMigration)}`
     )
+  }
```
LarsTheGlidingSquirrel: This will throw the error if we get a state that is neither the (old) document state nor the newest StorageFormat state, so all possible states in between.

LarsTheGlidingSquirrel: Maybe something like: 
```ts
if (isOldDocumentFormat(stateBeforeMigration)) {
  migratedState = {
    id: uuid_v4(),
    type: documentType,
    variant,
    domainOrigin: window.location.origin,
    version: 0,
    editorVersion: getEditorVersion(),
    dateModified: getCurrentDatetime(),
    document: JSON.parse(JSON.stringify(stateBeforeMigration)),
  }
  stateChanged = true
} else {
  migratedState = JSON.parse(JSON.stringify(stateBeforeMigration))
}
```

LarsTheGlidingSquirrel (changes requested):

packages/editor/src/package/storage-format.ts
CodingDive: I prefer it here over having a dedicated migration. If we see some content that won't have a domainOrigin, I'm happy to add an extra migration, but would try to keep it lean for now.

packages/editor/src/package/storage-format.ts
CodingDive: Thank you! It failed with the state you sent me. Fixed now

packages/editor/src/package/storage-format.ts
```diff
@@ -106,24 +131,42 @@ const EditorVariantType = t.union([
...
+const DocumentType = t.type({
+  plugin: t.string,
+  state: t.unknown,
 })
+
+type Document = t.TypeOf<typeof DocumentType>
+
+function isValidDocument(obj: unknown): obj is Document {
+  return DocumentType.is(obj)
+}
```
kulla: Unecessary. Use `DocumentType.is(obj)` instead since it is already a type guard

kulla (approved):
LGTM

packages/editor/src/package/storage-format.ts
LarsTheGlidingSquirrel: Hm, I prefer a separate migration. See https://github.com/serlo/frontend/pull/4070

packages/editor/src/package/storage-format.ts
```diff
@@ -65,30 +69,41 @@ export function createEmptyDocument(
...
-  // Create deep copy
-  let migratedState = JSON.parse(JSON.stringify(stateBeforeMigration)) as {
-    version: number
+  // Check if the state is the (old format)
+  if (
+    DocumentType.is(stateBeforeMigration) &&
+    !StorageFormat.is(stateBeforeMigration)
+  ) {
+    migratedState = {
+      ...createEmptyDocument(variant),
```
LarsTheGlidingSquirrel, line 88: This would mix up the state types: On one hand we have all the properties of `version: 1` but on the other hand we set `version: 0`.

PR #4070, opened 2024-09-05
feat(editor-package): Add migration and some code changes in `storage-format.ts`

LarsTheGlidingSquirrel:
# Changes
- Separate migration for `domainOrigin` & made it mandatory. Reason: Before, we changed an existing migration and added it optionally. But this might cause this property to exist in some migrated content but not in others. I feel like it is easier for the devs if we can look at the migration code and know relatively quickly what variants of types can be out there. So, I prefer mandatory properties and making sure that after all migrations ran, they are there for sure. 
- Added code comments and did some renaming to hopefully improve clarity

# TODO 
- [x] Test

### DIFF
--- a/packages/editor/src/package/editor.tsx
+++ b/packages/editor/src/package/editor.tsx
@@ -15,7 +15,7 @@ import { defaultSerloEditorProps } from './config'
 import { editorData } from './editor-data'
 import { getEditorVersion } from './editor-version'
 import {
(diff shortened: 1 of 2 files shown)


CodingDive (approved):
Great job! Definitely slightly more verboten, but very clear and I love the many comments you added.
Tested it in the Web component with no problems whatsoever.

PR #4071, opened 2024-09-05
refactor(editor): rework learner event handling

elbotho:
## intro
This PR moves the serlo.org specific learners event handling into the editor core (since we will need that in other integration at some point) and cleans up the implementation.

---

## trigger events from plugins

Each plugin can now trigger relatively basic events that are based on `xAPI`-verbs but use only editor specific context.
Example: 
```ts
editorLearnerEvent.trigger?.({
   verb: 'answered',
   correct,
   contentType: 'blanks-exercise',
})
```

## handle events from outside the editor

Each integration can then provide a `triggerLearnerEvent` function that decides what to do with that event (e.g. add user- and context information and send an xAPI event to an LRS).

The current implementation only logs the results to the console:
```ts
function triggerLearnerEvent(data: LearnerEvent) {
  // eslint-disable-next-line no-console
  console.log(data)

  const { correct, contentType } = data

  // eslint-disable-next-line no-console
  console.log({
    path: asPath,
    entityId: exerciseTrackingId,
    revisionId,
    result: correct ? 'correct' : 'wrong',
    type: contentType,
  })
}
```

## important changes

- this removes some (I think unused) a/b logic
- this would remove the current reporting to the api (it's my understanding it's not used currently)
- this removes some custom serlo rendering wrappers and simplifies others


kulla (approved):
Do I understand it correctly that at the moment the Trigger would be set by a provider, right? In the future I would set it via editor-api. Anyways this is a good implementation right now

elbotho:
> Do I understand it correctly that at the moment the Trigger would be set by a provider, right? In the future I would set it via editor-api. Anyways this is a good implementation right now

Thanks for the feedback. Not sure I fully understand your question.

The function `triggerLearnerEvent` is provided before the editor just like we do with `createPlugins` and renderers.
See `editorLearnerEvent.init(triggerLearnerEvent)` e.g. in `___editor_preview`.

In my book that means it's part of the editor API. Not sure what you mean by provider.

kulla:
> In my book that means it's part of the editor API. Not sure what you mean by provider.



> > Do I understand it correctly that at the moment the Trigger would be set by a provider, right? In the future I would set it via editor-api. Anyways this is a good implementation right now
> 
> Thanks for the feedback. Not sure I fully understand your question.
> 
> The function `triggerLearnerEvent` is provided before the editor just like we do with `createPlugins` and renderers. See `editorLearnerEvent.init(triggerLearnerEvent)` e.g. in `___editor_preview`.
> 
> In my book that means it's part of the editor API. Not sure what you mean by provider.



> > Do I understand it correctly that at the moment the Trigger would be set by a provider, right? In the future I would set it via editor-api. Anyways this is a good implementation right now
> 
> Thanks for the feedback. Not sure I fully understand your question.
> 
> The function `triggerLearnerEvent` is provided before the editor just like we do with `createPlugins` and renderers. See `editorLearnerEvent.init(triggerLearnerEvent)` e.g. in `___editor_preview`.
> 
> In my book that means it's part of the editor API. Not sure what you mean by provider.

This is only possible in frontend but not in the integrations where the `@serlo/editor` NPM package is used like https://github.com/serlo/serlo-editor-as-lti-tool => I mean to add the function to https://github.com/serlo/frontend/blob/2406b3bef322dfe34d99cd3b66cc8d0158363d1b/packages/editor/src/package/editor.tsx#L28-L37

However since no integration can read xAPI we can omit this for right now...

PR #4077, opened 2024-09-09
Feat/expose all plugins in editor

CodingDive:


packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,576 @@
...
+import IconTextArea from '@editor/editor-ui/assets/plugin-icons/icon-input-exercise.svg?raw'
+import IconMultimedia from '@editor/editor-ui/assets/plugin-icons/icon-multimedia.svg?raw'
+import IconSpoiler from '@editor/editor-ui/assets/plugin-icons/icon-spoiler.svg?raw'
+import IconTable from '@editor/editor-ui/assets/plugin-icons/icon-table.svg?raw'
+import IconText from '@editor/editor-ui/assets/plugin-icons/icon-text.svg?raw'
+import IconVideo from '@editor/editor-ui/assets/plugin-icons/icon-video.svg?raw'
+import { EditorPluginType } from '@editor/types/editor-plugin-type'
+
+import { Instance } from '@/fetcher/graphql-types/operations'
+import { getLoggedInData } from '@/helper/feature-i18n'
```
kulla: Please avoid those imports => those are next-js imports and thus bad

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,576 @@
...
+    description,
+  }
+}
+
+const getInternationalizedPluginStrings = (type: EditorPluginType) => ({
+  de: getPluginNameAndDescription('de', type),
+  en: getPluginNameAndDescription('en', type),
+})
+
+export const AllPlugins = [
```
kulla:
```suggestion
export const pluginDescriptions = [
```

kulla: Can you do `typeof allPlugins = Record<PluginType, PluginTypeDescription>` => So a dict instead of a list. Can you also export PluginType as a enum?

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,576 @@
...
+            illustrating: true,
+            width: 50,
+          },
+        },
+      ],
+    },
+  },
+  {
+    ...getInternationalizedPluginStrings(EditorPluginType.SerloTable),
+    icon: IconTable,
```
kulla:
```suggestion
    icon: {
       svgSource: string,
       Component: React.Node
    }
```

What do you think about this?

kulla (changes requested):

packages/editor/src/core/all-plugins.tsx
CodingDive: EditorPluginType is already exported

packages/editor/src/core/all-plugins.tsx
CodingDive: Possibly twice the bundle size, but I'm open to it.

CodingDive: I'd say for the React package, we can find a better API. Something like this

```TSX
import React from 'react';
import { EditorPluginType } from '@editor/types/editor-plugin-type';
import * as Icons from './icons'; 

interface IconProps {
  pluginType: EditorPluginType;
  className?: string;
}

export const PluginIcon: React.FC<IconProps> = ({ pluginType, className }) => {
  const SvgIcon = Icons[pluginType as keyof typeof Icons];
  
  if (!SvgIcon) {
    console.warn(`No icon found for plugin type: ${pluginType}`);
    return null;
  }

  return <SvgIcon className={className} />;
};
```

It's a little bit sad, that we have to export the svg sources, but in the future, we can export them exclusively in the editor-web-component. For now, I think we don't have to worry about the bundle size too much and whoever is using the React package, their code bundler should strip out all the unused svgs anyways.

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,576 @@
...
+          },
+        },
+      ],
+    },
+  },
+  [EditorPluginType.TextAreaExercise]: {
+    ...getInternationalizedPluginStrings(EditorPluginType.TextAreaExercise),
+    icon: IconTextArea,
+    type: EditorPluginType.TextAreaExercise,
+    initialState: { plugin: 'textAreaExercise', state: { feedback: '' } },
```
kulla: this should be inside an exercise plugin.

kulla (changes requested):

packages/editor/src/core/all-plugins.tsx
CodingDive: Nice catch, thank you! :heart: I built a select in the editor-lit and will probably find more small errors in the initial states (or plugins that are not enabled in the integrations)

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,612 @@
...
+    description,
+  }
+}
+
+const getInternationalizedPluginStrings = (type: EditorPluginType) => ({
+  de: getPluginNameAndDescription('de', type),
+  en: getPluginNameAndDescription('en', type),
+})
+
+export const AllPlugins = {
```
elbotho: I had the same discussion a few days ago with Lars. 
If we make the editor states explicit here we have to make sure we keep them up to date in this file. 
If we initialise a box plugin with `{plugin: EditorPluginType.Box}` instead of the complete explicit initialState the responsibility to create a valid initialState stays inside the plugin.
Imho this is easier to maintain because it keeps up to date with changes in the plugin.

To be honest I still don't fully understand the use case so maybe this would not work here, but that's what I would prefer if possible.

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,576 @@
...
+import IconTextArea from '@editor/editor-ui/assets/plugin-icons/icon-input-exercise.svg?raw'
+import IconMultimedia from '@editor/editor-ui/assets/plugin-icons/icon-multimedia.svg?raw'
+import IconSpoiler from '@editor/editor-ui/assets/plugin-icons/icon-spoiler.svg?raw'
+import IconTable from '@editor/editor-ui/assets/plugin-icons/icon-table.svg?raw'
+import IconText from '@editor/editor-ui/assets/plugin-icons/icon-text.svg?raw'
+import IconVideo from '@editor/editor-ui/assets/plugin-icons/icon-video.svg?raw'
+import { EditorPluginType } from '@editor/types/editor-plugin-type'
+
+import { loggedInData as loggedInDataDe } from '@/data/de'
+import { loggedInData as loggedInDataEn } from '@/data/en'
```
kulla: There are some plugins which lack a description (see `rows`). Can you add one?

kulla (changes requested):

packages/editor/src/core/all-plugins.tsx
CodingDive: Good point. However, it'll still need to be wrapped in a row, exercise or similar. Maybe I can create some factory functions and then use the initial state of the respective plugins. Also if we give it v1, or leave it as is in the old format (over the new storage-format), the migration algorithm should technically take care of it, even if we don't want to update it here

packages/editor/src/core/all-plugins.tsx
CodingDive: Rows is not a plugin that gets exposed.

CodingDive: All the plugins that I expose, have a description

packages/editor/src/core/all-plugins.tsx
kulla: @elbotho We have two issues to solve:

* Some plugins like `blanksExercise` need to be wrapped inside a `exercise` plugin.
* Some plugins like `blanksExercise` are two plugins we want to have in external menus to include the plugin, like one with mode `typing` (fill-in-gaps exercise) and one with the mode `drag-and-drop`) (drag-and-drop words exercise).

Is the a simpler solution for this?

packages/editor/src/core/all-plugins.tsx
```diff
@@ -0,0 +1,612 @@
...
+                  plugin: 'text',
+                  state: [{ type: 'p', children: [{ text: '' }] }],
+                },
+              ],
+            },
+          },
+        },
+      ],
+    },
+  },
```
elbotho:
```suggestion
    initialState: {
      plugin: EditorPluginType.Rows,
      state: [{ plugin: EditorPluginType.Box }]
  },
```

elbotho: 1. use `EditorPluginType`
2. let editor fill in initialState where possible

packages/editor/src/core/all-plugins.tsx
elbotho: @kulla your custom cases make sense. I made an example suggestion for all other plugins (most of them I think) [here](https://github.com/serlo/frontend/pull/4077/files#r1750605297).

packages/editor/src/core/all-plugins.tsx
elbotho: If that works I'd prefer this.

packages/editor/src/core/index.ts
```diff
@@ -1,6 +1,7 @@
 export { SubDocument } from './sub-document'
 export type { SubDocumentProps } from './sub-document'
 export { Editor } from './editor'
+export { AllPlugins } from './all-plugins'
```
kulla: remove this

packages/editor/src/core/all-plugins.tsx
CodingDive: Yeah, that worked great! Unfortunately, I couldn't get rid of wrapping it within a rows plugin, because the migration algorithm and possibly some other places in the editor expect this `plugin, state` structure.

elbotho: does that mean we would get the prominent add-button under every plugin now?

CodingDive:
image

:face_exhaling: 

@kulla @elbotho please check again :blush: 

@hejtful here is an example of how it can be used. I also added some docs https://github.com/serlo/serlo-editor-lit/pull/1

packages/editor/src/core/all-plugins.tsx
CodingDive: Yes, for now. 

This is something that most integrations actually want, e.g Moodle. But some others may want to have one plugin and one plugin only

elbotho: It's my understanding that those who want the default behaviour won't be using the helpers in this PR.
This means the requirements of FWU are not solved, correct?

CodingDive: > This means the requirements of FWU are not solved, correct?

Correct. Not all requirements.

I think we should solve the ability to block adding a new row not in this PR, as we can't solve it for all plugins like exercises anyways. For now, we have a nice export on how to build plugin menus, get initial states and icons. The package consumers can already hide the plus button with targeted css. 

If we have the need, let's think about a good API for hiding the `+` button in a follow up PR.

Edit: I'm exploring the css solution now.

CodingDive: > It's my understanding that those who want the default behaviour won't be using the helpers in this PR.

I think Moodle will be happy to use the code in this PR and simplify their current logic. From talking to Vicky, they'd always want the ability to add some extra text, or other plugins after a table, box, image, etc.

CodingDive: I added a data attribute to the plus button now.

Can be disabled with the following css snippet 

```css
[data-button-type="add-row"] {
    display: none;
}
```

packages/editor/src/plugins/rows/components/add-row-button-floating.tsx
```diff
@@ -37,6 +37,7 @@ export function AddRowButtonFloating(props: AddRowButtonFloatingProps) {
       ref={buttonRef}
       className={cn(baseStyles, stateStyles, interactionStyles)}
       onClick={onClick}
+      data-button-type="add-row"
```
elbotho, line 42: maybe a comment so it does not get deleted at some point?

elbotho (approved):
One more comment.
Nice that you could solve the bundle size!

packages/editor/src/plugins/rows/components/add-row-button-floating.tsx
CodingDive, line 42:
```suggestion
{/* needed so that the "add plugin" row can be hidden in CSS for editor integrations */}
data-button-type="add-row"
```

PR #4086, opened 2024-09-11
feat(editor-package): Enable video and h5p plugin

CodingDive:
Weirdly, in the integrations the free text area plugin does not work yet, despite being in the list of create-basic-plugins. If anyone has any ideas on why that could be, I'm all ears!


elbotho:
I can be convinced otherwise but I'd be surprised if the H5P plugin (developed more as a quick win for internal serlo use) works without adaptations in other integrations.
Did you test/check that?

kulla:
@CodingDive Is this the place to enable the plugins in all integrations?

CodingDive:
> @CodingDive Is this the place to enable the plugins in all integrations?

I'm trying to figure this out. These plugins are already enabled in the renderer, but not in the editor :thinking: 

Previously, we had a config that was removed in my vacation and now the plugins are not working. Even with this PR, locally I run into the following problem

image

kulla:
> Previously, we had a config that was removed in my vacation and now the plugins are not working. Even with this PR, locally I run into the following problem

How can I reproduce the bug?

elbotho:
Closing for a cleaner PR overview. Please reopen if you continue work here.

PR #4087, opened 2024-09-11
refactor(editor)!: Distinguish between plugins and addable elements

kulla:
IMHO we have introduced technical debt in https://github.com/serlo/frontend/pull/4077: As stated yesterday in a meeting there is a distinction between a technical plugin and an element which can be added via our menu in the rows plugin for adding a new element or any other menu in a block editor doing the same:

* Technically we have one plugin `blanksExercise`. This can be a fill in the gap exercise with `mode === "typing"` and a drag and drop exercise with `mode === "drag-and-drop". The same goes with single choice / multiple choice which are two elements in the "add new element" menu in the rows plugin but is technically one plugin.
* When any interactive plugin is added we add a "exercise" plugin with a predefined configuration.

https://github.com/serlo/frontend/pull/4077 does not reflect this. This PR introduces (1) a new naming and (2) introduces a new enum separate from the list of technical plugins for the elements which can be included via the rows plugin.

### ToDos

- [x] Fix the bug with the rows plugin (it is definitely a bug when a rows plugin is needed as root element)
- [x] Update documentation


kulla:
@CodingDive ready for merge and deployment (see slack for info) Can you double check that everything is alright and publish the editor packages?

CodingDive:
Sure, will do!

packages/editor/src/package/index.ts
```diff
@@ -2,7 +2,16 @@ export { SerloEditor, type SerloEditorProps } from './editor'
...
+
+// We need to make a distinction between entires on our menu and technical
+// plugin types Internally we have for example a `scMcExercise` which has a
+// configuration whether it is a single choice or multiple choice exercise.
+// However on the menu there are the two entries `singleChoiceExercise` and
+// `multipleChoiceExercise`. When somebody else uses our editor as a block
+// inside their own editor they would like to have the two entries
+// `singleChoiceExercise` and `multipleChoiceExercise`.
+// Thus we export our menu entries here as plugin.
+export { PluginMenuItem as Plugin, pluginMenu as plugins } from './plugin-menu'
```
CodingDive, line 14: Love the naming! :heart: We have a beautiful api now :heart_eyes: Thank you for all the work!

CodingDive (approved):

PR #4089, opened 2024-09-11
fix(editor): Fix migration of documents of format `{ plugin: string }`

kulla:
Fixes https://github.com/serlo/frontend/pull/4077#discussion_r1752375896

I created a separate PR so that this can be more easily be reviewed...


kulla:
@CodingDive Can you test that the removal of `__EDITOR_VERSION__` works in bundling the editor?

elbotho:
Am I reading this right that migrations and everything else of a state without a surrounding rows plugin actually works and this PR just adds a test and removes the rows plugin in Mikeys code?

CodingDive (approved):
Fixed the built! Works great :tada: 

You can try it here https://github.com/serlo/serlo-editor-lit/pull/2

```
cd packages/editor && yarn yalc:publish
cd ../editor-web-component && yalc add @serlo/editor && yarn yalc:publish
```

Then in the repo 

```
yarn yalc add @serlo/editor-web-component && pnpm start
```

kulla:
> Am I reading this right that migrations and everything else of a state without a surrounding rows plugin actually works and this PR just adds a test and removes the rows plugin in Mikeys code?

A little change in `packages/editor/src/package/storage-format.ts` is needed => in the decoder `state` needed to be present so that having `initialState={{ plugin: "..." }}` would throw an error. But yes, next to this fix this PR adds the possibility for tests which we can use in the future and simplifies the setting of the editor version (`__EDITOR_VERSION__` does not need to be replaced at build time)

elbotho (approved):

PR #4099, opened 2024-09-13
wip: migrate() should accept newer versions of a document

kulla:
See https://linear.app/serlo/issue/PE-151/bug-migration-algorithm-does-not-accept-newer-versions


LarsTheGlidingSquirrel:
Hm, right now function `migrate` ensures it will return type `StorageFormat`. Changing it like suggested would mean it can also return other (unknown) types. 

The two use cases where I could see this happen: 
- Editor version is downgraded but some content in the newer version was already created
- Copy & paste content across platforms where the editor versions do not match

I am not sure it we want to be more strict or loose with the type. But I am leaning more towards being strict and failing early.

elbotho:
keeping branch as reference for the issue but closing PR

PR #4107, opened 2024-09-15
WIP: Add new media plugin to simplify embeddings of external resources

kulla:
Fixes https://linear.app/serlo/issue/PE-149/implement-the-proposal-for-a-new-asset-management-in-an-example-pr

There are some changes:

* I didn't use schema.org to describe the embedding types
* This PR does not remove `image` and `video` types which can in the long run be replaced with the new media plugin (and a config for the allowed plugin) => see the plugins `mediaImage` and `mediaVideo` instead.


packages/editor/src/plugins/media/services/types.ts
elbotho, line 1: Thank you for the comments, they actually help a lot here.
While I'm in awe of you typescript skills as always, I'd prefer stupid manual & explicit types here to be honest.

Why? Because it's easier to understand and we don't plan on adding new media types all the time.
E.g. when working with the types (e.g. `Resource`) it's so much easier to see the actual type instead of 
```
(alias) type Resource<H extends Hosting = Hosting> = { [T in H]: ResourceAdditonalInformation[T] & {
    hostingService: T;
}; }[H]
import Resource
```
+ a comment.

I'm probably missing reasons where we need the type functions but I think we should consider static types where possible.

packages/editor/src/plugins/media/services/html-video/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+          resource: {
+            hostingService: Hosting.CDN,
+            embeddingType: Embed.HTMLVideo,
+            contentUrl: url.href,
+          },
+        })
+      }
+
+      video.onerror = () => {
+        if (videoExtensions.some((ext) => url.pathname.endsWith('.' + ext))) {
```
elbotho, line 32: should we not check that first, before even trying to load?

elbotho:
some todos:
- i18n
- alt/description field
- renders
- EmbedWrapper
- search interface
- …

elbotho:
@kulla how would we implement a video plugin that supports html-video & e.g. YouTube and Vimeo?

elbotho:
Did you have a way in mind to handle the custom state that e.g. the image plugin needs (caption, alt, source, license) with this approach?

packages/editor/src/plugins/media/services/types.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Generic type which creates given the list of embed Identifiers the
+ * union of its embedding types. For example Embedding<Embed.GeoGebraApplet> is
+ * equivalent to:
+ *
+ * interface HTMLImageEmbedding {
+ *   type: Embed.HTMLImage,
+ *   appletId: GeoGebraApplet
```
LarsTheGlidingSquirrel:
```suggestion
 * interface GeoGebraAppletEmbedding {
 *   type: Embed.GeoGebraApplet,
 *   appletId: GeoGebraApplet
```

kulla:
> @kulla how would we implement a video plugin that supports html-video & e.g. YouTube and Vimeo?

See https://github.com/serlo/frontend/pull/4107/commits/a2d2230ad977a0f400fc7aad129b6ac5e828188f

packages/editor/src/plugins/media/services/html-video/index.tsx
kulla: > should we not check that first, before even trying to load?

A URL for a mp4 might not end with `.mp4` (same for images by the way) – even though this will be an edge case

packages/editor/src/plugins/media/services/types.ts
kulla: > I'm probably missing reasons where we need the type functions but I think we should consider static types where possible.

Yep, let's simplify the types in the end

kulla:
At https://serlo.slack.com/archives/C05QWHQP1HT/p1728588506725659 we have described a solution. Since in the implementation we will start new PRs I close this PR since it is not needed any more.

PR #4111, opened 2024-09-17
fix(editor-plugin-menu): Make mouse hover not change selected plugin

LarsTheGlidingSquirrel:
**Before**
Having the mouse hover unintentionally over a plugin in the menu caused the wrong plugin to be created on pressing 'enter'. 
Here the mouse hovers over 'Lückentext' and this will be created even though the user typed in 'text' and expects the first item in the list to be created. Even if the mouse is never moved / used. 
image

I think this also caused e2e test "focus plugins with arrow keys" to fail sometimes. 

**After**
Mouse hover does no longer select a plugin. But clicking to create still works & tooltips show up. I think "mouse hover to select" is a feature we can remove.

### DIFF
--- a/e2e-tests/tests/480-focus.ts
+++ b/e2e-tests/tests/480-focus.ts
@@ -102,17 +102,15 @@ Scenario('focus plugins with arrow keys', ({ I }) => {
   I.say('add first text plugin, type in it, check that it has focus')
   I.click('Füge ein Element hinzu')
   I.type('Text')
-  I.pressKey('Tab')
   I.pressKey('Enter')
   I.type('First text plugin')
   I.see('First text plugin', 'div[data-slate-editor="true"]:focus')
 
   I.say('add second text plugin, type in it, check that it has focus')
-  I.pressKey('Tab')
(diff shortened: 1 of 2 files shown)


hejtful:
There's this problem now:
Screenshot 2024-09-17 at 13 59 14
Hovering over an item, after using the arrow keys, makes two tooltips appear over each other.

I actually prefer it as it was so far.

LarsTheGlidingSquirrel:
Thank you for testing! I can also make two (overlapping) tooltips appear even without the changes in this pull request. First mouse hover, then arrow keys. Although this pull request causes a new way for this to happen. Not sure if this is important enough to fix right now.

LarsTheGlidingSquirrel:
To clarify the issue this pull request tries to solve: Even if the user only uses the keyboard to create a new plugin and never touches the mouse, pressing 'enter' will create the plugin under the mouse pointer, which could be over another plugin type by coincidence. 

My assumption is that this also made some e2e test fail.

hejtful (approved):
Alright, if the bug isn't new, then I'm good with the changes :)

PR #4114, opened 2024-09-17
refactor(dropzone-image): some refactoring, simplifications, a fix …

elbotho:
This became a bit bigger than planned…

There is [one important fix](https://github.com/serlo/frontend/pull/4114/commits/c52110939bfc2534874c5051664e11e6b8e70f1f) in there the rest is mostly moving stuff around.
I opted to removed the `AnswerZonesContext` (that was also providing plugin state). A few more props but a lot cleaner overall imho.


elbotho:
@hejtful sorry, I tried to extract the mandatory modal title into it's own PR but I still see it in the diff which makes this PR a bit unwieldy. Should I try to clean it up or is it okay to review like this?

hejtful (approved):

PR #4123, opened 2024-09-19
refactor(editor): plugins and plugin menu

hejtful:
Issue: PE-148


elbotho:
- exported interface should not change
- `EditorPluginType.PasteHack`: is deactivated for now, should not really be needed any more

elbotho:
@hejtful unfortunately we now have the same bundle size increase Mikey's PR ran into. It's probably the added SVGs strings inside the editor bundle.

hejtful:
Bug: When adding a `BlanksExercise` mode DnD from Exercise interface, a `BlanksExercise` mode Typing is added instead.

elbotho (approved):

PR #4135, opened 2024-09-20
Refactor remove exercise generation from serlo.org

CodingDive:
Replaced by Kullas AI content generation integration directly in the editor


elbotho (approved):
Thank you.
I removed more helpers, everything else looks fine.

PR #4155, opened 2024-09-27
refactor(editor): move editor strings to editor

elbotho:
PE-177

This moves the editor i18n strings back to the editor.
They are split into `editStrings` (needed for the editing view) and `staticStrings` (needed for the static renderers) and this also adds editor-specific i18n providers.

**Naming**
I'm not super happy with `staticStrings` but in the context it works okay (basically the same problem as with StaticRenderer). I'm also open to change this.

**Why I split the stings**
This split allows us to e.g. load the few staticStrings always and the editStrings only when actually editing. 
(this PR uses that approach for the frontend to keep the bundle size down)

**Until we setup crowdin:**
The first step (this PR) provides de and en strings manually and you have to always add both strings until we setup crowdin (once this is in staging).

**LoC:**
Overall there should be less LoC but again until crowdin runs we have a lot of duplicated strings.

**Language support**
We will only support en and de for now. That's no change for the packages but in the frontend we will lose es language support for now (we will add it again later).

**Reviewing**
Like it's parent #4154 just checking if everything still works is probably enough.


LarsTheGlidingSquirrel:
I get the error `Error: Attempt to use instance data outside of provider!` because in the integrations we do not provide `InstanceDataProvider`. Not sure if we should.

hejtful (approved):

hejtful:
I just noticed the comment from Lars. I didn't test in the integrations, only in frontend.

elbotho:
> I just noticed the comment from Lars. I didn't test in the integrations, only in frontend.

Thank you. I added a fix and asked Lars to retest via slack.

LarsTheGlidingSquirrel (approved):

PR #4157, opened 2024-09-28
refactor(editor): move localStorage saving to editor

elbotho:
part of PE-155


hejtful (approved):
Double cleanup, sweet!

LarsTheGlidingSquirrel:
In `serlo-editor-as-lti-tool` we are in an iframe so `window.location.pathname` would give us the same value `/app` everywhere. I have not checked the implementation thoroughly, but I think this might break thinks?

PR #4161, opened 2024-10-07
chore(e2e): update packages

elbotho:
i fixed some failing tests on staging in the meantime. I don't think the remaining tests fail because of the updates, but the missing updates make the tests fail on staging atm.


hejtful (approved):
Thank you! Do you expect the failing tests to pass on staging?

elbotho:
> Do you expect the failing tests to pass on staging?

No, I expect them to run and still fail. One step better than not running at all I think ;)

PR #4162, opened 2024-10-08
fix(plugin-rows): hydration error in anchor link copy tool

elbotho:
loading is lazily does not really make a difference any more so this is easier and solves the critical hydration error that is on staging right now.


elbotho:
merging to easily check the effect on the e2e tests

PR #4163, opened 2024-10-08
refactor(plugin-page-team): remove plugin, replace team page

elbotho:
… with static page in the frontend

for PE-157
[Demo](https://frontend-git-refactor-team-page-plugin-serlo.vercel.app/team)

changes to be communicated to ppl&culture:
- only en/es/de supported
- intro text changes via software unit
- rest as before via google sheet, but changes only appear on next deploy


hejtful (approved):
It's my first time looking at the code for this page, but this seems like a good solution to me.

PR #4171, opened 2024-10-09
ci(editor): Write custom github action for generating changelog from PR names and throw out the changeset package

CodingDive:
Split it into two actions. One that does the automated changelog based on branch names (we can still change it manually), runs on every branch. And the existing github action that publishes it to npm and pushes the tag as a release to Github. This one only runs on staging


.github/workflows/editor.yaml
```diff
@@ -4,105 +4,52 @@ on:
...
-          NEXT_VERSION=$(node -p "require('./packages/editor/package.json').version")
-          body=$(git log -1 --pretty=format:%B)
-          gh release create "v$NEXT_VERSION" --notes "$body"
+          VERSION=$(node -p "require('./packages/editor/package.json').version")
+          git tag v$VERSION
+          git push origin v$VERSION
+
+      - name: Create GitHub Release
+        if: steps.check_published.outputs.already_published == 'false'
+        uses: actions/create-release@v1
```
hugotiburtino: This action is not maintained anymore. You could use one of the maintained actions that are listed in the readme of [actions/create-release](https://github.com/actions/create-release).

hugotiburtino (changes requested):

.github/workflows/editor-changelog.yaml
```diff
@@ -0,0 +1,167 @@
+name: Generate Editor Changelog
```
hugotiburtino, line 1: Interesting code. 
Have you searched for actions out there that do the same? Because, if there is any, we wouldn't have to maintain this code here.

CodingDive: It's very specific to our monorepo structure and we only care about changes in /packages/editor. I couldn't find a good one with monorepo support and thought it'd be easier to write the code myself

.github/workflows/editor.yaml
```diff
@@ -4,105 +4,52 @@ on:
...
+        with:
```
hugotiburtino: BTW, I see that you create a changelog in the other workflow, but the body of the release remains clean. Is that intentional?

.github/workflows/editor.yaml
CodingDive: Thank you, Hugo! 

I decided to go with this one. Do we need to add it from the market place or will the github action runner pick it up automatically? 

https://github.com/elgohr/Github-Release-Action

.github/workflows/editor.yaml
CodingDive: It was not intentional, I think it should extract the changelog of /editor/packages correctly, but I haven't been able to run the code to test it out :crossed_fingers:

.github/workflows/editor.yaml
```diff
@@ -4,105 +4,66 @@ on:
...
+        run: echo "CURRENT_VERSION=$(node -p "require('./packages/editor/package.json').version")" >> $GITHUB_ENV
 
-      - name: Check if the next version is already published
+      - name: Check if current version is already published
         id: check_published
         run: |
-          NEXT_VERSION=$(node -p "require('./packages/editor/package.json').version")
-          if npm view @serlo/editor@$NEXT_VERSION > /dev/null 2>&1; then
+          CURRENT_VERSION=$(node -p "require('./packages/editor/package.json').version")
+          if npm view @serlo/editor@$CURRENT_VERSION > /dev/null 2>&1; then
```
hugotiburtino: There is some repetition in this part.
What about merging both?

.github/workflows/editor.yaml
```diff
@@ -4,105 +4,66 @@ on:
...
-          done
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Publish to npm (after PR merge)
-        if: steps.check_published.outputs.already_published == 'false' && steps.create_pr.outputs.pr_url
-        run: yarn editor:publish
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          CURRENT_VERSION=$(node -p "require('./packages/editor/package.json').version")
```
hugotiburtino: If you stored the CURRENT_VERSION before, you just need to retrieve it from `env.CURRENT_VERSION`

hugotiburtino (approved):

.github/workflows/editor.yaml
hugotiburtino: github action runner pick it up automatically

packages/editor/.prettierignore
```diff
@@ -1,3 +1,4 @@
 # this also checks .gitignore
 
-dist
\ No newline at end of file
+dist
+CHANGELOG.md
```
hugotiburtino: Instead of gitignoring it, what about adding a prettier step `yarn prettier -w packages/editor/CHANGELOG.md` before commiting it?

CodingDive: Good idea! Let's try it :crossed_fingers:

PR #4172, opened 2024-10-09
fix(e2e): adapt math tests

elbotho:


elbotho:
merging so I can test against staging again

PR #4174, opened 2024-10-09
feat(editor): add simple preview without frontend

elbotho:
### work on the editor offline and without next running

run `yarn dev` in the editor and you can test both the web component and react (package) implementations as http://localhost:5173/

image

--- 

~@CodingDive how problematic is [removing this line](https://github.com/serlo/frontend/compare/feat/simple-editor-dev?expand=1#diff-1547d15e49275239192ba31e15717b74874cba0aa537e1df244e332b61b5bf4cR34)? With this set the dev server does not work somehow 🤷~


CodingDive:
> @CodingDive how problematic is [removing this line](https://github.com/serlo/frontend/compare/feat/simple-editor-dev?expand=1#diff-1547d15e49275239192ba31e15717b74874cba0aa537e1df244e332b61b5bf4cR34)? With this set the dev server does not work somehow 🤷

Believe it or not, this will break the build, or rather cause an error in the integrations that I haven't found a way to supress :sweat_smile: 

Maybe we can use two files
vite.config.lib.js - Configuration for building the packages
vite.config.dev.js -  Config for running dev server

```bash
vite build --config vite.config.lib.js
vite --config vite.config.dev.js
```

Or modes / commands

```javascript
export default defineConfig(({ command, mode }) => {
  if (command === 'build' && mode === 'lib') {
    return {
      ....
    }
  } else {
    return ... 
  }
})
```

Then pass the args in our package.json run scripts like this

```javascript
vite build --mode lib
```

elbotho:
@CodingDive would that work? https://github.com/serlo/frontend/pull/4174/commits/9ef60b430aca541869fa91317451978e9fe36a39 
Or even better: Can you tell me how I can check myself?

CodingDive:
> @CodingDive would that work? https://github.com/serlo/frontend/commit/9ef60b430aca541869fa91317451978e9fe36a39

Works right now :raised_hands: 

Thank you!

packages/editor/demo/lit/index.js
```diff
@@ -0,0 +1,130 @@
...
+      <select @change=${this.handlePluginChange.bind(this)}>
+        ${Object.values(pluginMenuDe).map(
+          ({ type, title }) => html`
+            <option value="${type}" ?selected=${type === this.selectedPlugin}>
+              ${title}
+            </option>
+          `
+        )}
+      </select>
+      <div style="margin-top: 20px;">
```
hejtful:
```suggestion
      <div style="margin-top: 70px;">
```
Just so that the buttons and select are usable when the plugin toolbar is present.

hejtful (approved):
This is so nice 😍 Thank you, it will be very helpful in the long run

hejtful:
@elbotho It would be great to have a preview for the editor rendered within an iframe as well, as that's how it's rendered for some LTI integrations. Would you prefer if we add it to this PR, or open a follow-up PR once this is merged?

elbotho:
> @elbotho It would be great to have a preview for the editor rendered within an iframe as well, as that's how it's rendered for some LTI integrations. Would you prefer if we add it to this PR, or open a follow-up PR once this is merged?

Good idea, no preference :) 

I'm just waiting for a 👍 from @LarsTheGlidingSquirrel here before merging this to see if it works for him.

LarsTheGlidingSquirrel (approved):
Seems to work great! I think it will be really valuable for us to have during development. Thank you so much! <3

PR #4175, opened 2024-10-10
fix(editor-packages): Use one redux store per instance of editor

CodingDive:


packages/editor/src/plugins/dropzone-image/components/answer-zone/answer-zone-answer.tsx
```diff
@@ -21,12 +22,15 @@ interface AnswerZoneAnswerProps {
...
+  const store = useStore()
   const { answer, isOnlyAnswer, onEditAnswer, onRemoveAnswer } = props
 
   const pluginStrings = useEditStrings().plugins.dropzoneImage
 
-  const answerImageUrl = getAnswerZoneImageSrc(answer.image.id)
-  const answerText = getAnswerZoneText(answer.text.id)
+  const answerImageUrl = getAnswerZoneImageSrc(answer.image.id, () =>
+    store.getState()
+  )
```
CodingDive, line 32: I know that here I could technically just pass `store.getState` but to me, it's less clear that we are passing in a fn than using the arrow function syntax. Let me know which syntax you prefer

hejtful: This is fine with me.

In general, these utils shouldn't be calling the store state getter themselves, they should be provided only the relevant piece of state for better encapsulation...but that's another topic, unrelated to this PR 😁

packages/editor/src/store/store.ts
```diff
@@ -1,38 +1,68 @@
...
+  sagaMiddleware.run(saga)
+
+  return store
+}
+
+// one store: Record<editorId, { state }>
+// multiple stores: one store per editor instance / per editor id
+
+// Export singleton instance of store
+// export const store = createStoreInstance()
```
hejtful: Leftover from our call? 😁

hejtful (approved):
Clean and simple, thank you for the quick fix!

packages/editor/src/store/store.ts
CodingDive: Ahh nice catch! 😅

PR #4178, opened 2024-10-10
Some changes to make editor code work in serlo-editor-as-lti-tool

LarsTheGlidingSquirrel:
# Changes
- **fix(rows): check if interactive exercise plugins are supported in menu items filter** For the integrations we need to disable e.g. H5P (for edu-sharing also dropzone image because we do not support the image plugin there) 
- **fix(plugin-menu): show icons when choosing exercise type in exercise plugin** - @elbotho I created a new component `plugin-menu-item.tsx`. This is now used in the plugin menu and also when choosing the exercise type in the exercise plugin. 
- **fix(table-plugin): only show cell switch button when image plugin exists** Now uses `isSupported` because `getByType` returns unsupported plugin which is not falsy. 
- **fix: correct property `fill-opacity` in svgs** Opacity value is no longer ignored. 
- **fix(plugin-menu): add edusharing plugin to plugin menu**
- **chore: use new endpoints for edusharing embeds**


packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx
```diff
@@ -83,7 +83,18 @@ export function PluginMenuModal({ onInsertPlugin }: PluginMenuModalProps) {
...
+          const isSupported =
+            type === 'singleChoiceExercise' || type === 'multipleChoiceExercise'
+              ? editorPlugins.isSupported(EditorPluginType.ScMcExercise)
+              : type === 'blanksExerciseDragAndDrop'
+                ? editorPlugins.isSupported(EditorPluginType.BlanksExercise)
+                : editorPlugins.isSupported(type)
+          return (
+            initialState.plugin === EditorPluginType.Exercise && isSupported
+          )
+        }
```
elbotho: Okay, I think the actual bug is in [line 70](https://github.com/serlo/frontend/pull/4178/files#diff-06129561f4c9cfba0cb6e732548c338ac3524f2298bf969febdb828410e49496L70) where we check for plugin support (but since `exercises` is supported we don't filter the interactive exercise plugins).

elbotho (approved):
Commented on the "hack", will probably create a PR so you can check if my hunch is correct.
Rest looks good! Thank you 🎉

packages/editor/src/plugins/rows/components/plugin-menu-modal.tsx
elbotho: [PR](https://github.com/serlo/frontend/pull/4180)

hejtful (approved):
<3

PR #4180, opened 2024-10-10
fix(rows): check if interactive exercise plugins are supported in menu items filter

elbotho:


LarsTheGlidingSquirrel (approved):
Thank you! It works and it seem like a better place for this check. :slightly_smiling_face:

PR #4189, opened 2024-10-14
refactor(image): simplify image uploading

elbotho:
replace custom image configs with one that supports S3, new api route for signed urls

(checkout bucket at https://play.min.io:9443/browser/serlo-test-bucket with `minioadmin // minioadmin` as credentials)

### todo:
change api url to get presigned request depending on env or editor prop


elbotho:
closed in favour of #4572

PR #4199, opened 2024-10-19
refactor(editor): move main toolbar / undoredo to editor

elbotho:
first step of PE-160


packages/editor/src/editor-ui/editor-toolbar/leave-confirmation-render-null.tsx
hejtful, line 1: This file is still using the `useLeaveConfirm` hook from frontend. That hook is only used in this file, but it uses Next router, so not sure what to do with it. Just wanted to point it out, in case you missed that there is still a frontend import here.

\+ Not so important, but I would call this file `leave-confirmation.tsx`, to match the name of the component.

packages/editor/src/editor-ui/editor-toolbar/save-button.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { SaveModal } from '@serlo/frontend/src/serlo-editor-integration/components/save-modal'
+import { useRef, useState } from 'react'
+
+import { ClientOnlyPortal } from './client-only-portal'
+import { LeaveConfirmation } from './leave-confirmation-render-null'
+import { entity } from '../../plugins/serlo-template-plugins/common/common'
+
+interface ToolbarMainProps {
+  changes?: StateTypeReturnType<(typeof entity)['changes']>
+  licenseId?: StateTypeReturnType<(typeof entity)['licenseId']>
```
hejtful, line 17: I had no idea we could do this! Sweet :3

hejtful (approved):

packages/editor/src/editor-ui/editor-toolbar/leave-confirmation-render-null.tsx
elbotho: you are right of course. I could remove this file completely in a later PR.

PR #4200, opened 2024-10-19
refactor(frontend): remove email notification setting

elbotho:
We don't send notifications any more so the setting is obsolete.


hejtful (approved):
cqG5aFdTkk5ig

PR #4202, opened 2024-10-19
refactor(editor): move save modal outside of editor, refactor

elbotho:
For PE-157

(removes the function function to skip review for authorized users)


hejtful (approved):
Thanks!

PR #4206, opened 2024-10-20
refactor(editor): refactor providers, hack last frontend deps

elbotho:
- removes save context
- rename isSerlo context into SerloExtraContext
- use SerloExtraContext to provide last extrawurst serlo stuff


packages/editor/src/utils/serlo-extra-context.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+interface SerloExtraData {
+  isSerlo?: boolean
+  licenses?: LicenseData[]
+  ArticleAddModal?: (props: ArticleAddModalProps) => JSX.Element
+}
+
+/** used to enable features only in a serlo context and not e.g. on edusharing */
+export const SerloExtraContext = createContext<SerloExtraData>({})
```
hejtful: Would you consider a name like `SerloOnlyFeaturesContext` or `SerloExtraFeaturesContext`, to make it more clear what this context is about?

packages/editor/src/utils/serlo-extra-context.ts
hejtful, line 1: This file feels a bit messy, because there are multiple unrelated things going on. Maybe add a comment explaining the purpose of the file? Something like:
```
// Some features of the editor are only enabled on serlo.org.
// This file provides types relating to these features, and a
// React Context that handles data relating to these features.
```

(What confused me at first was that you moved `ArticleAddModalProps` to this file, but still keep importing `ArticleProps`. I figured out the logic behind that, but I still have a feeling that this file might be confusing the next time I look at it)

hejtful (approved):

elbotho:
@hejtful good suggestion, thank you

PR #4207, opened 2024-10-20
refactor(editor): separate editor and frontend

elbotho:
- removes last frontend imports from editor  
- removes unused packages in both editor and frontend
- removes import aliases
- cleanup


apps/web/src/components/pages/team/team-grid.tsx
```diff
@@ -1,8 +1,9 @@
 import { cn } from '@editor/utils/cn'
```
hejtful: Unrelated to this PR probably, but this looks like a wrong import. It's importing `cn` from the editor, even though we have one in web.

packages/editor/src/package/editor.tsx
```diff
@@ -22,7 +22,9 @@ import {
   type EditorVariant,
 } from './storage-format'
 
-import '@/assets-webkit/styles/serlo-tailwind.css'
+// TODO: figure out styling
```
hejtful, line 26: Nice that there's already a Linear issue for this! <3

hejtful (approved):

PR #4212, opened 2024-10-22
fix(editor): prevent creating exercises in multimedia, other exercise, solution

LarsTheGlidingSquirrel:
I tested some nestings and if seems to work. @elbotho Could you check as well? Maybe I forgot some nestings.


elbotho (approved):
Thank you! 
Tbh. I did not realise this was an editor problem. 

> Maybe I forgot some nestings.

For now it looks good. Rest of the bugs will show themselves :)

PR #4213, opened 2024-10-23
fix(edusharing-plugin): not use `dangerouslySetInnerHtml` & sanitize html to prevent XSS

LarsTheGlidingSquirrel:
Addresses feedback from security audit. 

Changes: `dangerouslySetInnerHtml` is not used any more because it might allow XSS attacks. Instead, we render all embeds in an iframe which provides encapsulation from the host and sanitize the HTML beforehand.  

Followup: Check for visual errors in media embeds


hejtful (approved):
From what I understood, it would have been enough to use either:
- iframes with `IframeResizer` or
- `dangerouslySetInnerHTML` with `DOMPurify`

But a double layer of protection can't hurt 👍

PR #4216, opened 2024-10-23
feat(editor): show user a warning in test environments that content might disappear

LarsTheGlidingSquirrel:
image

Background: Some people were excited to use the editor and started to create content they want to keep even in the test environments. But for the devs this makes it more complicated. 

Change: All test environments show a warning label above the editor. 

@elbotho @GregorZupan Any wishes in terms of design / wording? Would you prefer to not have something like this?

### DIFF
--- a/packages/editor/src/i18n/strings/de/edit.ts
+++ b/packages/editor/src/i18n/strings/de/edit.ts
@@ -5,6 +5,8 @@ export const editStrings = {
   confirmRouteChange:
     'Willst du wirklich die Seite verlassen ohne zu speichern?',
   noChangesWarning: 'Bisher hast du nichts geändert',
+  savedContentMightDisappearWarning:
+    '⚠️ Dies ist eine Testumgebung. Bitte erstell hier noch keine Inhalte, die du langfristig behalten willst.',
   addPluginsModal: {
     searchInputPlaceholder: 'Suche...',
     basicPluginsTitle: 'Inhalte',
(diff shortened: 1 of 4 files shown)


packages/editor/src/i18n/strings/de/edit.ts
```diff
@@ -5,6 +5,8 @@ export const editStrings = {
   confirmRouteChange:
     'Willst du wirklich die Seite verlassen ohne zu speichern?',
   noChangesWarning: 'Bisher hast du nichts geändert',
+  savedContentMightDisappearWarning:
+    '⚠️ Dies ist eine Testumgebung. Bitte erstell hier noch keine Inhalte, die du langfristig behalten willst.',
```
elbotho:
```suggestion
    '⚠️ Dies ist eine Testumgebung. Bitte erstelle hier noch keine Inhalte, die du langfristig behalten willst.',
```

packages/editor/src/package/editor.tsx
```diff
@@ -74,6 +76,7 @@ export function SerloEditor(props: SerloEditorProps) {
       <EditStringsProvider value={editStrings}>
         <EditorVariantContext.Provider value={editorVariant}>
           <LtikContext.Provider value={_ltik}>
+            {!isProductionEnvironment ? renderTestEnvironmentWarning() : null}
```
elbotho:
```suggestion
            {isProductionEnvironment ? null : renderTestEnvironmentWarning()}
```
For me this is more readable / less cognitive load (but it might mostly be a preference now since this is how other people told me to do it :)

elbotho (approved):
Two small things.
I like it overall, thank you.
(Maybe I will make it more prominent at some point)

PR #4227, opened 2024-10-28
fix(dropzone-image): ignore empty dropzones for check button visibility logic

hejtful:


LarsTheGlidingSquirrel (approved):
Nice! Works great! :relaxed:

PR #4239, opened 2024-10-30
refactor(edusharing-asset): replace react-modal

elbotho:


LarsTheGlidingSquirrel:
Made the iframe full height and nearly full width (so you can still see it is a frame floating above the content) because we show the full edu-sharing UI in there and the more space we have the better. 
image

LarsTheGlidingSquirrel (approved):

PR #4251, opened 2024-11-05
feat(editor): add user id to image upload

elbotho:
also some refactoring.
I'm adding a new context `EditorMetaContext` that combines `EditorVariantContext`, `LtikContext` and adds `userId`.


hugotiburtino (approved):

LarsTheGlidingSquirrel (approved):
I like these changes! Thank you!
Only a small thing: The `ltik` contains a bunch of context information (userId, ...). So when we pass a `ltik` the `userId` is just a duplicate. We could make the API accept either an `ltik` (only in LTI) OR a `userId` (everywhere else). But I am also fine like it is right now.

elbotho:
> Only a small thing: The ltik contains a bunch of context information (userId, ...). So when we pass a ltik the userId is just a duplicate. We could make the API accept either an ltik (only in LTI) OR a userId (everywhere else). But I am also fine like it is right now.

Thanks, good point.
I think for now I slightly prefer that the editor does not need to know about how ltik is constructed but let's see.

PR #4252, opened 2024-11-06
fix(editor): commit temporary actions to history only after resolving them

### LINKED ISSUE #4241: editor(plugin-image): uploaded image not in state until next change
To reproduce:
1. Go to https://de.serlo-staging.dev/___editor_preview
2. Upload an image
3. Copy the editor state
4. Observe that the image src **is not** in the editor state
5. Make another change (for example, type in the caption field)
6. Copy the editor state again
7. Observe that the image src **is** in the editor state now

hejtful:
Resolves #4241 

Note: First commit is only cleaning up, second commit fixes the bug.

This took way longer to figure out than I'd like to admit 😬
The editor emits `onChange` events when actions are committed to history, not when actions are resolved.
For most changes, this doesn't matter, as we first resolve actions, then commit them to history.
There are, however, these "temporary" actions, used in `asyncScalar`. One use-case of `asyncScalar` is uploading an image, where we had the bug. And in these "temporary" actions, we were first committing actions to history, then resolving them, which led the editor to emit `onChange` before its state was updated.

I snooped around the code to see if I could find any reason for this difference between regular and "temporary" actions, and I couldn't find any. I also tested the editor, and everything seems to work fine, but it would be great if you could test a bit as well.

Explanation for "temporary" actions (`asyncScalar`): I believe these were initially created to support loading states of async actions. So once a user uploads an image, a temporary action would show a loading screen, and then a normal action would show the uploaded image and the temporary action would be removed from history. That would make sure that undo/redo works as expected -- never showing the loading screen. I couldn't figure out how to use this, though. Either the feature was never finished, or we broke it unknowingly.

@elbotho This also fixes the ___editor_preview -- the right side will now show the image as soon as it's uploaded.

### DIFF
--- a/packages/editor/src/store/history/saga-actions.ts
+++ b/packages/editor/src/store/history/saga-actions.ts
@@ -9,5 +9,5 @@ export const runCommitActionToHistorySaga = createAction<ReversibleAction[]>(
 
 export const runCommitTemporaryActionToHistorySaga = createAction<{
   initial: ReversibleAction[]
-  executor?: StateExecutor<ReversibleAction[]>
+  executor: StateExecutor<ReversibleAction[]>
 }>('history/commitTemporaryActionToHistory')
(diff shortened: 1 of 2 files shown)


elbotho (approved):
👍 thank you for the deep dive! 
Not sure I would have figured that one out.

LarsTheGlidingSquirrel (approved):
I don't know enough about the relevant code but I tested the editor and performed a lot of undo/redo and found no issue. 
Thank you for looking into this and finding a solution!

PR #4258, opened 2024-11-08
fix(plugin-edusharing): fix visual errors in embeds

LarsTheGlidingSquirrel:
Can be improved but fixes a lot of visual errors already


packages/editor/src/plugins/edusharing-asset/renderer.tsx
```diff
@@ -32,6 +31,12 @@ const EmbedJson = t.type({
   }),
 })
 
+const iframeResizerHtml =
+  '<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.3.9/iframeResizer.contentWindow.min.js"></script>'
+
+// Set some styles to default values to create consistency between different browsers similar to a CSS reset, see https://meyerweb.com/eric/tools/css/reset/
+const defaultStyle = 'padding: 0; margin: 0; border: 0;'
```
hejtful: Imo, CSS reset is widely known enough for this comment to be unnecessary. Maybe consider:
```suggestion
const cssReset = 'padding: 0; margin: 0; border: 0;'
```

hejtful (approved):
Looks good to me! 👍

packages/editor/src/plugins/edusharing-asset/renderer.tsx
LarsTheGlidingSquirrel: I changed it. :slightly_smiling_face:

PR #4273, opened 2024-11-13
Fix SSO button after upgrade of kratos in backend

hugotiburtino:
https://frontend-git-fix-auth-oidc-button-serlo.vercel.app/auth/login

## Bug
After clicking on an SSO button we get the following error message:

image


elbotho (approved):
Thank you @hugotiburtino 
I added a small commit. If you agree you can merge.

elbotho:
PS: We should probably have a simple SSO e2e test, but since this is not super high priority right now it's probably backlog

PR #4274, opened 2024-11-13
chore: upgrade eslint to v9

hejtful:
Seems to work alright. Issue: PE-115

Follow-ups:
- Upgrade "@typescript-eslint/eslint-plugin" and "@typescript-eslint/parser"
- Switch to recommended default rulesets, or at least reduce the amount of custom rules


elbotho (approved):
👍 Nice! Very glad it worked. 
Thank you for adding the type definitions in the configs.
Also nice how many ignore comments this get's rid of (even tough I'm not sure why)

hejtful:
I think that eslint just started reporting unnecessary disable comments as warnings in this version. None of these were needed in the current state of the code, except `svg.d.ts`'s "no-default-export" disable, but I didn't look deep into it, as "no-default-export" is not a very important rule.

PR #4289, opened 2024-11-15
fix(link-plugin): Allow absolute serlo urls in integrations

CodingDive:
https://linear.app/serlo/issue/DEV-233/bug-inserting-links-is-confusing

The problem was that it strips the serlo.org and then treats every URL as a relative URL. This lead to you being lead to some weird moodle/chancenwerk pages that don't exist and yield 404s


packages/editor/src/plugins/text/utils/link.ts
```diff
@@ -1,14 +1,28 @@
-export function getCleanUrl(inputUrl: string, instance?: string) {
-  const testId = parseInt(
-    inputUrl.match(
-      /https?:\/\/([a-z]{2}\.)?serlo\.org(\/[a-z]+)?\/([1-9]?[0-9]+)/
-    )?.[3] ?? 'NaN'
+export function getCleanUrl(
+  inputUrl: string,
+  instance?: string,
+  yieldAbsoluteUrl: boolean = false
```
elbotho: I'd prefer to also call it `isSerlo` inside this function since we only allow serlo absolute urls and this is all still very serlo specific.

elbotho: Follow up: If `isSerlo` is false there is not much happening. I think this function might be simpler with a 
`if(!isSerlo) return …` early return and keeping the rest of the code as is.

elbotho (changes requested):
Thanks you for looking into it!

packages/editor/src/plugins/text/utils/link.ts
```diff
@@ -1,14 +1,35 @@
-export function getCleanUrl(inputUrl: string, instance?: string) {
-  const testId = parseInt(
-    inputUrl.match(
-      /https?:\/\/([a-z]{2}\.)?serlo\.org(\/[a-z]+)?\/([1-9]?[0-9]+)/
-    )?.[3] ?? 'NaN'
-  )
+export function getCleanUrl(
+  inputUrl: string,
+  instance?: string,
+  isSerlo: boolean = true
```
elbotho:
```suggestion
  isSerlo: boolean,
  instance?: string
```

elbotho:
@CodingDive does it make sense to support entering just a serlo Id outside of serlo?

CodingDive:
> @CodingDive does it make sense to support entering just a serlo Id outside of serlo?

I can also change our placeholder. There it says /1234. I think it's a nice convenience, but not sure how often it'll be used.

image

elbotho (approved):

PR #4292, opened 2024-11-15
chore: upgrade yarn & node

elbotho:


elbotho:
vercel not yet supporting node 22 😢

PR #4298, opened 2024-11-19
fix(plugin-edusharing): allow focusing the plugin by clicking on the content

LarsTheGlidingSquirrel:
**Before**

Clicking on the embedded content did not focus the editor plugin because it is inside an iframe. It was nearly impossible to make the toolbar show up to select another type of content or change the size. 

**After**

Render an overlay over the iframe if the plugin is unfocused to detect if the user wants to focus the plugin. After this initial click to focus the plugin the user can interact with it.

### DIFF
--- a/packages/editor/src/plugins/edusharing-asset/editor.tsx
+++ b/packages/editor/src/plugins/edusharing-asset/editor.tsx
@@ -59,22 +59,33 @@ export function EdusharingAssetEditor({
     <>
       {renderPluginToolbar()}
       {renderModal(ltik)}
-      <EdusharingAssetRenderer
-        nodeId={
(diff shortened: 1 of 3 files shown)


hejtful (approved):
👍

elbotho:
@LarsTheGlidingSquirrel to test this myself I would need to use yalc and run this in the other repo, right?

LarsTheGlidingSquirrel:
> @LarsTheGlidingSquirrel to test this myself I would need to use yalc and run this in the other repo, right?

Yes, you can test an image embed locally in `serlo-editor-as-lti-tool`. Or all types of embeds in Uberspace staging through either the edu-sharing or itslearning staging environment.

PR #4301, opened 2024-11-19
refactor(editor): normalize links only on serlo.org

elbotho:
@CodingDive this is roughly what I had in mind, what do you think?


CodingDive:
Works for me, thank you! ❤️ I'll change the placeholder

PR #4307, opened 2024-11-20
fix(editor): Render toasts in Toaster

CodingDive:


LarsTheGlidingSquirrel (approved):
Works great! Now I get a bunch of notifications that I didn't know existed. :sweat_smile: 
Possible improvement: We could try to have `<Toaster />` only in one place (in the editor)

CodingDive:
Serlo web also needs it e.g save dialog

LarsTheGlidingSquirrel:
> Serlo web also needs it e.g save dialog

Ah, ok. Then I guess we need multiple occurrences.

PR #4308, opened 2024-11-21
chore: upgrade Slate

hejtful:
- Upgrade @prezly/slate-lists, slate and slate-react
- Add slate-dom to dependencies to fix an error
- Remove @types/slate-react as it was unnecessary
- Add @types/is-hotkey as it became necessary 🤷 
- Make Slate-related versions fixed in package.json
- \+
- Fix failing e2e tests on staging


CodingDive (approved):
Works great on my end! Thank you ❤️ 
Just curious, were these the fuzzy tests before, or did they break as part of the upgrade?

hejtful:
They're from the focus/hover redesign, I just noticed them on staging after merging 😁

LarsTheGlidingSquirrel (approved):
Thank you! Tested the text plugin a bit and found no issue regarding the slate upgrade.

e2e-tests/tests/430-multimedia-plugin.ts
```diff
@@ -43,7 +43,7 @@ Scenario('Multimedia plugin toolbar controls', async ({ I }) => {
     'aria-selected': 'true',
   })
   I.click('$modal-close-button')
-  I.seeElement('.media-wrapper.mobile\\:w-1\\/4')
+  I.seeElement('.media-wrapper.mobile\\:basis-1\\/4')
```
CodingDive, line 46: Would be nice to replace this with data-qa attribute

hejtful: I wouldn't agree in this case. This line is checking if the "quarter width media" option has been set correctly, and the best way to check that is by checking if the `.media-wrapper` element has the correct class.

PR #4313, opened 2024-11-25
feat(plugin-rows): new plugin reorder buttons

hejtful:
[Linear](https://linear.app/serlo/project/make-moving-the-plugins-on-canvas-easier-24b0fb77e755/issues)

[Demo](https://frontend-git-feat-rows-new-reorder-buttons-serlo.vercel.app/entity/repository/add-revision/277232)


LarsTheGlidingSquirrel:
The animation looks really nice and helps a lot with keeping track of where stuff is flying! :green_heart:  
Only when the moved element moves off screen, the content jumps and it is a little disorienting to me.

hejtful:
Any ideas on how to improve this case? :)

packages/editor/src/plugins/rows/components/rows-inner-editor.tsx
```diff
@@ -16,6 +17,9 @@ import {
   PluginMenuContext,
 } from '../contexts/plugin-menu'
 
+const loadMotionFeatures = () =>
+  import('motion/react').then((res) => res.domMax)
+
```
LarsTheGlidingSquirrel, line 22: What was the reason for doing it this way?

packages/editor/src/plugins/rows/editor-row-renderer.tsx
```diff
@@ -191,15 +190,15 @@ export function EditorRowRenderer({
...
-          '[&:has(.rows-editor-renderer-container:focus-within)>.rows-tools]:opacity-0',
-          '[&:hover>.rows-tools]:!opacity-100',
+          // Row controls (left sidebar) styles
+          '[&:focus-within>.row-controls]:opacity-100',
+          '[&:has(.rows-editor-renderer-container:focus-within)>.row-controls]:opacity-0',
+          '[&:hover>.row-controls]:!opacity-100',
           isMultimediaPlugin &&
-            '[&>.rows-tools]:!-left-1 [&>.rows-tools]:!-top-9'
+            '[&>.row-controls]:!-left-1 [&>.row-controls]:!-top-9'
         )}
```
LarsTheGlidingSquirrel, line 199: Could these styles also be part of component `RowControls`?

packages/editor/src/plugins/rows/components/row-editor.tsx
```diff
@@ -36,8 +37,9 @@ export function RowEditor({
   }
 
   return (
-    <div
-      key={row.id}
+    <motion.div
+      layout
+      transition={{ duration: 0.3, type: 'linear' }}
```
LarsTheGlidingSquirrel, line 42: Maybe ask Gregor for transition type and duration in terms of good UX.

packages/editor/src/plugins/rows/components/row-controls.tsx
```diff
@@ -0,0 +1,65 @@
...
+import type { RowsPluginState } from '..'
+
+interface RowDragButtonProps {
+  rows: StateTypeReturnType<RowsPluginState>
+  index: number
+}
+
+export function RowControls({ rows, index }: RowDragButtonProps) {
+  function handleUp() {
+    rows.move(index, index - 1)
```
LarsTheGlidingSquirrel: Good to know that this exists already. :slightly_smiling_face:

LarsTheGlidingSquirrel (approved):
Added some comments but nothing blocking merging into staging. Thank you! :green_heart:

LarsTheGlidingSquirrel:
> Any ideas on how to improve this case? :)

Hm, either no automatic scrolling or smooth scrolling. No scrolling might move the element out of sight and if you want to move it further you have to find it again first. Without thinking about it much, I think smooth scrolling after a short delay could help.

packages/editor/src/plugins/rows/components/rows-inner-editor.tsx
hejtful: Reducing the initial load. Before lazy loading, the increase in bundle size was much bigger.

I followed the guide from Motion themselves, just omitted the extra file they recommend, as I didn't think we needed that: https://motion.dev/docs/react-lazy-motion

packages/editor/src/plugins/rows/editor-row-renderer.tsx
hejtful: This Tailwind way of applying CSS hacks is consistently used by Botho across the project, and I honestly prefer it to passing props along components for CSS hacks.

Ideally, we would try to separate UI components from functional components fully, and then we could re-think this approach. But for now, I'd keep this here.

Please note that I only renamed the classes to follow the new component name.

packages/editor/src/plugins/rows/components/row-editor.tsx
hejtful: Gregor saw the demo while the PR was a draft already 😄 

And the desired animation is shown in [the ticket](https://linear.app/serlo/issue/EDTR-169/moving-animation-when-switching-with-arrows) already. Would you say the end result doesn't match the video from the ticket?

hejtful:
> Hm, either no automatic scrolling or smooth scrolling. No scrolling might move the element out of sight and if you want to move it further you have to find it again first. Without thinking about it much, I think smooth scrolling after a short delay could help.

Thanks for your input! I'll talk about it with Gregor.

The problem is that, when an element is moved above the viewport, the browser scrolls up automatically, and we can't control this jump, or at least I didn't find a way to control it. If Gregor agrees to apply a smooth scroll, I'd need a brainstorm with Botho and you to figure out how to prevent or control this scroll-up behavior 😄

hejtful:
Added smooth scrolling, to the place where the moved plugin will move, right before moving the plugin. On Gregor's request, also added some offset between the plugin and the edge of the window.

LarsTheGlidingSquirrel:
Thank you for explaining! :green_heart:

elbotho:
really nice with the scrolling 👍

As always there is some nesting fun. It's not great but also no real blocker I guess.
image
image
image

PR #4320, opened 2024-11-28
feat(interactive-video): add tool to remove all marks

elbotho:
also: [try save info text instead of save button](https://github.com/serlo/frontend/pull/4320/commits/058dc183febdd9a5f019512bb6201236f859de4d)


hejtful (approved):
Thanks!

PR #4323, opened 2024-11-29
fix(page-editor-preview): hide editor toolbar artifacts

LarsTheGlidingSquirrel:
Removes the white artifact on the right
image

The container for the toolbar save button was rendered. 

Also removes slight oversizing of the left and right page.


elbotho (approved):
thank you!
one small suggestion in https://github.com/serlo/frontend/pull/4325

PR #4324, opened 2024-11-29
chore: add open iframe resizer code to assets

elbotho:
@LarsTheGlidingSquirrel this way we can start using it from https://de.serlo.org/_assets/open-iframe-resizer-1.2.1.min.js as soon as this is deployed.
If you prefer we can also find a way to host it on uberspace via editor.serlo.org


LarsTheGlidingSquirrel (approved):
Nice! Thank you!

PR #4330, opened 2024-12-02
lernpfad intro modal first draft

elbotho:
https://linear.app/serlo/issue/DEV-278/intro-ansicht-vor-beginn-der-aufgabe


elbotho:
now part of #4331

PR #4340, opened 2024-12-06
feat(editor-web-component): Expose language prop

CodingDive:


CodingDive:
Need to release editor version and editor-web-component next!

PR #4343, opened 2024-12-09
fix(plugin-rows): whole plugin drag handle bug

hejtful:
I introduced this bug while implementing the new arrow buttons functionality.


elbotho (approved):
👍 thank you!

PR #4344, opened 2024-12-10
refactor(image-plugin): Show better error messages when the image upload fails

CodingDive:
**Before**

No error message 

**After** (old upload)

image

### DIFF
--- a/packages/editor/src/editor-integration/image-with-testing-config.ts
+++ b/packages/editor/src/editor-integration/image-with-testing-config.ts
@@ -127,7 +127,10 @@ export function createReadAndUploadFile(secret: string) {
 
     const { data } = (await response.json()) as { data: MediaUploadQuery }
 
-    if (!data?.media?.newUpload) {
+    if (
+      !data?.media?.newUpload?.uploadUrl ||
+      !data.media.newUpload.urlAfterUpload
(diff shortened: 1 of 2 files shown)


elbotho:
Pretty sure this is changing the new upload not the old. 
Sorry, can't really review without some context.

CodingDive:
Sorry, yes you are right.  This is adding some more error handling to the new method. I found the code for the old way with the secret and adding it there too right now :+1: 

The problem is that in integrations like Moodle, no error is shown when no testing secret is set, or something goes wrong.

As we know, locally they'll already use the new bucket and in their staging / prod, the old way. So both should always show a toast when something goes wrong!

CodingDive:
@hejtful I refactored and simplified the upload code a little bit. There was a file reader that was not used at all for streaming (other than a `onLoad` callback). I tested it and it seems to work great, but if there was a particular reason why it was there, that I missed, I can add it back.

hejtful (approved):
@CodingDive The code for `testingSecret` temporary solution was just copied from `/apps/web/src/serlo-editor-integration/image-with-serlo-config.ts` and adjusted. If you tested all realistic scenarios and everything works, I'm totally fine with any refactor :)

PR #4064, opened 2024-09-03
Demo: Integration of AI into the Serlo editor

kulla:
### Generating new content with a user prompt

1. Go to https://frontend-git-ai-poc-for-serlo-editor-serlo.vercel.app/entity/create/Event/60730
2. Click "Füge ein Element hinzu". Alternatively you can click at the "+" where you want to add the AI-content.
3. Click "Inhalt per AI generieren"
4. Enter a prompt

Screencast for this feature:

https://github.com/user-attachments/assets/778df7a7-06f9-4670-b44c-2bf618dcfad9

### Changing content with a user prompt

1. Go to https://frontend-git-ai-poc-for-serlo-editor-serlo.vercel.app/entity/create/Event/60730 and add some content
2. Click on the content you want to change and the three dots on the right upper corner.
3. Click "Aktuelles Plugin per AI überarbeiten"
4. Enter a prompt

https://github.com/user-attachments/assets/f059c416-5ddf-4b2e-9d93-09c941830f17

### Using the autocomplete feature

1. Go to https://frontend-git-ai-poc-for-serlo-editor-serlo.vercel.app/entity/create/Event/60730
2. Enter some text and wait a little bit
3. Enter "Tab" to accept the autocompleted text

https://github.com/user-attachments/assets/2ee26e4c-ec08-40ea-be80-6090185b101e

### Run this demo locally

1. Go to branch `ai-poc-for-serlo-editor` and run `yarn`.
6. Set ENV variable `OPENAI_API_KEY` via `.env` or bash command `export OPENAI_API_KEY=...`.
7. Run `yarn dev`
8. Go to http://localhost:3000/entity/create/Event/60730

### Update JSON-Schema for OpenAI

1. Open the file [`apps/web/src/api/serlo-editor/content-type.ts`](https://github.com/serlo/frontend/blob/ai-poc-for-serlo-editor/apps/web/src/api/serlo-editor/content-type.ts)
9. Make changes to the TypeScript definition of the Serlo Editor content format.
10. Run "yarn generate".

### ToDos for @kulla 

- [x] Add spinner where the content will be generated
- [x] Fix issue in JSON schema that children in math element cannot be `[]`
- [x] Add yarn command to generate JSON schema from types
- [x] Set up the prototype in Vercel
- [x] Document the prototype for other devs
- [x] Add ErrorBoundary in rows plugin
- [x] Make it work on https://frontend-git-ai-poc-for-serlo-editor-serlo.vercel.app/entity/create/Event/60730
[...]

PR #4348, opened 2024-12-10
feat(editor): add ai experiments (keep)

elbotho:
moving the experiments from #4064 over into the editor so we can test them also in integrations.
not meant for production.
i try to keep a clear separation between the experiments and the rest of the editor if possible.

**status**:
- change and generate: improvements needed
- autocomplete: not started

[Demo](https://frontend-git-feat-editor-ai-experiments-serlo.vercel.app/___editor_preview)


packages/editor/src/plugins/ai-generation/components/promt-form.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEditStrings } from '@editor/i18n/edit-strings-provider'
```
hugotiburtino: typo at the name of the file, _prompt_

elbotho: thanks!

PR #4354, opened 2024-12-15
feat(bildungsraum-share): copy content to clipboard

hejtful:
Linear issue: [DEV-303
](https://linear.app/serlo/issue/DEV-303/create-copy-content-button-in-the-share-modal-of-serloorg-content)
To test: Open any Article, Exercise or ExerciseGroup ([for inspiration](https://serlo.org/___content)), choose the share option, and then click the copy content button from the modal. Paste inside a Text plugin.

@elbotho Design and text content is improvised by me, please feel free to suggest improvements.

- [Demo (Article)](https://frontend-git-feat-bildungsraum-share-copy-editor-content-serlo.vercel.app/27801)
- [Demo Exercise](https://frontend-git-feat-bildungsraum-share-copy-editor-content-serlo.vercel.app/54210)
- [Demo ExerciseGroup](https://frontend-git-feat-bildungsraum-share-copy-editor-content-serlo.vercel.app/53205)


elbotho:
When I test it in the preview I get `{"kind":"not-found"}`.

Might be a problem with the preview. But either wayBut I still get the `👌 Content copied!` so that's unexpected.

elbotho:
On localhost it works in general.
When trying to paste the content from `/1555` I get an error. Not sure why.

elbotho:
Maybe we should make the decoder a bit more relaxed and just skip unsupported plugins instead of failing completely? Not sure.

hejtful:
@elbotho Thanks for the review and testing thoroughly!
> When I test it in the preview I get {"kind":"not-found"}.
>
>Might be a problem with the preview. But either wayBut I still get the 👌 Content copied! so that's unexpected.

Tried fixing by changing the API base to staging, didn't work. I'd like to check if it's a preview issue by merging to staging, once everything else has been resolved, if you agree.

> On localhost it works in general.
> When trying to paste the content from /1555 I get an error. Not sure why.

Nice find! Rows plugin was missing from validation. I added it and improved logging for the future. [commit](https://github.com/serlo/frontend/pull/4354/commits/51865e8eb8ead15418d39a4ec1f4d12b6ced259a)

> Maybe we should make the decoder a bit more relaxed and just skip unsupported plugins instead of failing completely? Not sure.

Not sure either, I'm afraid that would lead to frustration, if the user expects everything to be copied, and it's not. We could show a snack saying that the plugin of a certain type was not copied, because it's invalid, but not sure how helpful that would be, it will still be frustrating to look through a whole article of rows to see what wasn't copied.

elbotho:
> Not sure either, I'm afraid that would lead to frustration, if the user expects everything to be copied, and it's not. We could show a snack saying that the plugin of a certain type was not copied, because it's invalid, but not sure how helpful that would be, it will still be frustrating to look through a whole article of rows to see what wasn't copied.

I suspect the frustration is even bigger if you want to import an article and it does not work at all because of one plugin. 
If you don't mind and if it's not too hard (nesting…) I'd prefer to import everything but the failing plugins. Maybe with an "unsupported" plugin their place.

hejtful:
> I suspect the frustration is even bigger if you want to import an article and it does not work at all because of one plugin.

I wouldn't agree, but respect your UX seniority 😁 

> too hard (nesting…)

The state decoder I borrowed from paste-hack didn't validate nested plugins anyway.

> Maybe with an "unsupported" plugin their place.

The default functionality of the editor is to replace invalid plugins with the Unsupported plugin. 

> I'd prefer to import everything but the failing plugins.

So, do you prefer:
1. Being very helpful and validating nested plugins, and showing a snack for each invalid plugin type in the state? (complex TS magic for nested decoder)
2. Relying on default editor handling of invalid plugins? (very fast, no TS magic)

I already have the in-between solution of validating first level plugins and showing snacks for them, but a half-solution like that isn't great.

elbotho:
> Relying on default editor handling of invalid plugins? (very fast, no TS magic)

def. this.
Maybe adding a single toast like: "there were upsupported plugins…".

hejtful:
@elbotho Ok, current state:
- Only validate basic Rows plugin structure. I'm thinking it's still good to fail if the root Rows plugin is invalid.
- In order to only notify of unsupported plugins on paste, added a custom native event helper. Cleanest solution I could think of + can be re-used if a need, to notify of unsupported plugins in other places, arises.

Would you mind adding the missing German strings that break the pipeline (or just writing them here and I'll add them)?

apps/web/src/components/user-tools/share/share-modal.tsx
```diff
@@ -36,9 +38,22 @@ interface EntryData {
...
+    return 'https://de.serlo-staging.dev'
+  if (currentHost.endsWith('serlo.vercel.app'))
+    return 'https://de.serlo-staging.dev'
+  if (currentHost.endsWith('serlo.org')) return 'https://' + currentHost
+
+  return process.env.NODE_ENV === 'development'
+    ? 'http://localhost:3000'
+    : 'https://de.serlo.org'
+}
+
```
elbotho: this should work as well I think:
``` ts
const base =
        process.env.NODE_ENV === 'development'
          ? 'http://localhost:3000'
          : 'https://' + serloDomain
```

apps/web/src/components/user-tools/share/share-modal.tsx
```diff
@@ -70,10 +85,41 @@ export function ShareModal({
     }
   }
 
+  async function copyContentToClipboard() {
+    try {
+      if (!pathOrId) {
+        throw new Error('No path or entity id provided.')
+      }
```
elbotho: since component does not render without `pathOrId` this should be enough:
```suggestion
    if (!pathOrId) return
    try {
```

apps/web/src/pages/api/frontend/bildungsraum-share.ts
```diff
@@ -0,0 +1,121 @@
...
+          return
+        }
+
+        if (uuid.__typename === 'ExerciseGroup') {
+          const exerciseGroup = parseDocumentString(
+            uuid.currentRevision.content
+          ) as EditorExerciseGroupDocument
+          respondWithContent({ plugin: 'rows', state: [exerciseGroup] })
+          return
+        }
```
elbotho: maybe personal preference :)
```suggestion
        if (
          uuid.__typename === 'Exercise' ||
          uuid.__typename === 'ExerciseGroup'
        ) {
          const exercise = parseDocumentString(uuid.currentRevision.content) as
            | EditorExerciseDocument
            | EditorExerciseGroupDocument
          respondWithContent({ plugin: 'rows', state: [exercise] })
          return
        }
```

apps/web/src/pages/api/frontend/bildungsraum-share.ts
```diff
@@ -0,0 +1,121 @@
...
+          url
+        }
+      }
+      ... on Applet {
+        currentRevision {
+          url
+        }
+      }
+    }
+  }
```
elbotho: I'd suggest either reusing the injection query or (preferred) using a simplified query:
``` ts
query shareEditorContent($path: String!) {
    uuid(alias: { path: $path, instance: de }) {
      __typename

      ... on AbstractEntity {
        currentRevision {
          content
        }
      }
    }
  }
```

(`yarn codegen` for updated types)

apps/web/src/pages/api/frontend/bildungsraum-share.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        if (uuid.__typename === 'ExerciseGroup') {
+          const exerciseGroup = parseDocumentString(
+            uuid.currentRevision.content
+          ) as EditorExerciseGroupDocument
+          respondWithContent({ plugin: 'rows', state: [exerciseGroup] })
+          return
+        }
+
+        return res.status(422).json('unknown entity type')
```
elbotho: small & optional, but we probably know the type. we just don't like it ;)
```suggestion
        return res.status(422).json('unsupported entity type')
```

elbotho (changes requested):

apps/web/src/components/user-tools/share/share-modal.tsx
hejtful: You mean `currentHost` instead of `serloDomain`, or?

apps/web/src/pages/api/frontend/bildungsraum-share.ts
hejtful: Nice, didn't even notice they're the same after cleaning up.

elbotho (approved):
Thank you for the improvements!
After [this PR](https://github.com/serlo/frontend/pull/4362) is merged I think we can test this in staging.

PR #4356, opened 2024-12-16
feat(editor): add iframed demo

elbotho:


LarsTheGlidingSquirrel (approved):
Nice! I will make some followup changes: 
- Make iframe cross-origin. Many issues only come up in cross-origin embedding. The LTI integrations will all be cross-origin embeddings. We could simulate this by using different port numbers.
- Adding some iframe security flags (clipboard copy, ...)

elbotho:
> Adding some iframe security flags (clipboard copy, ...)

`allow="clipboard-read *; clipboard-write *; fullscreen *"` is already on there, did you mean something else?

PR #4360, opened 2024-12-18
fix(editor): Modal in moodle

CodingDive:
I don't love it, but for the modal to not collide with the navigation bar of Moodle, we'd need a z-index higher than 1030 😬 

image

### DIFF
--- a/packages/editor/src/editor-ui/editor-modal.tsx
+++ b/packages/editor/src/editor-ui/editor-modal.tsx
@@ -104,6 +104,8 @@ export function EditorModal({
   )
 }
 
+// The moodle navigation bar has a z-index of 1030... We are using 1040 to make
+// sure the modal can be on top and is not cut off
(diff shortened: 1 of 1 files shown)


elbotho (approved):
"in the z-index wars there are no winners"

(I checked: The modal should be the highest z-index element currently so even though I also don't like this it should be ok)

Related:
Will we not be iframed in moodle as well? (that would make this change unnecessary, right?)

CodingDive:
> Will we not be iframed in moodle as well? (that would make this change unnecessary, right?)

We won't for now, only in Chancenwerk. In Moodle they were able to turn off the outside Shadow DOM. I think style collisions and issues will be an ongoing fight!

elbotho:
>I think style collisions and issues will be an ongoing fight!

and is that better than the alternatives in your opinion?

PR #4363, opened 2024-12-18
feat(editor-ui): onboarding welcome modal and hook

hejtful:
Linear issue: [EDTR-200](https://linear.app/serlo/issue/EDTR-200/onboarding-stepper-with-video)

[Demo](https://frontend-git-feat-onboarding-welcome-modal-serlo.vercel.app/entity/repository/add-revision/277232) (open in incognito tab to see again)


packages/editor/src/editor-ui/welcome-modal/use-welcome-modal.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const localStorageKey = 'hasUserSeenWelcomeModal'
+
+export const useWelcomeModal = () => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  useEffect(() => {
+    const localStorageValue = localStorage.getItem(localStorageKey)
+    if (localStorageValue === null) return setIsOpen(true)
+    const hasSeenModal = decodeWelcomeModalData(localStorageValue)
```
elbotho, line 14: If you don't reuse the preferences:
Could we simplify this by just checking if the key exists?
That way we wouldn't need any type checks and parsing.

packages/editor/src/editor-ui/welcome-modal/use-welcome-modal.ts
```diff
@@ -0,0 +1,36 @@
+import { fold } from 'fp-ts/lib/Either'
+import { pipe } from 'fp-ts/lib/function'
+import * as t from 'io-ts'
+import { useEffect, useState } from 'react'
+
+const localStorageKey = 'hasUserSeenWelcomeModal'
```
elbotho: In the `PreferenceContext` I tried to start a new pattern that all editor storage keys start with `serlo-editor::`.
Maybe you could use that here as well or maybe even use the `PreferenceContext` here?
It also uses localStorage and has helper functions to get and set the Preference without using the context.

Then we would just have a new preference, something like:
``` ts
welcomeModalShown: {
    storageKey: 'serlo-editor::welcome-modal-shown',
    defaultValue: false,
  },
```

packages/editor/src/editor-ui/welcome-modal/welcome-modal.tsx
```diff
@@ -0,0 +1,105 @@
...
+const steps = [1, 2]
+
+const slideWidth = 600
+
+const slideProps = {
+  style: { width: slideWidth },
+  className: 'px-4',
+}
+
+const translateXValuesMap: Record<number, string> = {
```
elbotho, line 107: just curious: why do you set the type here?

packages/editor/src/editor-ui/welcome-modal/welcome-modal.tsx
```diff
@@ -0,0 +1,105 @@
...
+        >
+          <div {...slideProps}>
+            <h1 className="serlo-h1">
+              Herzlich Willkommen! <br />
+              <span className="text-brand-600">beim Serlo Editor</span>
+            </h1>
+            <p className="serlo-p">
+              Der Serlo Editor hilft Dir,{' '}
+              <b>Texte, Bilder und interaktive Aufgaben</b>
+              direkt zu bearbeiten und sofort sehen, wie sie später aussehen.
```
elbotho: after `</b>` and `direkt` is a space missing I think

elbotho (changes requested):
Very clean, thank you.
Also really nice that you animated the transistion already 👍 

Some suggestions.

packages/editor/src/editor-ui/welcome-modal/welcome-modal.tsx
```diff
@@ -0,0 +1,105 @@
...
+                <source
+                  src="https://storage.googleapis.com/assets.serlo.org/wikimedia/Der_Menstruationszyklus.webm"
+                  type="video/webm"
+                />
+              </video>
+            </div>
+            <p className="serlo-p mb-0 mt-4">
+              Weitere <b>Erklärungen und Videos</b> findest du jeweils in der
+              Toolbar der einzelnen Plugins. Einfach auf das{' '}
+              <FaIcon icon={faCircleQuestion} /> klicken.
```
elbotho: sorry one small text-change:
```suggestion
              <FaIcon icon={faCircleQuestion} />&#8202;-Symbol klicken.
```

packages/editor/src/editor-ui/welcome-modal/welcome-modal.tsx
hejtful: Because of line 34: `translateXValuesMap[currentStep]`

> Element implicitly has an 'any' type because expression of type 'number' can't be used to index type '{ 1: string; 2: string; }'.
  No index signature with a parameter of type 'number' was found on type '{ 1: string; 2: string; }'.ts(7053)

packages/editor/src/editor-ui/welcome-modal/use-welcome-modal.ts
hejtful: I do like the idea of reusing an existing context, but this isn't exactly a preference, no?

Naming pattern is great, thanks for pointing that out!

elbotho: Yeah… not a preference. But behaves like one in many ways.
We could also call it `showWelcomeModalOnStartup` then it's almost a preference (just one that you can't change right now) 😉 
Your call.

packages/editor/src/editor-ui/welcome-modal/use-welcome-modal.ts
hejtful: We could, but I like the idea of decoding any value coming from `localStorage`, think it's a good practice.

Also, if we later want to expand on this in any way, for example using logged-in user ID to check if a specific user has seen the modal, in case of shared school devices, we can easily expand the codec.

Don't get me wrong, if you implemented it by just checking if the key exists, I wouldn't mind it. But since we already have this code, I wouldn't remove it.

packages/editor/src/editor-ui/welcome-modal/use-welcome-modal.ts
hejtful: If it's my call, then I'd leave it like it is. No need for it to live in a context when it's only used by one component, it's not a preference and this hook is simple anyway.

PR #4368, opened 2024-12-19
feat(plugin-video): allow video upload via new infra

elbotho:
do not merge before new asset infra is used

TODO: 
- [x] ~investigate bundle size increase!~
- [x] adapt e2e tests
- [x] sanity check backend, file sizes, file types etc
- [x] should we use a player lib or is native video good enough for now?
- [x] test with interactive video

@hejtful this is unfortunately blocked by us not using the new media infra in production yet, but maybe you can already review it. Should work fine in local/staging.


packages/editor/src/plugins/image/controls/upload-button.tsx
```diff
@@ -60,7 +61,7 @@ export function UploadButton({
         <input
           type="file"
           multiple={!!config.onMultipleUpload}
-          accept="image/*"
+          accept="image/*,.gif,.jpg,.jpeg,.png,.svg,.webp"
```
hejtful, line 64: Just curious, why was this necessary? In theory, `image/*` should cover all of these added types.

hejtful (approved):
Looks good to me!

packages/editor/src/plugins/image/controls/upload-button.tsx
elbotho: Good question. Now that I think of it `.gif,.jpg,.jpeg,.png,.svg,.webp` would be better, since we then exclude unwanted image types like `bmp` or `tiff`.

elbotho: ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept))

PR #4372, opened 2024-12-20
fix(editor): reset styles for moodle

CodingDive:
In Moodle we are running in some style collisions with bootstrap. It's hard to validate, but I expect that the following line I added should reset quite a few styles. I'd have preferred to do it in the package editor instead of core, but we'd have to add a div there. 

Let me know what you prefer!   

`all: initial;`

One unintended consequence was that the font was reset to times-new-roman. I then saw that they are technically not loaded in the editor project which is why I had to add them.


elbotho (changes requested):
We don't use karla in other integrations right now.
We fall back to tailwinds system font stack (sans-serif): 
`ui-sans-serif,system-ui,sans-serif,"Apple Color Emoji","Segoe UI Emoji",Segoe UI Symbol,"Noto Color Emoji"`

LarsTheGlidingSquirrel:
Merging this now to be able to release a new editor version. @CodingDive told me it should be fine to merge.

PR #4373, opened 2024-12-22
fix(editor): Modal on serlo.org by applying high z-index only in integrations on content instead of overlay

CodingDive:
Reverts and fixes the problem I introduced in https://github.com/serlo/frontend/pull/4360

### DIFF
--- a/packages/editor/src/editor-ui/editor-modal.tsx
+++ b/packages/editor/src/editor-ui/editor-modal.tsx
@@ -76,7 +76,7 @@ export function EditorModal({
               'serlo-modal',
               className,
               // The moodle navigation bar has a z-index of 1030...
-              isSerlo && 'z-[1040]'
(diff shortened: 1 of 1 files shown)


packages/editor/src/editor-ui/editor-modal.tsx
```diff
@@ -70,7 +72,12 @@ export function EditorModal({
             className={cn(defaultModalOverlayStyles, extraOverlayClassName)}
           />
           <Dialog.Content
-            className={cn('serlo-modal', className)}
+            className={cn(
+              'serlo-modal',
+              className,
+              // The moodle navigation bar has a z-index of 1030...
+              isSerlo && 'z-[1040]'
```
elbotho: this applies it only on serlo, right?

elbotho (approved):

PR #4374, opened 2024-12-22
feat: New audio plugin UI

CodingDive:
Here are some left over todos for someone to take over

- [ ] Talk to Gregor once he's back about visual feedback and UX iteration
- [ ] Adjust styles https://www.figma.com/design/AWxOL9woZFKRlU4mid4yEY/Mockup-Audio-Player?node-id=0-1&p=f&t=Ul3IwCLw7d7IZh1G-0
- [ ] Hook up to the asset server to upload file
- [ ] Consider adding wavesurf visualization to renderer. Right now I only hooked it up when recording (in the editor)

Consider enabling it in the integrations that use our asset server (I only enabled it on Serlo.org for now)


elbotho:
🧹keeping PR but closing since we don't work on it right now. Added to Editor Prototypes and Demos Doc.

PR #4375, opened 2024-12-23
feat(editor): Add basic canvas plugin with tldraw

CodingDive:
Screenshot from 2024-12-23 02-04-38


elbotho:
🧹keeping PR but closing since we don't work on it right now. Added to Editor Prototypes and Demos Doc.

PR #4379, opened 2024-12-23
fix(math-renderer): sanitize html from katex

LarsTheGlidingSquirrel:


elbotho (approved):
This article is probably a good place to test :)
https://frontend-git-math-renderer-sanitize-html-serlo.vercel.app/90410

LarsTheGlidingSquirrel:
> This article is probably a good place to test :)
https://frontend-git-math-renderer-sanitize-html-serlo.vercel.app/90410

That's super useful! Seems to work everywhere. :)

PR #4391, opened 2025-01-10
refactor: remove Next router from Course

hejtful:
[Demo](https://frontend-git-refactor-remove-next-router-from-course-serlo.vercel.app/51979)


apps/web/src/serlo-editor-integration/create-renderers.tsx
```diff
@@ -160,7 +159,6 @@ export function createRenderers(): InitRenderersArgs {
         renderer: DropzoneImageStaticRenderer,
       },
       { type: EditorPluginType.Box, renderer: BoxStaticRenderer },
-      { type: EditorPluginType.Course, renderer: CourseSerloStaticRenderer },
```
elbotho, line 163: this `create-renderers` did not fall back on the other the other `create-renderers` file before. Did you change that? I'd expect that we would need to import CourseStaticRenderer here instead.

apps/web/src/serlo-editor-integration/serlo-plugin-wrappers/course-serlo-static-renderer.tsx
```diff
@@ -1,35 +0,0 @@
...
-  const router = useRouter()
-  const isRevisionView = useContext(RevisionViewContext)
-  const { pages } = props.state
-
-  const strings = useStaticStrings()
-
-  return (
-    <>
-      {pages.length ? null : (
-        <InfoPanel icon={faExclamationCircle} type="warning" doNotIndex>
```
elbotho, line 24: Removing this has SEO implications since empty courses would be indexed.
I'd like to still have this, but we could/should probably move it into `entity.tsx` or something like that.

packages/editor/src/plugins/course/static/course-footer.tsx
```diff
@@ -38,10 +33,7 @@ export function CourseFooter({
   const courseStrings = useStaticStrings().plugins.course
 
   function navigate(toPath: string, newIndex: number) {
-    void router.push(toPath, undefined, { shallow: true })
-    scrollIfNeeded(document.querySelector('#course-title'))
-
-    void router.push(toPath, undefined, { shallow: true })
+    window.location.pathname = toPath
```
elbotho, line 36: How do you feel about the full page load vs the js navigation here?

packages/editor/src/plugins/course/static/static.tsx
```diff
@@ -1,33 +1,31 @@
...
-}: EditorCourseDocument & {
-  router: DummyNextRouter
-  isRevisionView?: boolean
-}) {
+}: EditorCourseDocument) {
   const { pages } = state
 
-  const routerCourseId = getCoursePageIdFromPath(router.asPath)
+  const courseStrings = useStaticStrings().plugins.course
+  const { isRevisionView } = useContext(SerloOnlyFeaturesContext)
```
elbotho, line 17: if we wanted to simplify the context (and are okay with a solution that is a little less clean) we could also set isRevisionView by checking for `/entity/repository/compare/` in the window location instead.
Might need an effect.

apps/web/src/serlo-editor-integration/create-renderers.tsx
elbotho: (and if that works we could potentially remove some more of them)

hejtful: Maybe I misunderstood something, but in the original PR I was removing calls to this `createRenderers` with a plan to remove the file as well in the end. The Editor package calls its own `createRenderers`.

apps/web/src/serlo-editor-integration/serlo-plugin-wrappers/course-serlo-static-renderer.tsx
hejtful: Oh, good to know! I'll try moving it into `entity.tsx` and let you know if I need help there.

packages/editor/src/plugins/course/static/course-footer.tsx
hejtful: The way I see it, the functionality is maintained, the UX got worse, but I didn't spend much time on it, because you mentioned that the UX could suffer if it's only on serlo.org (if I understood you correctly).

packages/editor/src/plugins/course/static/static.tsx
hejtful: Your call :) I slightly prefer keeping the Serlo-only things together, but we can also add a comment above that effect.

apps/web/src/serlo-editor-integration/serlo-plugin-wrappers/course-serlo-static-renderer.tsx
hejtful: Commit [here](https://github.com/serlo/frontend/pull/4391/commits/5bec2f88daa667269982f104070926adb7fb742f). Is that kind of what you had in mind?

apps/web/src/serlo-editor-integration/create-renderers.tsx
elbotho: Thanks for the explanation

elbotho: I'd love to find a way to not move all the revision view logic into the editor.
That's one of the reasons we maintained those extra renderers.
Let's see.

packages/editor/src/plugins/course/static/static.tsx
elbotho: In my mind the entity plugins are all serlo-only as well so it does not really matter.

elbotho: but I'm okay with both solutions. Let's maybe see how we handle the other revision view stuff and then decide.

apps/web/src/serlo-editor-integration/serlo-plugin-wrappers/course-serlo-static-renderer.tsx
elbotho: Thank you!

elbotho (approved):

PR #4393, opened 2025-01-13
refactor: remove direct Editor imports from Editor presentation

hejtful:


elbotho:
So far so good 👍 

Would def. be nice to not show those warnings in this demo but maybe we just hide them with css since this is probably another serlo-only problem :)

image

PS: localstorage box is also currently there so this is only partly relevant for this PR

elbotho (approved):

PR #4405, opened 2025-01-16
fix(editor-package): make testing secret optional & enable image (and related) plugins even if testing secret is missing

LarsTheGlidingSquirrel:
Testing secret available -> Show upload button
Testing secret not available -> Hide upload button


elbotho (approved):
code looks nice and simple, thank you.
Did not do any testing.

PR #4409, opened 2025-01-16
refactor: replace InjectionStaticRenderer with SerloRenderer

hejtful:
/


elbotho:
Quick question. Could we instead of duplicating the plugin out site of the editor just use the editor static renderer and give it an injection plugin document (instead of using the injection static renderer directly)?

hejtful:
> Quick question. Could we instead of duplicating the plugin out site of the editor just use the editor static renderer and give it an injection plugin document (instead of using the injection static renderer directly)?

I will try :)

hejtful:
Works like a charm, thank you for the suggestion!

elbotho (approved):
Chef's kiss

PR #4410, opened 2025-01-16
refactor: export dispatchReplaceRootDocument from package

hejtful:
Not sure if there should be a comment that it's only exported for serlo.org, other consumers might find it handy as well.

### DIFF
--- a/apps/web/src/serlo-editor-integration/components/external-revision-loader.tsx
+++ b/apps/web/src/serlo-editor-integration/components/external-revision-loader.tsx
@@ -1,6 +1,4 @@
-import { TemplatePluginType } from '@editor/package'
-import { runReplaceDocumentSaga, useAppDispatch } from '@editor/store'
-import { ROOT } from '@editor/store/root/constants'
+import { type BaseEditor, TemplatePluginType } from '@editor/package'
 import { faFileImport } from '@fortawesome/free-solid-svg-icons'
 import request from 'graphql-request'
 import NProgress from 'nprogress'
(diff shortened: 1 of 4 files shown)


elbotho (approved):
Nice, thank you!

elbotho:
> Not sure if there should be a comment that it's only exported for serlo.org, other consumers might find it handy as well.

Hmm good question. For now I'd add a comment that it's just for serlo or that it's experimental. I think that's the kind of deep integration that we hope others don't have to do :)

elbotho:
PS: Not important. We could rename `dispatchReplaceRootDocument` to `replaceRootDocument` or something even more descriptive like `replaceWholeEditorState`.

hejtful:
>Hmm good question. For now I'd add a comment that it's just for serlo or that it's experimental. I think that's the kind of deep integration that we hope others don't have to do :)

Fair enough, I'll add a comment then, and also use the [deprecated](https://www.typescriptlang.org/play/?#example/jsdoc-deprecated) tag to make it obvious it shouldn't be used even without reading the source code.

>PS: Not important. We could rename dispatchReplaceRootDocument to replaceRootDocument or something even more descriptive like replaceWholeEditorState.

Because we're marking it as Serlo-only, I would leave the name very technical and internal, as we're not providing a public method to replace the whole Editor state, but we're providing a temporary way for serlo.org to dispatch a certain action. If we wanted to make this a public property, `replaceWholeEditorState` would be a great name.

PR #4412, opened 2025-01-16
fix(bildungsraum-share): copy articleIntroduction as Multimedia plugin

hejtful:
Linear issue: [EDTR-266](https://linear.app/serlo/issue/EDTR-266/erklarung-mit-multimedia-inhalt-ist-not-copied)

### DIFF
--- a/apps/web/src/pages/api/frontend/bildungsraum-share.ts
+++ b/apps/web/src/pages/api/frontend/bildungsraum-share.ts
@@ -1,8 +1,10 @@
+import { EditorPluginType } from '@editor/package'
 import { parseDocumentString } from '@editor/static-renderer/helper/parse-document-string'
 import type {
-  EditorArticleDocument,
+  EditorArticleIntroductionDocument,
   EditorExerciseDocument,
   EditorExerciseGroupDocument,
+  EditorRowsDocument,
 } from '@editor/types/editor-plugins'
 import { gql } from 'graphql-request'
 import type { NextApiRequest, NextApiResponse } from 'next'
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/api/frontend/bildungsraum-share.ts
```diff
@@ -50,11 +52,28 @@ export default async function handler(
...
-          const articleContent = articleDocument.state.content
-          respondWithContent(articleContent)
+          const { introduction, content } = (
+            parseDocumentString(uuid.currentRevision.content) as {
+              state: {
+                introduction: EditorArticleIntroductionDocument
+                content: EditorRowsDocument
+              }
+            }
+          ).state
```
elbotho, line 62: maybe something like that would work?
```suggestion
            parseDocumentString((uuid.currentRevision as EditorArticleDocument).content)).state
```

elbotho (approved):
lovely, thank you!
one idea, but already good to go.

apps/web/src/pages/api/frontend/bildungsraum-share.ts
hejtful: Sadly not, because then the `state` property of `content` and `introduction` is `unknown`, which gives TS errors down the line.

elbotho: @hejtful Hmmm interesting. Another Idea in editor-plugins.ts
``` ts
export interface EditorArticleDocument {
  plugin: EditorPluginType.Article
  state: PrettyStaticState<ArticlePluginState> & {
    introduction: EditorArticleIntroductionDocument
    content: EditorRowsDocument
  }
  id?: string
  serloContext?: {
    articleTitle?: string
  }
}
```

If you agree I can add it in another PR as well.

PR #4413, opened 2025-01-17
refactor: Editor package exports for ___experiments page

hejtful:
I decided to just export it under serlo-only comment for now, we can decide the long-term strategy later.


elbotho (approved):
👍

elbotho:
PS: The bundle size increase here is concerning. Looks like tree shaking does not work like it should… 🎄

PR #4426, opened 2025-01-27
fix: render Karla font on serlo.org

hejtful:
Style reset has been introduced for the needs of Moodle integration here: https://github.com/serlo/frontend/pull/4372, but it's not compatible with serlo.org integration. This PR makes the style reset optional, by adding a `styleReset` property. Assuming that we want style reset for non-Serlo integrations, `styleReset` is on by default, and turned off for serlo.org.

Reintroduces Karla font to https://github.com/serlo/frontend/pull/4380.


elbotho (approved):
Thank you!

elbotho:
PS: Could you add the prop to the ___editor_preview editor as well?

PR #4440, opened 2025-01-30
fix(plugin-image): upload pixabay images to our bucket

elbotho:
https://linear.app/serlo/issue/EDTR-293/pixabay-images-break

(with this at least we don't produce more breaking images, not sure what we should do with all the already broken images to be honest)

[Demo](https://frontend-git-fix-plugin-image-upload-pixabay-serlo.vercel.app/___editor_preview)


hejtful (approved):
Thanks for the superfast fix!

Not sure what we could do with the existing images either. For serlo.org, we could write in the community chat letting authors know about the problem and kindly asking that they pick the Pixabay images again. For integrations, nothing comes to mind :/

PR #4442, opened 2025-01-31
chore(editor-upload): use new asset infra in serlo.org production

elbotho:


packages/editor/src/plugins/image/utils/upload-file.ts
```diff
@@ -11,18 +12,19 @@ type UploadMeta = Pick<EditorMeta, 'editorVariant' | 'userId'>
...
+  const isSerlo = useIsSerlo()
 
-  const uploader = (file: File) => uploadFile({ file, editorVariant, userId })
-  return shouldUseNewUpload() ? uploader : oldUploader
+  const uploader = (file: File) =>
+    uploadFile({ file, editorVariant, userId, isSerlo })
+  return shouldUseNewUpload(isSerlo) ? uploader : oldUploader
 }
 
 // while testing
```
hejtful: Is this obsolete now?

Also the `console.log('using new upload method and temporary bucket')` line?

packages/editor/src/plugins/image/utils/upload-file.ts
```diff
@@ -88,10 +93,11 @@ async function uploadFile({
   return Promise.resolve(fileUrl)
 }
 
-const signedUrlHost =
-  process.env.NODE_ENV === 'development'
-    ? 'editor.serlo.dev'
-    : 'editor.serlo.dev' // TODO: Change to production bucket after testing
+const isSerloProduction = process.env.NEXT_PUBLIC_ENV === 'production'
```
hejtful, line 96: Just to mention that I added a helper hook to https://github.com/serlo/frontend/pull/4348/ , in the `packages/editor/src/core/hooks/use-is-serlo.ts` file, when merging staging to that branch:
```
export function useIsNextjsProduction(): boolean {
  return process.env.NEXT_PUBLIC_ENV === 'production'
}
```

So this will be duplicated once both PRs are merged.

hejtful (approved):
Thanks!

packages/editor/src/plugins/image/utils/upload-file.ts
elbotho: I wish :)
I think it's not because there are integrations outside of serlo.org that are not staging/dev and I want to activate this for serlo.org production only right now.

hejtful: I only meant the `// while testing` comment and the console log. Might be confusing, if we're using it on production now.

Same for integrations, the console log is not super informative without context.

elbotho: Ahh I see.
On Serlo the console.log won't trigger (early return).
But I moved the comment down so it's a bit more clear.
I'd keep the console.log for now for us to check which method is active when working with integrations.

PR #4447, opened 2025-02-03
fix: allow pixabay cdn without image proxy

elbotho:
(I'd probably hotdeploy this as well after merge)


hejtful (approved):
Don't fully understand the need for this, but don't see any problems with it either.

PR #4452, opened 2025-02-04
fix(editor): restrict @editor/package imports to avoid dep cycle

hejtful:
This is mostly to prevent running into cyclical imports.

But also, only `web` should import from `@editor/package`, Editor files should import from other Editor files directly.


packages/editor/src/___dev-demo/standalone.tsx
```diff
@@ -1,4 +1,4 @@
-import { SerloEditor } from '@editor/package'
+import { SerloEditor } from '@editor/package/editor'
```
elbotho: Super minor:
I'm not sure about this one. The demo should simulate how another integration would use the editor so I'd expect an import from `@editor/package` with an ignore comment.

elbotho (approved):
awesome, thank you.

PR #4453, opened 2025-02-04
fix(plugin-interactive-video): solve 'stuck when playing video' in edu-sharing preview.

LarsTheGlidingSquirrel:
In the edu-sharing preview we have `SerloRenderer` plus iframe without allow attribute. I still do not understand why it did not work before, but this change fixes it for now.


elbotho (approved):
okay, let's try it like this.
Did you find a way to reproduce this locally so we could test?
There might be some more changes coming in the next days and I'm not really sure how to realistically test if I break something.

LarsTheGlidingSquirrel:
I could reproduce them locally. I am making some changes to the editor demo page right now so we can easily check for iframe / cross-origin / ... related problems during development.

PR #4454, opened 2025-02-05
fix(blanks-exercise): consistent BlankDraggableAnswer preview positioning

hejtful:
Fixes the drag and drop preview bug from [EDTR-318](https://linear.app/serlo/issue/EDTR-318/investigate-and-fix-interactive-video-testing-feedback).

Changes:
- **Main change**: Render dragged item preview right next to the dragged item, so that the preview positioning would be relative to the dragged item original position. Makes the positioning consistent, even if the dragged item is rendered in a modal etc.
- Remove custom `UseEmptyPreview` hook as it doesn't really work and replace with a helper from [the docs](https://react-dnd.github.io/react-dnd/docs/backends/html5#extras).
- Remove `CustomDragLayer` and replace with a `BlanksExercise`-specific `DraggableAnswerPreview`. As previews will be rendered next to their items from now on, a one-size-fits-all `CustomDragLayer` becomes obsolete.
- Extract styles into its own file to avoid cyclical import.


elbotho (approved):
thank you for the cleeean fix

PR #4460, opened 2025-02-05
chore: simulate cross-origin iframe in editor demo page

LarsTheGlidingSquirrel:
The editor preview page is wrapped in an iframe where the embed using a subdomain. This simulates the cross-origin iframe embedding we have in the LTI integrations. 

With this, we can quickly check if editor code changes also work in the LTI world. Also, we can quickly test iframe security attributes like `sandbox` and `allow`.

Unexpectedly, hot reload still works even though we are in an iframe. :partying_face: 

However, the state is not added to the browser address bar.


elbotho:
Could you give a simple example of something that works same-origin but would fail with the new setup (for testing)?

LarsTheGlidingSquirrel:
> Could you give a simple example of something that works same-origin but would fail with the new setup (for testing)?

The interactive video worked fine in same-origin, but got stuck in loading for cross-origin. Maybe that is easy to test. 
Also: 
- Access to local storage should be restricted for cross-origin, but I have not tested that yet. 
- Setting cookies behaves differently

elbotho (approved):
👍

PR #4462, opened 2025-02-06
chore(editor-dev): simulate cross origin with one dev server

elbotho:
After some research I'm pretty sure that (unless we change some settings) using a subdomain should have exactly the same effects (and in my testing the behaviour is the exact same as with localhost:3001/localhost:3000)


LarsTheGlidingSquirrel (approved):
Nice! I did not know this is possible. Definitely an improvement. :relaxed:

PR #4464, opened 2025-02-06
refactor(save-modal): remove obsolete LocalStorageButton component

hejtful:
> obsolete code since we now save all changes periodically to localStorage using the url (or part of it) as id


elbotho (approved):
Thank you.
One follow up: The feedback is not very helpful now, we should probably still inform the user that they can come back, restore the content and try saving again later.

hejtful:
@elbotho Please feel free to change the message within this PR, or just leave the text as a comment and I'll change it.

PR #4465, opened 2025-02-07
refactor: adjust Frontend to new Editor storage format

hejtful:
Linear: [EDTR-98](https://linear.app/serlo/issue/EDTR-98/use-the-new-migration-algorithm-also-at-serloorg)
API (db-migrations) PR: https://github.com/serlo/api.serlo.org/pull/1865
Infra PR: https://github.com/serlo/infra/pull/120

Should only be merged once the DB migration has been run on staging.


apps/web/src/serlo-editor-integration/convert-editor-response-to-state.ts
```diff
@@ -105,88 +97,123 @@ export function convertEditorResponseToState(
...
+          articles: [],
+          courses: [],
+          videos: [],
+        },
+        sources: [],
+      },
+    })
+  }
+
+  return { editorMetadata, templateContent }
```
LarsTheGlidingSquirrel, line 191: Why is there `template` in the name?

apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx
```diff
@@ -45,7 +45,10 @@ function Content({
   props: EntityCreateProps
 }) {
   const initialState = {
-    plugin: AllowedPlugins[entityType],
+    ...createEmptyDocument('serlo-org'),
+    document: {
+      plugin: AllowedPlugins[entityType],
+    },
```
Comment (LarsTheGlidingSquirrel) [inline apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx:51]: Maybe `createEmptyDocument` could generate different initial editor states in the future. Example: `createEmptyDocument(editorVariant: 'serlo-org', type: 'article')`

LarsTheGlidingSquirrel (approved):
There are some parts that I don't understand, but to me it looks fine. Let me know if I should look into something specific in more detail.

apps/web/src/serlo-editor-integration/convert-editor-response-to-state.ts
hejtful: Because it's representing the content of `TemplatePluginType`. Would `templatePluginContent` be more understandable?

apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx
Comment (hejtful) [reply apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx:51]: If you don't mind, could you please explain the benefits of that in more detail? Is it to avoid this destructuring approach, or are there other benefits?

apps/web/src/serlo-editor-integration/convert-editor-response-to-state.ts
LarsTheGlidingSquirrel: Ah, I see. How about just `content`? Or is this function only for `TemplatePluginType`?

apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx
Comment (LarsTheGlidingSquirrel) [reply apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx:51]: When I created this function in the past there was only one root plugin `type-generic-content` so it made sense to hard code it. Now, there are different types. I think I would slightly prefer having all that logic in function `createEmptyDocument`. But the way it is now is also fine for me. :relaxed:

apps/web/src/serlo-editor-integration/convert-editor-response-to-state.ts
hejtful: >Or is this function only for TemplatePluginType?

Yes. So the editor metadata will be in its `content` property. But the editor's `migrate` function isn't expecting that. More detail [here](https://serlo.slack.com/archives/CEB781NCU/p1738862348768259).

apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx
Comment (hejtful) [reply apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx:51]: Fair point! And it should stay like that, actually. I only exported the `createEmptyDocument` function through the package as a special case for Serlo (which is also documented by comments in the package index file, and also why it's not mentioned in the README). So it's a temporary hacky solution, and I wouldn't change the original function for it.

It's definitely not great, but I decided that these refactorings need to go step by step, if I tried to make everything perfect on the first go, I'd get lost in it 😁

Comment (LarsTheGlidingSquirrel) [reply apps/web/src/pages/entity/create/[type]/[taxonomyId].tsx:51]: Ah, alright. Thanks for clarifying. Going step by step sounds great to me. Just wanted to share my thoughts.

PR #4468, opened 2025-02-10
fix(plugin-edusharing): handle new format of brockhaus embeds

LarsTheGlidingSquirrel:


elbotho:
@LarsTheGlidingSquirrel just to be sure: so "brockhaus embeds" are handled exactly like links now?

LarsTheGlidingSquirrel:
> @LarsTheGlidingSquirrel just to be sure: so "brockhaus embeds" are handled exactly like links now?

Yes! We cannot easily distinguish them anymore using the meta data. But they are basically the same as link embeds anyway.

elbotho (approved):

PR #4493, opened 2025-02-13
refactor(editor): try replacing editorContent with normal child

elbotho:
([problem](https://serlo.slack.com/archives/D04QD2L89FD/p1739887489440559) with event 145590, investigate after migration)


elbotho:
@hejtful I added some adaptations and some more refactoring. Mind having another look?

hejtful (approved):
I'm not sure what the benefits of this refactoring are, but everything works, so your call to merge or not.

PR #4496, opened 2025-02-14
feat(e2e): add more tests for saving modal and saving content

elbotho:


e2e-tests/tests/424-text-plugin-link.ts
```diff
@@ -85,7 +85,7 @@ Scenario('Edit existing link', async ({ I }) => {
   I.seeElement(locate({ css: '.editor-core a' }).withAttr({ href: '/1288' }))
 })
 
-Scenario.only('Remove existing link', async ({ I }) => {
```
hejtful, line 88: This is probably from me, sorry :(

e2e-tests/tests/310-save-changes.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    I.see('Meta-Titel: Differenzenquotient')
+    I.see('Meta-Beschreibung')
+    I.see(
+      'Der Differenzenquotient beschreibt die Steigung einer Funktion zwischen zwei Punkten.'
+    )
+  }
+)
+
+// … maybe add more tests for other entities?
```
hejtful: I usually add `Scenario.todo`s with descriptions of what I plan to test in the future. Not like I often come back to those todos though, so might not be the best way. 😁

hejtful (approved):

e2e-tests/tests/424-text-plugin-link.ts
elbotho: Downside of removing it: Test take a lot longer again 😄

PR #4497, opened 2025-02-14
refactor: remove unnecessary selectors

hejtful:
Motivation for PR - console warning:
> The result function returned its own inputs without modification. e.g
`createSelector([state => state.todos], todos => todos)`
This could lead to inefficient memoization and unnecessary re-renders.
Ensure transformation logic is in the result function, and extraction logic is in the input selectors.

Makes sense, `createSelector` tries to memoize, but there's no need for memoization, as there's no transformation to memoize. So this PR removes selectors without transformation, and their usage is replaced with simple property access.

### DIFF
--- a/packages/editor/src/core/editor-children.tsx
+++ b/packages/editor/src/core/editor-children.tsx
@@ -14,7 +14,6 @@ import {
   selectPendingChanges,
   selectHasUndoActions,
   selectHasRedoActions,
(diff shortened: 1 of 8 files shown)


elbotho (approved):
Thank you,
If we see no behaviour changes this looks sehr gut ;)

PR #4498, opened 2025-02-17
feat: Handle Vidis IdP Hint Query Param 

hugotiburtino:
To test
1. https://frontend-git-vidis-idp-hint-serlo.vercel.app/auth/login?vidis_idp_hint=vidis-idp (Try also changing the value of vidis_idp_hint)
2. Click 'Anmelden mit VIDIS'
3. The vidis_idp_hint= has to be forwarded

### DIFF
--- a/apps/web/src/components/auth/flow.tsx
+++ b/apps/web/src/components/auth/flow.tsx
@@ -247,11 +247,20 @@ export function handleFlowError<S>(
         setFlow(undefined)
         await router.push(flowPath)
         return
-      case 'browser_location_change_required':
+      case 'browser_location_change_required': {
         // Ory Kratos asked us to point the user to this URL.
 
+        const searchParams = new URLSearchParams(window.location.search)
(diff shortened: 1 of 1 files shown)


apps/web/src/components/auth/flow.tsx
```diff
@@ -247,11 +247,20 @@ export function handleFlowError<S>(
         setFlow(undefined)
         await router.push(flowPath)
         return
-      case 'browser_location_change_required':
+      case 'browser_location_change_required': {
         // Ory Kratos asked us to point the user to this URL.
```
elbotho: maybe comment is now misleading. maybe add a comment about vidis?

elbotho (approved):
nice, that would be wonderful if that is enough 🎉

hugotiburtino:
> nice, that would be wonderful if that is enough 🎉

Don't worry, it is really just it that they want.

edit: no it wasn't :/ see follow up PR

PR #4502, opened 2025-02-19
refactor(template-plugins): unify field names in template plugins to simplify

elbotho:


hejtful (commented):
Would this need a new DB migration?

elbotho:
> Would this need a new DB migration?

Oh no, I'd never :) 
As it's intended it only changes the template plugin types that we construct in serlo.org to feed it to the editor but not save in the DB.

hejtful (approved):

PR #4505, opened 2025-02-19
feat(editor): add optional showUndoRedoButtons prop

elbotho:
For https://linear.app/serlo/issue/EDTR-103/integrate-undo-redo-inside-the-editor-for-serloorg-all-integrations

(deployment breaks for … staging reasons, very likely unrelated)

### DIFF
--- a/apps/web/src/serlo-editor-integration/serlo-editor.tsx
+++ b/apps/web/src/serlo-editor-integration/serlo-editor.tsx
@@ -83,6 +83,7 @@ export function SerloEditor({
         isProductionEnvironment={isProduction}
         initialState={initialState}
         styleReset={false}
+        showUndoRedoButtons
         extraSerloPlugins={extraSerloPlugins}
         extraSerloRenderers={extraSerloRenderers}
         onChange={(state) => {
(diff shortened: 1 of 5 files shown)


LarsTheGlidingSquirrel (approved):
Nice! Thank you! Sadly, I think we need to improve the positioning. In the editor package preview the buttons are offscreen. And in the integrations, I fear they could be offscreen or clipped as well. 
How about we render them above content and within the dimensions of the editor?

elbotho:
> Nice! Thank you! Sadly, I think we need to improve the positioning. In the editor package preview the buttons are offscreen. And in the integrations, I fear they could be offscreen or clipped as well.
How about we render them above content and within the dimensions of the editor?

It's responsive and in the normal editing view on serlo it only jumps to the side when we have enough space. 
So for iframes it should be alright. For the preview we could either overwrite that (since it's an internal tool I don't mind) or switch to container queries for the positioning. Never actually used them but support should be good enough by now :)

elbotho:
@LarsTheGlidingSquirrel are you okay with merging it like this and tweaking it later if needed for the integrations?

LarsTheGlidingSquirrel:
Ah, ok. I only tested the editor preview. Great to hear that it should be positioned well everywhere else. 

We can merge now and fix the preview later. :blush:

PR #4512, opened 2025-02-28
chore: add AGPL-v3 license to editor

hejtful:
Version 0.23.0 with the updated license to be published on 10.03, if adopters approve.

### DIFF
--- a/packages/editor-web-component/package.json
+++ b/packages/editor-web-component/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@serlo/editor-web-component",
-  "version": "0.13.2",
+  "version": "0.14.0",
   "homepage": "https://de.serlo.org/editor",
   "bugs": {
     "url": "https://github.com/serlo/frontend/issues"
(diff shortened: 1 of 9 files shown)


packages/editor-web-component/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serlo/editor-web-component",
-  "version": "0.13.2",
+  "version": "0.14.0",
```
kulla:
```suggestion
  "version": "0.14.0-beta.0",
```

packages/editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serlo/editor",
-  "version": "0.22.2",
+  "version": "0.23.0",
```
kulla:
```suggestion
  "version": "0.23.0-beta.0",
```

I have the feeling we should first check whether it looks good on the NPM site (e.g is the string `AGPL-3.0-or-later` the correct one. I would first publish a release candidate and when everything is alright we publish `0.23.0`

kulla (changes requested):

packages/editor/package.json
hejtful: Sounds good to me.

PR #4516, opened 2025-03-03
fix(editor): plugin menu restrictions

hejtful:
Fixes:
1. `PageLayout` plugin appearing in Plugin Menu Modal outside `Page` entity ([slack thread](https://serlo.slack.com/archives/C05QWHQP1HT/p1740993956398599))
2. `InteractiveVideo` appearing as the only interactive plugin on `Page` entity. I'm not 100% sure about this one, but it seems logical to me that we want to restrict `InteractiveVideo` wherever `Exercise` is restricted.


elbotho (approved):
ty

PR #4528, opened 2025-03-17
Revert contentOnly pages: Use old iframe resizing method using `iframe-resizer` instead of `open-iframe-resizer`

LarsTheGlidingSquirrel:
Quick fix for serlo injections that do not resize in RLP 'production'. 

Issue: RLP 'production' uses an outdated `iframe-resizer` host script. serlo.org uses a newer `open-iframe-resizer` content window script. Those probably cannot communicate correctly and the resizing fails.


elbotho (approved):
thank you,
feel free to merge / test

PR #4546, opened 2025-03-28
refactor: replace KaTeX with MathML, temml and Fira Math font

elbotho:
Experiment.
Demo: https://frontend-git-katex-to-mathml-serlo.vercel.app/community/298700/lerne-latex
Current version: http://serlo.org/community/298700/lerne-latex

I would like to test this on serlo.org staging, ~but it needs a way import styles before releasing in the editor~

I just realized that MathML support is very good now across the main browsers. 

Upsides:
- very pretty sans serif font 
- less markup
- less traffic
- probably better performance
- better for SEO
- less custom stuff
- mhchem should also work
- a bit less bundle size (but still no lightweight)

Potential downsides:
- unknown how compatible our LaTeX code is (but [coverage should be good](https://temml.org/docs/en/supported))
- newer solution, less established
- …?

Unfortuately copy/pasting from/to other software does not work in both approaches.

Todo: Test, discuss, ~add macros again~.


packages/editor/src/plugins/text/static-components/static-math.tsx
```diff
@@ -30,52 +46,35 @@ export function StaticMath({ src, inline }: StaticMathProps) {
...
+            'inline-block pb-1 [page-break-inside:avoid]',
+            inline ? 'text-[1.1rem]' : 'text-[1.33rem]'
+          )}
+          dangerouslySetInnerHTML={{ __html: mathML }}
+        />
+      )
+    } catch {
+      // eslint-disable-next-line no-console
+      console.error('formula could not be rendered')
+      return <span />
```
LarsTheGlidingSquirrel: Maybe send error to Sentry as well so we as devs get notified about these errors directly?

LarsTheGlidingSquirrel:
For using this in the editor npm package, I think we need a way to apply the styling there. Do you have any preferences how to do this?

packages/editor/src/plugins/text/static-components/static-math.tsx
elbotho: Good idea, but so far we have no Sentry support in the editor at all and I don't really feel like adding more serlo-only stuff right now :)

LarsTheGlidingSquirrel: I was thinking of only temporarily adding it (and only active on serlo.org) to check how much existing latex on serlo.org does not work.

LarsTheGlidingSquirrel: But, I see other ways to check that might be easier. For example, instead of rendering an empty `<span>`, rendering an error message and asking the user to let us know.

elbotho: Ah I see. This gives me an idea of bulk checking almost all formulas, then we don't need the users to unknowingly test it for us :)

LarsTheGlidingSquirrel: That would be optimal! :green_heart: 
In general, I think rendering an error message like "Sorry, your latex could not be rendered" instead of an empty `span` would be better for the user experience as well.

elbotho: update:
- ran almost all of our formulas through the temml renderer
- found 40 ones that could not be rendered
- temml-author fixed first error already [upstream](https://github.com/ronkok/Temml/issues/84)
- provided them with the rest of the failing cases, let's see… hopefully they are also easy fixes
- added a small error message instead of the empty span

LarsTheGlidingSquirrel: Great news! That already sounds promising! :smiling_face_with_three_hearts:

elbotho: With the [new version of Temml](https://github.com/ronkok/Temml/issues/84#issuecomment-2773356219) all formulas render without throwing 🥳 
I also fixed some faulty formulas in the content so now everything should render without errors.

LarsTheGlidingSquirrel: Awesome! :partying_face:

elbotho:
@LarsTheGlidingSquirrel sorry for moving the tailwind files in here, that makes the PR a lot less readable than it could be.

I did some more test and it looks find for serlo.org content so far.
A bit smaller / compact in some places, but nothing obviously wrong I could find.
Will merge to staging for testing and make a new editor version.

PR #4548, opened 2025-03-30
feat(editor): add `loading="lazy"` to images

elbotho:
☺️ 

follow up: we could load the first image in an article (the one in the intro) eagerly or make other rules


hejtful (approved):
🦥

PR #4549, opened 2025-03-31
refactor: simplify courses

elbotho:
It's a bit annoying because courses had so much refactoring already…
But I gave this more thought and did more SEO research I think I'm pretty happy with this approach.

The idea is to put all course page below each other on one page with some space in between the pages.
Navigation between the pages then works by hash-links (or if users prefer by scrolling).

[Demo](https://frontend-git-refactor-simplify-courses-serlo.vercel.app/65414)

This can be reviewed and discussed, but not merged yet ([needs cf-worker redirects](https://github.com/serlo/cloudflare-worker/pull/1043)).

Thoughts:
- UX: We want to keep the "step by step" feeling to not overwhelm the learners with content (done with vertical space)
- UX: We want quick navigation between the steps (is immediate now)
- SEO: Many course pages are pretty short an that's not great for SEO. So all on one page is probably better overall
- SEO: While Google says they value invisible, but accessible content (like in accordions) the same as visible content test show they really clearly prefer visible content (no content is hidden)
- DX: Having one url per entity is definitely cleaner and technically easier with less special treatment for courses
- DX: one less custom serlo variable in the editor: always nice


packages/editor/src/plugins/course/editor.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <CoursePagesRenderer
+        pages={pages.map((page) => {
+          return {
+            id: page.id.value,
+            title: page.title.value,
+            titleElement: <EntityTitleInput title={page.title} forceFocus />,
+            contentElement: page.content.render(),
+          }
+        })}
+      />
```
hejtful: Minor, but useMemo helps performance and readability a little bit here.

```suggestion
      <CoursePagesRenderer pages={rendererPages} />
```

```typescript
const rendererPages = useMemo(() => pages.map((page) => ({
  id: page.id.value,
  title: page.title.value,
  titleElement: <EntityTitleInput title={page.title} forceFocus />,
  contentElement: page.content.render(),
})), [pages])
```

packages/editor/src/plugins/course/editor.tsx
```diff
@@ -0,0 +1,68 @@
+import { AddButton } from '@editor/editor-ui'
+import { useEditStrings } from '@editor/i18n/edit-strings-provider'
+import { EntityTitleInput } from '@editor/plugins/serlo-template-plugins/common/entity-title-input'
+import { EditorPluginType } from '@editor/types/editor-plugin-type'
+import { useEffect } from 'react'
+import { v4 as uuidv4 } from 'uuid'
+
+import { EditorCourseNavigation } from './editor-course-navigation'
+import type { CourseProps } from './index.jsx'
```
hejtful: Probably:
```suggestion
import type { CourseProps } from '.'
```

hejtful (approved):
Appreciate the SEO tests!

PR #4571, opened 2025-04-28
refactor(editor): add config to diable video upload

elbotho:


packages/editor/src/plugins/video/components/video-selection-screen.tsx
```diff
@@ -48,34 +50,39 @@ export function VideoSelectionScreen({
...
-              >
-                {uploadStrings.invalidUrl}
-              </span>
-              <EditorTooltip text={uploadStrings.invalidUrlMessage} />
-            </>
-          )}
-        </div>
 
-        <UploadButton src={src} />
+        {config.disableFileUpload ? null : (
```
LarsTheGlidingSquirrel: Maybe an "exit-early if" higher up would improve readability.

LarsTheGlidingSquirrel (approved):
Thank you for thinking ahead. Disabling video upload instead of the whole plugin is better than what I proposed. :relaxed:

packages/editor/src/plugins/video/components/video-selection-screen.tsx
elbotho: thanks, I think there was also a bug.
I refactored it into a new comp now.

elbotho:
closed in favour of #4572 4572

PR #4572, opened 2025-04-29
refactor(editor): simplify editor plugin config, remove testingsecret image upload

elbotho:
- only rows, image and textplugin have configs now, and also less fields
- no plugin needs variables in `create-plugin.tsx` (no `create[…]Plugin`) any more
- only support for new media upload infra
- new editor prop for disabling media upload

Integrations that used the `testingsecret` approach should now use the new infra via `editor.serlo.dev` now by default, where CORS is set to accept all hosts. 

---

remaining config:

```  ts
export interface RowsPluginConfig {
  allowedPlugins?: (EditorPluginType | string)[]
  parentType: string
  textPluginPlaceholder?: string
}

export interface ImagePluginConfig {
  onMultipleUpload?: (files: File[]) => void
}

export interface TextEditorConfig {
  placeholder?: string
  isInlineChildEditor?: true
  formattingOptions?: TextEditorFormattingOption[]
  noLinebreaks?: boolean
  controls?: InlineTextEditorControls
}

export interface InlineTextEditorConfig {
  placeholder?: TextEditorConfig['placeholder']
  noLinebreaks?: TextEditorConfig['noLinebreaks']
  controls: Required<InlineTextEditorControls>
}

```

Breaking change for Editor-Package: Removed prop `_testingSecret` from Editor react component.


LarsTheGlidingSquirrel (approved):
Looks great! I also tested the new editor package in the LTI tool: 
- `disableMediaUpload` works fine 
- plugin nesting did not change
- `isProduction` controls choice of media bucket (uploading to 'production' failed but probably because of CORS not allowing localhost)

